PR #26, opened 2023-10-03
Support simulated environment

vinayada1:


rynowak (approved):
This is minimal on details, but I'm going to mark it as approved so we can keep the historical record for this change.

PR #30, opened 2023-11-07
Adding design doc for supporting publishing recipes to insecure registry

vishwahiremat:
Adding design doc for supporting publishing recipes to insecure registry.

Issue: https://github.com/radius-project/radius/issues/6648

Design Doc: https://github.com/radius-project/design-notes/blob/c92de0aeb006c48887388cb1726bb16e9ca37f65/recipe/2023-11-support-insecure-registries.md

### DIFF
--- a/recipe/2023-11-support-insecure-registries.md
+++ b/recipe/2023-11-support-insecure-registries.md
@@ -1,0 +1,185 @@
+# Adding Support for publishing recipes to insecure registries
+
+* **Status**: Pending
+* **Author**: Vishwanath Hiremath (@vishwahiremat)
+
+## Overview
+
+Today we publish/pull recipes to/from the registry that is OCI compliant and only support secure connections i.e registries that are SSL/TLS enabled. And we don't have support to use insecure registry e.g locally hosted k3d registry. And we hit _"http: server gave HTTP response to HTTPS client"_ error if insecure registry is used. This document is centered around addressing this issue.
+
+## Terms and definitions
+
+| Term        | Definition |
+| :-- |:--|
+| insecure registry      | An insecure registry is the one not having either valid registry certificate or is not using TLS and does not use secure communication protocols like HTTPS. |
+| ORAS      | ORAS is the de facto tool for working with OCI Artifacts.https://oras.land/docs/ |
+| Recipes      | Recipes enable a separation of concerns between infrastructure operators and developers by automating infrastructure deployment. https://docs.radapp.dev/author-apps/recipes/ |
+
+## Objectives
+
+> **Issue Reference:** https://github.com/radius-project/radius/issues/6648
+
+### Goals
+
+-   Enable the support to publish and pull recipes using a non-secure registry.
+
+### Non goals
+
+
+### User scenarios (optional)
+
+#### User story 1
+As a Radius user I would like to use locally hosted insecure registry to publish bicep recipes.
+
+Scenario: I create a k3d managed cluster running:
+```
+k3d registry create myregistry.localhost --port 5000
+```
+And use localhost:5000 as the registry path to publish bicep recipes.
+```
+rad bicep publish --file ./redis-test.bicep --target br:localhost:5000/myregistry/redis-test:v1
+```
+I would like to see `rad bicep publish` command successfully publish bicep file to locally hosted registry and also I should be able to read it from the registry during recipe deployment. 
+
+
+## Design
+
+### Design details
+Since we use ORAS client to communicate with the registry, it provides an option "plain-http" to allow insecure connections to registry without SSL check. `plainHttp` property should be set to `true` when client to the remote repository is created to support using insecure registries.
+
+#### Introduce "insecure-http" opt-in flag for users.
+
+Users can use `insecure-http` flag with `rad bicep publish` command and `insecureHttp` property while defining bicep recipes to specify the registry used is insecure and allow connections to registry without SSL check. And based on this property `plainHttp` option is set on the remote repository client. 
+
+
+
+### API design (if applicable)
+
+***Model changes***
+
+Addition of optional property `insecureHttp` to BicepRecipeProperties (as it's only valid for template kind `bicep`).
+
+environments.tsp
+```diff
+@doc("Represents Bicep recipe properties.")
+model BicepRecipeProperties extends RecipeProperties {
+  @doc("The Bicep template kind.")
+  templateKind: "bicep";
+
++  @doc("Allows an insecure connection to a Bicep registry without doing an SSL check. Used commonly for locally-hosted registries. Defaults to false (require SSL).")
++  insecureHttp?: boolean;
+}
+
+@doc("The properties of a Recipe linked to an Environment.")
+model RecipeGetMetadataResponse {
+  @doc("The format of the template provided by the recipe. Allowed values: bicep, terraform.")
+  templateKind: string;
+
+  @doc("The path to the template provided by the recipe. Currently only link to Azure Container Registry is supported.")
+  templatePath: string;
+
+  @doc("The version of the template to deploy. For Terraform recipes using a module registry this is required, but must be omitted for other module sources.")
+  templateVersion?: string;
+
+  @doc("The key/value parameters to pass to the recipe template at deployment.")
+  parameters: {};
+
++  @doc("Allows an insecure connection to a Bicep registry without doing an SSL check. Used commonly for locally-hosted registries. Defaults to false (require SSL).")
++  insecureHttp?: boolean;
+}
+```
+
+***CLI changes***
+
+Add insecure-http flag to allow insecure connection to a Bicep registry without doing an SSL check.
+#### Bicep Publish
+
+```
+rad bicep publish --file ./redis-test.bicep --target br:localhost:5000/myregistry/redis-test:v1 --insecure-http
+```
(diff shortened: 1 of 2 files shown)


recipe/2023-11-support-insecure-registries.md
```diff
@@ -0,0 +1,185 @@
...
+| ORAS      | ORAS is the de facto tool for working with OCI Artifacts.https://oras.land/docs/ |
+| Recipes      | Recipes enable a separation of concerns between infrastructure operators and developers by automating infrastructure deployment. https://docs.radapp.dev/author-apps/recipes/ |
+
+## Objectives
+
+> **Issue Reference:** https://github.com/radius-project/radius/issues/6648
+
+### Goals
+
+-   Enable the support to publish and pull recipes using a non-secure registry.
```
ytimocin, line 24: Can we explain why we are pursuing this goal? What blockage did we run into that made us think that we have to support insecure registries?

ytimocin: Was it something in the pipeline?

ytimocin: Or local testing?

recipe/2023-11-support-insecure-registries.md
```diff
@@ -0,0 +1,185 @@
...
+#### Recipe Register 
+
+```
+rad recipe register cosmosdb --template-kind bicep --template-path br:localhost:5000/myregistry/redis-test:v1 --resource-type Applications.Datastores/mongoDatabases --insecure-http
+```
+
+#### Recipe List/Show
+Add a new column to recipe list/show output table.
+
+| NAME        | TYPE           | TEMPLATE KIND  | TEMPLATE VERSION | TEMPLATE | INSECURE-HTTP|
```
AaronCrawfis, line 109: I'm worried that we might be scaling too the width of the table too far.

I think we can remove it from `rad recipe list` for the table format.

And for `rad recipe show` we may only want to show it on `rad recipe show -o json`

recipe/2023-11-support-insecure-registries.md
```diff
@@ -0,0 +1,185 @@
...
+
+Scenario: I create a k3d managed cluster running:
+```
+k3d registry create myregistry.localhost --port 5000
+```
+And use localhost:5000 as the registry path to publish bicep recipes.
+```
+rad bicep publish --file ./redis-test.bicep --target br:localhost:5000/myregistry/redis-test:v1
+```
+I would like to see `rad bicep publish` command successfully publish bicep file to locally hosted registry and also I should be able to read it from the registry during recipe deployment.
```
ytimocin, line 42: Great feature! Will be very helpful in local testing!

recipe/2023-11-support-insecure-registries.md
```diff
@@ -0,0 +1,185 @@
...
+
+
+## Design
+
+### Design details
+Since we use ORAS client to communicate with the registry, it provides an option "plain-http" to allow insecure connections to registry without SSL check. `plainHttp` property should be set to `true` when client to the remote repository is created to support using insecure registries.
+
+#### Introduce "insecure-http" opt-in flag for users.
+
+Users can use `insecure-http` flag with `rad bicep publish` command and `insecureHttp` property while defining bicep recipes to specify the registry used is insecure and allow connections to registry without SSL check. And based on this property `plainHttp` option is set on the remote repository client.
```
ytimocin, line 52: Is this flag going to be on all the time? Or just when the URL is something specific?

recipe/2023-11-support-insecure-registries.md
```diff
@@ -0,0 +1,185 @@
...
+        }
+      }
+    }
+  }
+}
+```
+
+## Alternatives considered
+
+#### Automatically identify the insecure registry.
```
AaronCrawfis, line 152: I still want to explore if we can special-case `localhost` so `insecure-http` is set to true automatically. The user can still set it to false manually for the edge case where they roll their own SSL certificate locally. And it will allow the publish and register commands to work be default the first time instead of getting an error message.

AaronCrawfis: Especially because a user won't know there's a problem for Recipe registration until a Recipe deployment is attempted. Having it "just work" without waiting until a deployment to see the error message is a much better user experience.

AaronCrawfis: For context the oras client special cases localhost to use insecure http.

recipe/2023-11-support-insecure-registries.md
```diff
@@ -0,0 +1,185 @@
...
+## Test plan
+
+#### Unit Tests
+-   Update environment conversion unit tests to validate insecureHttp property.
+-   Add cli unit tests to validate insecure-http flag.
+-   Update environment controller and config loader tests.
+
+#### Functional tests
+- 	Add a e2e test to verify insecure-http flag for `rad recipe register` and `rad bicep publish` commands
+-   Add a e2e test to deploy a recipe pushed to a locally hosted registry.
```
kachawla, line 172: Since this is specific to bicep, should we add some tests to ensure we reject the request if this flag is set for terraform recipes at registration?

recipe/2023-11-support-insecure-registries.md
ytimocin, line 1: Have we also thought about possible security issues insecure registries may bring?

nithyatsu: For the same reason, would it be good to restrict insecure registries to just localhost?

recipe/2023-11-support-insecure-registries.md
```diff
@@ -0,0 +1,185 @@
...
+```
+I would like to see `rad bicep publish` command successfully publish bicep file to locally hosted registry and also I should be able to read it from the registry during recipe deployment. 
+
+
+## Design
+
+### Design details
+Since we use ORAS client to communicate with the registry, it provides an option "plain-http" to allow insecure connections to registry without SSL check. `plainHttp` property should be set to `true` when client to the remote repository is created to support using insecure registries.
+
+#### Introduce "insecure-http" opt-in flag for users.
```
kachawla, line 50: Have we considered alternative parameter names that more explicitly describe the action being taken, such as 'skipTLSVerification'? The term 'insecure' can be somewhat vague and might inadvertently give a negative impression.

recipe/2023-11-support-insecure-registries.md
rynowak: 👍 

I'm not sure there's a need to show this in `rad recipe list/show`. There's a limit to what what we can do with tables. If users need to see every detail, that's what `-o json` is for.

recipe/2023-11-support-insecure-registries.md
rynowak: > I still want to explore if we can special-case `localhost` so `insecure-http` is set to true automatically. The user can still set it to false manually for the edge case where they roll their own SSL certificate locally. And it will allow the publish and register commands to work be default the first time instead of getting an error message.

I want to argue about this 😆 because I don't think it's useful.

For locally-hosted registries the pull and push URLs are asymmetric. Pushes use localhost, and pulls **DO NOT**. Users still deal with the same level of complexity because Kubernetes is involved. I don't think the heuristic helps. 

Our usecase is different from ORAS' use case. I'd rather be explict, and I don't think this detail is worth explaining to users, it feels like noob-trap.

There's also no such thing as setting a command-line **flag** to false, so it requires a different design. Security should be opt-out rather than opt-in.

recipe/2023-11-support-insecure-registries.md
rynowak: This is why weaking security is **explicit** for users. It's not a security **issue** if it's communicated clearly and the user understands the choice they are making. 

To paraphrase one of my security-expert friends:

> Good security features are **usable**. If security features are not usable, then users will find bad alternatives.

The user is able to understand their scenario and whether it's appropriate. Our code is general purpose and can't make assumptions.

recipe/2023-11-support-insecure-registries.md
vishwahiremat: By default the flag is set to false meaning: if the flag is not specified we consider user is providing secure registry.

recipe/2023-11-support-insecure-registries.md
vishwahiremat: We ran into issue when we are working test framework for recipes repository where we wanted to publish recipes to local k3d managed registry and we were hitting this issue
image

recipe/2023-11-support-insecure-registries.md
vishwahiremat: today docker uses "--insecure" flag and ORAS uses "--plain-http" flag to be set if user wants to use insecure registry. 
I and @rynowak discussed on this and decided on using `--insecure-http` to be explicit to the users that they would be using `http` link i.e `insecure` but  many of us were leaning towards using `--plain-http` flag to be consistent with ORAS and since we would be setting plainHttp flag in the backend. 
@ryanwaite wanted to know your thoughts on this.
cc: @kachawla, @AaronCrawfis , @willtsai

vishwahiremat: I Had an offline discussion with @ryanwaite and agreed upon using `--plain-http` flag.  Will update it in the doc

kachawla (approved):

PR #32, opened 2023-11-28
Add notes for Terraform template path validation

sk593:
This design doc is for adding Terraform template path validation for module sources. Radius supports Terraform module registries and HTTP URLs and will return a validation error if a different module source type is provided

### DIFF
--- a/recipe/2023-11-validate-template-path.md
+++ b/recipe/2023-11-validate-template-path.md
@@ -1,0 +1,96 @@
+# Validation for Terraform Recipe Template Paths
+
+* **Status**: Pending
+* **Author**: Shruthi Kumar (@sk593)
+
+## Overview
+
+Currently, we support only Terraform registry and HTTP URLs as allowed module sources for Terraform recipe template paths. However, we lack a proper validation mechanism for when an unsupported source is given during either recipe registration or recipe deployment. Consequently, when a deployment fails, the reason provided to users is ambiguous resulting into a poor user experience.
+
+## Terms and definitions
+
+| Term     | Definition                                                                                                                                                                                                 |
+| -------- | ---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| Terraform module source | The source argument in a module block tells Terraform where to find the source code for the desired child module |
+| Module registry | A module registry is the native way of distributing Terraform modules for use across multiple configurations |
+| Terraform registry | Terraform Registry is an index of modules shared publicly |
+| HTTP URLs | When you use an HTTP or HTTPS URL, Terraform will make a GET request to the given URL, which can return another source address | 
+
+## Objectives
+
+> **Issue Reference:** https://github.com/radius-project/radius/issues/6642
+
+### Goals
+
+- If a supported module source for a Terraform recipe is provided, the recipe registration should succeed 
+- If an unsupported module source for a Terraform recipe is provided, the recipe registration should fail and return a detailed error message to the user.
+
+
+### Non goals
+
+- Adding support and testing for additional module sources (Github, S3, etc). These will be addressed as needed based on customer feedback.  
+
+### User scenarios (optional)
+
+#### User story 1
+
+As a Radius user, I want to define a Terraform recipe with a Terraform module template path. My module source is a Terraform registry or HTTP URL. I am able to register my Terraform recipe with this source and deploy it successfully. 
+
+#### User story 2
+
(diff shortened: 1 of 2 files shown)


recipe/2023-11-validate-template-path.md
```diff
@@ -0,0 +1,96 @@
...
+#### User story 2
+
+As a Radius user, I want to define a Terraform recipe with a Terraform module template path. My module source is not the supported Terraform registry or HTTP URLs (i.e. Github, local paths, S3 buckets, etc). When I try to register the recipe to my environment, I recieve an error message notifying me that I can only provide a Terraform module registry or HTTP URL as a valid module source using Radius. 
+
+## Design
+
+### Design details
+
+### API design (if applicable)
+N/A. We are updating an existing API to include more validation, but this won't include any public-facing or user-facing changes.
```
rynowak, line 48: minor nitpick: tightening validation is **usually** a breaking change. I think we're ok with a breaking change in this case.

recipe/2023-11-validate-template-path.md
```diff
@@ -0,0 +1,96 @@
...
+HTTP URLs: 
+Terraform treats HTTP URLs as remote addresses and doesn't have functionality for checking HTTP URLs in isolation. We will need to implement our own validation for HTTP URLs. Go provides validation for URIs as part of the `url` [package](https://pkg.go.dev/net/url#ParseRequestURI), but we will likely need additional checks to ensure that a `.zip`/`.tar`/etc suffix is present as well. 
+
+Any module source that is not a Terraform registry or an HTTP URL will be treated as unsupported. We will return a `NewClientErrInvalidRequest` [error](https://github.com/radius-project/radius/blob/main/pkg/armrpc/api/v1/error.go#L64) in this case. The error message will indicate that we only support Terraform registries and HTTP URLs currently. 
+
+## Alternatives considered
+
+1. Adding a `ValidateTemplateKind` API endpoint (similar to `GetRecipeMetadata`).
+This was rejected because the logic could be added to the existing `CreateOrUpdateEnvironment` API endpoint instead. That way, there's less code to add and we can return a validation error before we save any environment information. 
+2. Support all Terraform module sources instead of restricting supported sources.
```
rynowak, line 69: A big reason why it seems ok to limit this is that we haven't done the work to enable private sources yet. Github, http(s) and public registry are the main sources for public modules. 

We'll need to revisit the set of supported sources in the future when we enable private sources.

recipe/2023-11-validate-template-path.md
```diff
@@ -0,0 +1,96 @@
...
+Support for other modules will be done on an as-needed customer basis. This needs to be investigated more to see which module sources need more support and if it's viable to support all of them easily. 
+
+
+## Test plan
+
+Unit testing:
+- Add tests for accepted module sources: HTTP URLs and Terraform module registries 
+  - For HTTP URLs, make sure to test negative cases as the Go URL package isn't very robust
```
rynowak, line 77: Thanks for writing this down :)

recipe/2023-11-validate-template-path.md
```diff
@@ -0,0 +1,96 @@
...
+| Terraform registry | Terraform Registry is an index of modules shared publicly |
+| HTTP URLs | When you use an HTTP or HTTPS URL, Terraform will make a GET request to the given URL, which can return another source address | 
+
+## Objectives
+
+> **Issue Reference:** https://github.com/radius-project/radius/issues/6642
+
+### Goals
+
+- If a supported module source for a Terraform recipe is provided, the recipe registration should succeed
```
sk593, line 25: Add non-goal for changing behavior of public registries

recipe/2023-11-validate-template-path.md
```diff
@@ -0,0 +1,96 @@
...
+## Test plan
+
+Unit testing:
+- Add tests for accepted module sources: HTTP URLs and Terraform module registries 
+  - For HTTP URLs, make sure to test negative cases as the Go URL package isn't very robust 
+- Add tests for module sources that are not accepted (validate error thrown): local paths, Github, S3, etc 
+- Add tests in environment conversion for different module sources (and validate errors/successes)
+
+## Security
+
```
rynowak, line 82: The main thing I'd call out for security is that **any** parsing code that operates on untrusted input can cause security issues, usually by introducing an infinite loop. 

We should avoid writing our own parsers, and rely on libraries. This is why it's important that we're using the built-in Go URL type.

kachawla (approved):

PR #33, opened 2023-11-30
Add diagnostics prototype

rynowak:


rynowak:
This is something I've been discussing with @MarioHewardt - looking for feedback on whether this seems useful, and how we could make it more magical!

resources/2023-11-application-diagnostics-prototype.md
```diff
@@ -0,0 +1,215 @@
...
+
+For the evolution of this feature it should not matter what language we use to build the application.
+
+#### Responding to an alert about application availability
+
+As an SRE, I need to support applications that I did not write, and I may not understand their architecture. I have to use the diagnostics from applications to understand each incident and recommend actions. For applications that are managed by Radius I get comprehensive metadata that helps with navigation and comprehension of the telemetry signals. I can easily identify the environment, application name, and the specific component that has issues. Then I can use the application graph to understand how the application is architected and which cloud resources are involved. This helps me identify the source of a problem as well as whom to contact.
+
+In our demo setup, we're going to create an alert for the error rate coming out of an HTTP application (error golden signal). Our SRE will use the metadata to query the application graph and understand the architecture. Then they will craft a log query that helps them identify the failure pattern. 
+
```
rynowak, line 76: Something I was thinking about here... It would be cool if we could have the recipe we're using for Redis hook up Azure Monitor as well. 

My idea goes something like this:

- SRE see the alert
- First thing they do is check the application graph to understand the architecture and see that the application is using Redis.
- They suspect that Redis is down, so they check that with an ad-hoc query. All green!
- So then they dive into logs and find the bug. 

Thoughts? /cc @AaronCrawfis @kachawla

MarioHewardt: An additional aspect we should consider for the prototype/discovery process is the iterative nature of diagnostics. Any diagnostics solution has an inherent overhead. Typically, the diagnostics dial (for example log level in case of logging) is turned way down in production _until_ an issue surfaces. When it does, the dial gets turned higher and higher (depending on root cause). This iterative process needs to be super simple and as non intrusive as possible.

MarioHewardt: For example, once I've identified the component that is failing via the architecture graph, even something as simple as moving a slider to up the log level would be super cool.

resources/2023-11-application-diagnostics-prototype.md
```diff
@@ -0,0 +1,215 @@
...
+  - Navigate to the application graph and understand the application architecture.
+  - Navigate to the logs and craft a query to understand the problem.
+  - *stretch* understand when and how the problem was introduced based on logs.
+
+## Terms and definitions
+
+| Term                 | Definition                                                                                                                                                                                                                                                              |
+| -------------------- | ----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| Diagnostics          | In this context, we're referring to metrics, tracing and logs. There are other technologies that could be considered diagnostic data such as profiling and crash-dumps, but they are out of scope for this discussion.                                                  |
```
MarioHewardt, line 23: Just to clarify, the other diagnostics sources (profiles, dumps etc) are out of scope for the demo/prototype but most certainly in scope for the overall diagnostics story?

nithyatsu:
This is really nice 😊.  To help my understanding, in order for Radius to support monitoring applications that are rad deployed, an approach is for the application components to use open telemetry and generate observability signals of interest. We also would like to define thresholds at which ops might want to make sure nothing is wrong instance, perhaps a storage service reaches threshold, CPU/memory is 100% or http error code rate is higher. Once radius notices these thresholds are crossed, it can make the component yellow (and cascade color for dependent services)/ trigger an alarm suggesting check observability tools for the yellow service (or take other steps like increase log verbosity for a few seconds). Is this what we are aiming for at a high level?

resources/2023-11-application-diagnostics-prototype.md
rynowak: This is another idea that's super cool 👍 

I think we can probably doing this in centralized way through the configuration of the collector: https://opentelemetry.io/docs/collector/configuration/#processors

rynowak:
> This is really nice 😊. To help my understanding, in order for Radius to support monitoring applications that are rad deployed, an approach is for the application components to use open telemetry and generate observability signals of interest. 


I think you're on the right track so far. The application developer should produce the right telemetry signals like logs and metrics. The libraries for this are getting better all the time so this is pretty much expected. 


> We also would like to define thresholds at which ops might want to make sure nothing is wrong instance, perhaps a storage service reaches threshold, CPU/memory is 100% or http error code rate is higher. Once radius notices these thresholds are crossed, it can make the component yellow (and cascade color for dependent services)/ trigger an alarm suggesting check observability tools for the yellow service (or take other steps like increase log verbosity for a few seconds). Is this what we are aiming for at a high level?

If you're describing the **scenario** then I think this is right. Users can set up their own alerts and dashboards like we did when we set up Grafana for Radius.

If you're talking about the technology, it's not going to be **Radius** doing the monitoring. It's going to be external systems like Grafana or SaaS monitoring solutions like Splunk. Those are whole products/projects/businesses - we're not going to try and rebuild that functionality inside Radius.  

OpenTelemetry is the standard format, which is what makes all of this possible. All of these tools understand that format so they are swappable.

resources/2023-11-application-diagnostics-prototype.md
```diff
@@ -0,0 +1,215 @@
...
+
+In our demo setup we're going to use Azure Monitor + Grafana for both metrics, and logs. We're not addressing tracing in this demo. These technologies are good choices because Grafana is self-hostable and widely used, and Azure Monitor as a backend is a good subtitute for any OTEL-compatible source. It would be easy for us to move this demo to another cloud by swapping out Azure Monitor.
+
+#### Collecting and storing diagnostics data automatically for applications
+
+As a developer, I can instrument my code using the OpenTelemetry SDK to provide metrics. I can write logs to the console using the logging packages I prefer for my application. When I deploy my application, using Radius, the configuration for OpenTelemetry will be automatically injected. I don't know have to know about the configuration of, or have access to where the diagnostics data goes.
```
AaronCrawfis, line 65: Does OpenTelemetry have a standard environment variable or way to configure itself? Not familiar with how it's usually setup

rynowak: For the SDKs yes. This section refers to **both**:

- Setting the standard env-var that the SDKs understand
- Managing the standard collector (in one of the **many** ways to use the collector, with options for configuration)

resources/2023-11-application-diagnostics-prototype.md
rynowak: Yes that's what I meant 👍. This prototype is focused on the things that OTEL supports.

rynowak:
> 
> Understood. So the way we are enabling is by allowing the radius (recipe definitions) to make necessary configurations for the applications to hook up and send the signals to the tools of their choice. Correct?

Yes that's right. There's two components of this:

- Configuration provided to Recipes - eg: how/where should I send telemetry for this Redis Cache?
- Configuration injected into compute - eg: the environment variable that configures the OTEL SDK

resources/2023-11-application-diagnostics-prototype.md
```diff
@@ -0,0 +1,264 @@
...
+}
+
+// NEW TYPE
+model EnvironmentDashboard {
+  @doc("URL of the dashboard where diagnostics can be viewed.")
+  url: string
+}
+
+// NEW TYPE
+model EnvironmentAzureMonitorDiagnostics  {
```
vinayada1, line 205: Should this be something generic to work with non-Azure monitor configs as well? Or are we focussing on the specific scenario here

resources/2023-11-application-diagnostics-prototype.md
```diff
@@ -0,0 +1,264 @@
...
+  @doc("The resource ID of the Azure Monitor instance")
+  id: string
```
vinayada1, line 207: Could you explain how the resource ID will be used?

resources/2023-11-application-diagnostics-prototype.md
```diff
@@ -0,0 +1,264 @@
...
+
+*This design-note describes a prototype we plan to build for application diagnostics. As this document is describing a prototype, it will be less complete than other design proposals.*
+
+Our goal is to prototype an end-to-end scenario for **application** diagnostics that emcompasses:
+
+- A platform engineer configures a metrics dashboard (Prometheus+Grafana) and a logging dashboard (probably Azure to start) and wiring these up to a Radius environment.
+- A developer deploys an application into the environment, and has metrics and log collection automatically configured for the application.
+- An SRE can configure alerts for the metrics. When they receive an alert, they can use the metadata from the metrics to:
+  - Navigate to the application graph and understand the application architecture.
+  - Navigate to the logs and craft a query to understand the problem.
```
vinayada1, line 16: This is outside of Radius, right?

resources/2023-11-application-diagnostics-prototype.md
rynowak: I think that's the right idea once we do some prototyping. My idea was to figure out if this is a *good idea* before we do costly design work. Does that sound like the right plan?

resources/2023-11-application-diagnostics-prototype.md
rynowak: For now I was planning to use this for informational purposes. For example we could *deep-link* from the portal. This is part of my goal of "making the implicit explicit". 

Open to suggestions if you have other ideas 👍

resources/2023-11-application-diagnostics-prototype.md
rynowak: The query part is outside of Radius. I think we could help with the navigation.

rynowak:
I'm closing this as stale until someone can pick this up as a project.

PR #36, opened 2023-12-21
Add irregular types proposal

rynowak:


ucp/2023-12-irregular-types.md
```diff
@@ -0,0 +1,269 @@
...
+For the irregular types, the URL segment they provide (eg. `resourceGroups`) must match the unqualified resource type. 
+
+Importantly, we should standardize on the **regular** format for resource IDs. For the resource types that represent a **scope** we can also include the scope format in a new property for use in client code.
+
+**Looking for input on the right property name. I'm not sure scope is a great one.**
+
+```jsonc
+{
+  "id": "/planes/radius/local/providers/System.Resource/resourceGroups/default", // Changed
+  "scope": "/planes/radius/local/resourcegroups/default", // Added
```
AaronCrawfis, line 213: Would we also add the `scope` field to all resources? Such that when you do a GET on a container you get the `id` of the container (no change), and you also get the `scope` as well?

rynowak: This is probably the part of the proposal I am the most unsure about 😆 

It's nice today that you can combine the `id` of a scope (like a resource group) with other parts to make a full `id`.

Maybe this belongs in `properties` instead, and only for things that are scope-like.

rynowak:
Closing for now. I may bring this back in the future.

PR #37, opened 2024-01-09
Add design doc for supporting private terraform modules

vishwahiremat:
Doc : https://github.com/radius-project/design-notes/blob/addaf095cccfcb423396a5a447f47680a1f2a0e0/recipe/2024-01-support-private-terraform-repository.md


recipe/2024-01-support-private-terraform-repository.md
```diff
@@ -0,0 +1,237 @@
...
+
+> **Issue Reference:** https://github.com/radius-project/radius/issues/6911
+
+### Goals
+- Enable support to use terraform modules from private Git repositories.
+- It should support git from any platforms(like Bitbucket, Gitlab, Azure DevOps Git etc.)
+
+### Non goals
+- To Support other terraform module sources like S3, GCP, Mercurial repository.
+- Registering Terraform modules stored in repositories from different git accounts.
```
kachawla, line 32: Not sure if I understand this.. do you mean only supporting one git account per environment?

recipe/2024-01-support-private-terraform-repository.md
```diff
@@ -0,0 +1,237 @@
...
+
+Git provides different ways to authenticate:
+
+#### Personal Access Token:
+Users need to a Git personal access token with very limited access (just read access) and also specify token validity. And this is used along with the username to clone the terraform module repository through HTTPS as part of `terraform init`.
+
+#### SSH key:
+SSH key can be used to provide access to private git repository. But it requires adding the generated ssh key to the users account.
+
+#### Service Principle(Only for Azure DevOps Git):
```
shalabhms, line 50:
```suggestion
#### Service Principal (Only for Azure DevOps Git):
```

recipe/2024-01-support-private-terraform-repository.md
```diff
@@ -0,0 +1,237 @@
...
+We could use the Azure Service Principle details used for azure score to authenticate Azure DevOps Git. But most often users have diff tenant IDs for the production environment and for git.
```
shalabhms, line 51:
```suggestion
We could use the Azure Service Principal details used for azure score to authenticate Azure DevOps Git. But most often users have diff tenant IDs for the production environment and for git.
```

recipe/2024-01-support-private-terraform-repository.md
```diff
@@ -0,0 +1,237 @@
...
+        }
+      }
+    }
+  }
+}
+```
+But the majority of users frequently update their git personal access tokens (like once a day), and which means updating the environment resource every time the token is updated.
+
+
+Option 4 : Using Applications.Core/secretStores
```
ytimocin, line 137: nit:
```suggestion
Option 3 : Using Applications.Core/secretStores
```

recipe/2024-01-support-private-terraform-repository.md
```diff
@@ -0,0 +1,237 @@
...
+    }
+    providers: {
+      ...
+    }
++   recipeConfig: {
++     terraform: {
++       secrets: {
++         "dev.azure.com": {
++            username : username
++            pat: token
```
sk593, line 118: Are there potential security issues here if usernames and PAT are being passed through the bicep templates?

recipe/2024-01-support-private-terraform-repository.md
```diff
@@ -0,0 +1,237 @@
...
+    compute: {
+      ...
+    }
+    providers: {
+      ...
+    }
++   recipeConfig: {
++     terraform: {
++       secrets: {
++         "dev.azure.com": {
```
ytimocin, line 209: Are we going to do a mapping of template paths with the keys here? How will we know if the terraform is in a private repo?

recipe/2024-01-support-private-terraform-repository.md
```diff
@@ -0,0 +1,237 @@
...
+        value: '<username>'
+      }
+    }
+  }
+}
+
+```
+But todays secret store implementation is tied to application scope and in this case, secretStore needs to be created before application and environment creation. So we can only use the secretStore resource if we extend it to have global scope(extending to environment scope would create cyclic dependency between environment and secret store).
+
+Option 4 : Using kubernetes secret
```
kachawla, line 191: Have we considered leveraging our existing rad credential register API for recipe related credentials?

shalabhms: +1

recipe/2024-01-support-private-terraform-repository.md
```diff
@@ -0,0 +1,237 @@
...
+
+Using existing kubernetes secret i.e asking the users to have the git credentials already stored in the kubernetes secret on the same cluster and use it in the recipeConfig.
```
shalabhms, line 193: What is the option that you leaning to? Is it this K8s secret ?

recipe/2024-01-support-private-terraform-repository.md
youngbupark: UCP credential is designed for IAM credentials. Ideally, UCP credential should not be fetched by RP due to the security reason. For this scenario, using secretstore makes sense.

recipe/2024-01-support-private-terraform-repository.md
AaronCrawfis, line 137:
```suggestion
Option 3 : Using Applications.Core/secretStores
```

vishwahiremat:
Notes : 01/09/2024
- We decided that option 3 of using Application.Core/secretStores is the better approach to store git credentials but since today its application scoped we may need to make few changes to the secret store implementation.
- Next steps is to investigate on the design for secret store changes.
- Verify the public repo url as module source works while running terraform init on the cluster.

recipe/2024-01-support-private-terraform-repository.md
kachawla, line 191: nit

```suggestion
Option 5 : Using kubernetes secret
```

kachawla: > UCP credential is designed for IAM credentials.

I see. We don't need to use the UCP API if the credentials are only relevant for recipes RP. An extension of the existing command like `rad credential register recipe` instead `rad credential register azure` would provide users the convenience of handing off credentials management without additional mental load of an entirely different command. The backend API for this can be directly implemented in the Recipes RP.

> Ideally, UCP credential should not be fetched by RP due to the security reason. For this scenario, using secretstore makes sense.

Could you share more on the security concern?

> For this scenario, using secretstore makes sense.

Ideally using a secret store resource is easier for Cloud provider credentials (that we manage in UCP) management as well, but it doesn't bring easiness for users since it puts the responsibility of storage on them. So it depends on what we support.

recipe/2024-01-support-private-terraform-repository.md
```diff
@@ -0,0 +1,286 @@
...
+          templatePath: 'https://dev.azure.com/test-private-repo'
+        }
+      }
+    }
+  }
+}
+```
+But the majority of users frequently update their git personal access tokens (like once a day), and which means updating the environment resource every time the token is updated, also this raises potentials security issues storing the these details as part of environment. 
+
+Option 3 : Using kubernetes secret
```
lakshmimsft, line 123: nit: Option 2?

recipe/2024-01-support-private-terraform-repository.md
```diff
@@ -0,0 +1,286 @@
...
+    providers: {
+      ...
+    }
++   recipeConfig: {
++     terraform: {
++       secrets: {
++         "dev.azure.com": {
++            secret: secretStore.id
++            namespace: <namespace>
++          }
```
ytimocin, line 144: This is a map of secrets, right? I can add any number of secrets here for each registry.

ytimocin: I guess this is Option 2.

recipe/2024-01-support-private-terraform-repository.md
ytimocin: Same question applies here too: This is a map of secrets, right? I can add any number of secrets here for each registry.

recipe/2024-01-support-private-terraform-repository.md
```diff
@@ -0,0 +1,286 @@
...
+  providers?: Providers;
+
+  @doc("Simulated environment.")
+  simulated?: boolean;
+
+  @doc("Specifies Recipes linked to the Environment.")
+  recipes?: RecipeProperties;
+
++  @doc("Specifies recipe configurations needed for the recipes.")
++  recipeConfig?: Record<Record<RecipeConfigProperties>>;
```
ytimocin, line 244: Why is this double Record?

recipe/2024-01-support-private-terraform-repository.md
```diff
@@ -0,0 +1,286 @@
...
+
+  @doc("The environment extension.")
+  @extension("x-ms-identifiers", [])
+  extensions?: Array<Extension>;
+}
+
++model RecipeConfigProperties {
++  @doc(Specifies the secrets linked to a git platform)
++  terraform?: Record<Secrets>;
```
ytimocin, line 253: Are we going to have Bicep here in near future?

recipe/2024-01-support-private-terraform-repository.md
AaronCrawfis: If they're an input parameter to the Bicep template they will need to be marked/decorated `@secure()`

```bicep
@secure()
param token string
```

This tells Bicep/ARM not to log/store the value.

And then it will be up to the resource provider to securely store the value and make it a read-only API with a separate way to access its value. Same way we handle connection strings and passwords for portable resources.

recipe/2024-01-support-private-terraform-repository.md
```diff
@@ -0,0 +1,286 @@
...
+E.g:
+```
+templatePath: "git::https://{username}:{PERSONAL_ACCESS_TOKEN}@example.com.com/test-private-repo.git?ref=v1.2.0"
+```
+
+Since adding sensitive information like tokens to the terraform configuration files which may be stored in version control can pose security issues. So we could use different ways to store the github credentials.
+
+Option 1: Saving it as part of the environment resource.
+
+Add a new property `recipeConfig` write-only to store the information about the git credentials, and have a custom action `getRecipeConfiglike` (like `list-secrets`) to get these details.
```
rynowak, line 83: I know this isn't part of the actual proposal, but I wanted to comment anyway 😆 

We also don't have to provide a custom action for everything that's a secret. Custom actions should be rare because they are complex to use and require extra work on our part to build.

I'd suggest that the right principle is to be scenario-based. We have custom actions on standard resources because they are useful for passing settings into your code.

recipe/2024-01-support-private-terraform-repository.md
```diff
@@ -0,0 +1,286 @@
...
+  properties:{
+-  	app: app.id
+    type: 'generic'
+    data: {
+      'pat': {
+        value: '<personal-access-token>'
+      }
+      'username': {
+        value: '<username>'
+      }
```
kachawla, line 206: Could you add an example of how this would look like if a user wants to add a reference to an existing secret store (keyvault etc), instead of providing this here in plaintext?

recipe/2024-01-support-private-terraform-repository.md
```diff
@@ -0,0 +1,286 @@
...
+
+## Development plan
+- Task 1:  
+    - Update Application.Core/secretStores to be a global scoped resource.
+    - Update unit tests and validating it doesn't break the existing functionality.
+- Task 2:
+    - Adding a new property recipeConfig to environment and making necessary changes to typespec, datamodel and conversions.
+    - Updating unit tests.
+- Task 3:
+    - Adding environment controller changes.
```
ytimocin, line 280: Are we going to need any changes in the controller loop (reconciler)?

recipe/2024-01-support-private-terraform-repository.md
rynowak, line 244: What's the purpose of the two levels of `Record<>` here? This doesn't seem to match the example.

recipe/2024-01-support-private-terraform-repository.md
```diff
@@ -0,0 +1,286 @@
...
+    compute: {
+      ...
+    }
+    providers: {
+      ...
+    }
++   recipeConfig: {
++     terraform: {
++       secrets: {
++         "dev.azure.com": {
```
kachawla, line 178: Where are the keys provided here `dev.azure.com` consumed?

recipe/2024-01-support-private-terraform-repository.md
```diff
@@ -0,0 +1,286 @@
...
+      'username': {
+        value: '<username>'
+      }
+    }
+  }
+}
+
+```
+SecretStore resource also provides an option to use the existing secret, which makes it better way store credentials. But todays secret store implementation is tied to application scope and in this case, secretStore needs to be created before application and environment creation. So we need to change scope of secret store resource to global (by removing the required flag for application property).
+And the referenced secret is retrieved from the secretStore in configLoader and secret details i.e personal-access-token and username as used to modify the template path to this format:
```
sk593, line 213: Why is the secretStore created before the app and env?

recipe/2024-01-support-private-terraform-repository.md
kachawla: If we are not supporting that scenario yet, then let's call that out

recipe/2024-01-support-private-terraform-repository.md
```diff
@@ -0,0 +1,286 @@
...
+
+
+## Test plan
+#### Unit Tests
+-   Update environment conversion unit tests to validate recipeConfig property.
+-   Update environment controller unit tests.
+-   AAdding new unit tests in recipe config loader to validate recipeConfig changes.
+
+#### Functional tests
+- 	Add e2e test to verify recipe deployment using a terraform module stored in a private git repository.
```
rynowak, line 270: How will we build that? Can you get in more detail about how this will work?

recipe/2024-01-support-private-terraform-repository.md
rynowak: How would this design change if we wanted to support service principals in the future? Or SSH keys?

recipe/2024-01-support-private-terraform-repository.md
AaronCrawfis: Nothing concrete at the top of our backlog yet, but anything from here could potentially be added in the future: https://learn.microsoft.com/en-us/azure/azure-resource-manager/bicep/bicep-config

recipe/2024-01-support-private-terraform-repository.md
```diff
@@ -0,0 +1,286 @@
...
+
+## Development plan
```
kachawla, line 272: For the updates needed to consume this config - are they tracked in one of the tasks here?

recipe/2024-01-support-private-terraform-repository.md
```diff
@@ -0,0 +1,286 @@
...
+        default: {
+          templateKind: 'terraform'
+          templatePath: 'https://dev.azure.com/test-private-repo'
+        }
+      }
+    }
+  }
+}
+```
+But the majority of users frequently update their git personal access tokens (like once a day), and which means updating the environment resource every time the token is updated, also this raises potentials security issues storing the these details as part of environment.
```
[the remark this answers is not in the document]
rynowak: We should, however talk about what you're actually saying about secret handling. 

There are two *general* approaches we could take:

- Resource bodies may contain secrets via write-only fields.
- Resource bodies *may not* container secrets. They always contain references to secrets, and secrets are stored somewhere else. 

If there's time I'd like to talk about this as a general topic. Ideally all of our APIs would work the same.

recipe/2024-01-support-private-terraform-repository.md
```diff
@@ -0,0 +1,318 @@
...
+}
+
++model RecipeConfigProperties {
++  @doc(Specifies the terraform config properties)
++  terraform?: Record<TerraformConfig>;
++}
+
++model TerraformConfig{
++  @doc(Specifies git credentials information for terraform module repository.)  
++  gitCredentials?: Record<Secret>
```
kachawla, line 277: Could you share an example of how this will evolve as we add support for sources other than git in the future?

recipe/2024-01-support-private-terraform-repository.md
```diff
@@ -0,0 +1,318 @@
...
++  recipeConfig?: RecipeConfigProperties;
+
+  @doc("The environment extension.")
+  @extension("x-ms-identifiers", [])
+  extensions?: Array<Extension>;
+}
+
++model RecipeConfigProperties {
++  @doc(Specifies the terraform config properties)
++  terraform?: Record<TerraformConfig>;
```
rynowak, line 272: What's the key of this `Record<>`? 

```json
{
  "properties": {
    "recipeConfig": {
      "terraform": {
        "???": {
          gitCredentials: {
            "???": {
              secret: "I'm afraid of spiders!"
            }
          }
        }
      }
    }
  }
}
```

recipe/2024-01-support-private-terraform-repository.md
```diff
@@ -0,0 +1,318 @@
...
++  secret?: string;
++}
+```
+In future if we want to support other types of authentication(e.g ssh), `gitCredentials` would include another level of properties to specify the kind of authentication used.
+
+```
+model TerraformConfig{
+  @doc(Specifies git credentials information for terraform module repository.)  
+  gitCredentials?: Record<Record<Secret>>
+}
```
rynowak, line 291: This doesn't sound right. We anticipate that other kinds of auth will be needed in the future, so we can plan now to avoid a breaking change to the API.

recipe/2024-01-support-private-terraform-repository.md
vishwahiremat: here key would be domain names of the git platform e.g "dev.azure.com", "github.com"

recipe/2024-01-support-private-terraform-repository.md
```diff
@@ -0,0 +1,318 @@
...
+      }
+    }
+  }
+}
+
+resource secretStore 'Applications.Core/secretStores@2023-10-01-preview' = {
+  name: 'github'
+  properties:{
+-  	app: app.id
+    type: 'generic'
```
youngbupark, line 202: In the future, we may have to support generator secret store type. See [Acorn's generated secrets](https://docs.acorn.io/authoring/secrets#generated-secrets). cc/ @rynowak

recipe/2024-01-support-private-terraform-repository.md
vishwahiremat: for other module sources we add a new config property  e.g mercurial: 
```
model TerraformConfig{
   @doc(Specifies git credentials information for terraform module repository.)  
   gitCredentials?: Record<Secret>
   mercurialCredentials? Record<Secret>
```

recipe/2024-01-support-private-terraform-repository.md
youngbupark: +1 having additional POST action is not useful unless you model recipeconfig as separate resource model.

recipe/2024-01-support-private-terraform-repository.md
```diff
@@ -0,0 +1,318 @@
...
+
++model RecipeConfigProperties {
++  @doc(Specifies the terraform config properties)
++  terraform?: Record<TerraformConfig>;
++}
+
++model TerraformConfig{
++  @doc(Specifies git credentials information for terraform module repository.)  
++  gitCredentials?: Record<Secret>
++}
```
lakshmimsft, line 278: fmi: Could you describe details on validating user input for credentials?

recipe/2024-01-support-private-terraform-repository.md
```diff
@@ -0,0 +1,328 @@
...
+## Objectives
+
+> **Issue Reference:** https://github.com/radius-project/radius/issues/6911
+
+### Goals
+- Enable support to use terraform modules from private Git repositories.
+- It should support git from any platforms(like Bitbucket, Gitlab, Azure DevOps Git etc.)
+
+### Non goals
+- To Support other terraform module sources like S3, GCP, Mercurial repository.
```
kachawla, line 31:
```suggestion
- To Support other terraform module sources like S3, GCP, Mercurial repository. Support for other sources will be implemented in future iterations.
```

recipe/2024-01-support-private-terraform-repository.md
kachawla: Can you please update this non-goal to be more explicit.. support for multiple git accounts per Radius environment?

recipe/2024-01-support-private-terraform-repository.md
```diff
@@ -0,0 +1,328 @@
...
+#### Personal Access Token:
+Users need to a Git personal access token with very limited access (just read access) and also specify token validity. And this is used along with the username to clone the terraform module repository through HTTPS as part of `terraform init`.
+
+#### SSH key:
+SSH key can be used to provide access to private git repository. But it requires adding the generated ssh key to the users account.
+
+#### Service Principal(Only for Azure DevOps Git):
+We could use the Azure Service Principal details used for Azure scope to authenticate Azure DevOps Git. But most often users have diff tenant IDs for the production environment and for git.
+
+Personal Access Token with Username way of authentication is supported by git from most of the platforms(e.g Bitbucket,Gitlab,Github etc) and generic git repository URL format as template path can be used to while registering recipe to access the terraform modules from private git repository from any platform.
```
kachawla, line 53:
```suggestion
**Proposed Option**
Personal Access Token with Username way of authentication is supported by git from most of the platforms(e.g Bitbucket,Gitlab,Github etc) and generic git repository URL format as template path can be used to while registering recipe to access the terraform modules from private git repository from any platform.
```

recipe/2024-01-support-private-terraform-repository.md
```diff
@@ -0,0 +1,237 @@
...
+- To Support other terraform module sources like S3, GCP, Mercurial repository.
+- Registering Terraform modules stored in repositories from different git accounts.
+
+### User scenarios (optional)
+
+As an operator I am responsible for maintaining Terraform recipes for use with Radius. Terraform modules used contains sensitive information, proprietary configurations, or data that should not be shared publicly and its intended for internal use within our organization and have granular control over who can access, contribute to, or modify Terraform modules. So I store the terraform modules in a private git repository. And I would like to use these github sources from a private repository as template-path while registering a terraform recipe.
+
+## Design
+### Design details
+Today, we support only Terraform registry and HTTP URLs as allowed module sources for Terraform recipe template paths. Git public repository are also supported by providing the HTTP URl to the module directory. But to support repositories we need a way to authenticate the git account where the modules are stored.
```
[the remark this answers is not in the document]
kachawla: This was confirmed to be supported.

recipe/2024-01-support-private-terraform-repository.md
```diff
@@ -0,0 +1,350 @@
...
+| Private Terraform Repository | A private Terraform repository typically refers to a version control repository that contains Terraform module code, but is not publicly accessible. | 
+
+
+## Objectives
+
+> **Issue Reference:** https://github.com/radius-project/radius/issues/6911
+
+### Goals
+- Enable support to use terraform modules from private Git repositories.
+- It should support git from any platforms(like Bitbucket, Gitlab, Azure DevOps Git etc.)
```
kachawla, line 28:
```suggestion
- It should support git from any platforms (like Bitbucket, Gitlab, Azure DevOps Git etc.)
```

kachawla (approved):

PR #38, opened 2024-01-23
Adding design doc for  to extend the capabilities of secretStores to a global scope

vishwahiremat:
Doc: https://github.com/radius-project/design-notes/blob/70894efa5cfc2ba22d445da23e0d7325bc4c1883/recipe/2024-01-global-scope-secret-store.md

### DIFF
--- a/recipe/2024-01-global-scope-secret-store.md
+++ b/recipe/2024-01-global-scope-secret-store.md
@@ -1,6 +1,6 @@
 # Adding support to extend the capabilities of secretStores to a global scope.
 
-* **Status**: Pending/Approved
+* **Status**: Pending
 * **Author**: Vishwanath Hiremath (@vishwahiremat)
 
 ## Overview
@@ -43,15 +43,17 @@ resource env 'Applications.Core/environments@2023-10-01-preview' = {
       ...
     }
    recipeConfig: {
-     terraform: {
-       gitCredentials: {
-         "dev.azure.com": {
-+            secret: secretStoreAzure.id
-          }
-          "github.com": {
-+            secret: secretStoreGithub.id
-          }
-       }
+      terraform: {
+        authentication:{
+          pat: {
+            "dev.azure.com": {
++             secret: secretStoreAzure.id
+            }
+            "github.com": {
++              secret: secretStoreGithub.id
+            } 
+          } 
+        }
      }
    }
     recipes: {      
@@ -68,7 +70,7 @@ resource env 'Applications.Core/environments@2023-10-01-preview' = {
 
 
 ## Design
-Today we use Application.Core/secretstores resource to store secrets/credentials information for an application but we cannot use it here as secretStores is application scoped and expects to have an application created before we create a secretStore.
+Today we use Application.Core/secretstores resource to store secrets/credentials information for an application but we cannot use it in this scenario as secretStores is application scoped and expects to have an application created before we create a secretStore.
 
 And we cannot have secretStore created for environment scope in this scenario as we are adding the secret to the environment which creates cyclic dependency between environment and secretStore resource.
 
@@ -113,9 +115,7 @@ We could make secretStores as a global scoped resource by removing application a
       ]
     },
 ```
-With this change we can create a secret store resource before an application or environment is created.
-
-If user is creating a secretStore of kind kubernetes secret, it is deployed in application/environment namespace depending on the scope. But for global scope we can use the `resource` property (which is used to specify the secret ref of the existing secret) to provide namespace and secret name details in this format `<namespace>/<secretname>` . If the secretstore is created with global scope then we expect user to provide `resource` details.
+With this change we can create a secret store resource before an application or environment is created. But, if the user is creating a secretStore of kind kubernetes secret, it is deployed in application/environment namespace. But for secret resource with global scope use the `resource` property which is used to specify the secret ref of the existing secret to provide namespace and secret name details. User is expected to provide these details in `<namespace>/<secretName>` format for secretStore with global scope.
 
 ```diff
 resource secretStore 'Applications.Core/secretStores@2023-10-01-preview' = {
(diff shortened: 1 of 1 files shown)


recipe/2024-01-global-scope-secret-store.md
```diff
@@ -0,0 +1,202 @@
+# Adding support to extend the capabilities of secretStores to a global scope.
+
+* **Status**: Pending/Approved
+* **Author**: Vishwanath Hiremath (@vishwahiremat)
+
+## Overview
+
+Today we have Application.Core/secretStores resource to securely manage secrets for the Application. However a gap exists if user wants to create a secret before application or environment is created. To address this limitation, we need to enhance the capabilities of the existing secretStores resource. The objective is to extend its support to a global scope, enabling users to store secrets prior to the creation of the application or environment.
```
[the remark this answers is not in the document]
vishwahiremat: By asking the users to provide `resource` property in the format of <namespace>/<secretName> if secret stores is a global scoped.

https://github.com/radius-project/design-notes/pull/38/files#diff-89d84238f05b572cd71ded0059391247faab67f69d3d9232b31350325b946cfaR125:~:text=%2B%20%20%20resource%3A%20%3Cnamespace%3E/%3C-,secretName,-%3E

recipe/2024-01-global-scope-secret-store.md
```diff
@@ -0,0 +1,259 @@
...
+    },
+```
+With this change we can create a secret store resource before an application or environment is created. But, if the user is creating a secretStore of kind kubernetes secret, it is deployed in application/environment namespace. But for secret resource with global scope use the `resource` property which is used to specify the secret ref of the existing secret to provide namespace and secret name details. User is expected to provide these details in `<namespace>/<secretName>` format for secretStore with global scope.
+
+```diff
+resource secretStore 'Applications.Core/secretStores@2023-10-01-preview' = {
+  name: 'github'
+  properties:{
+-  	app: app.id
++   resource: <namespace>/<secretName>
```
kachawla, line 125: Can this namespace be anything or it has to be match the environment namespace (when the environment gets created)?

recipe/2024-01-global-scope-secret-store.md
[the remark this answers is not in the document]
rynowak: I'm not sure if we're talking about the same thing. Let me try being more specific.

If I create an `Applications.Core/secretStores` resource today and I **don't** specify the kubernetes secret name + namespace, then it will use the namespace of the application (or environment).

For the global secret case, there's no application or environment, so there's no way to look up the namespace we're going to use.

recipe/2024-01-global-scope-secret-store.md
youngbupark, line 125: Alternatively, we might be able to use UCP style resource id instead of using `<namespace>/<secretName>`

e.g.

`/planes/kubernetes/local/namespaces/<namespace>/providers/cores/Secrets/<secretName>"`

recipe/2024-01-global-scope-secret-store.md
```diff
@@ -0,0 +1,259 @@
...
+| Private Terraform Repository | A private Terraform repository typically refers to a version control repository that contains Terraform module code, but is not publicly accessible. | 
+
+## Objectives
+
+> **Issue Reference:** https://github.com/radius-project/radius/issues/7030
+
+### Goals
+
+- Enable support to extend the capabilities of secretStores to a global scope.
+
```
rynowak, line 25: Something I am curious about is whether this type should move to UCP. `Applications.Core` defines it as a part of the Radius RP rather than part of UCP.

rynowak: If we have usecases for this type in other parts of UCP then it would be better to rename/move than it would be to have two different secret types.

recipe/2024-01-global-scope-secret-store.md
```diff
@@ -0,0 +1,259 @@
...
+  @doc("The resource id of external secret store.")
+  resource?: string;
+}
+```
+
+## Alternatives considered
+
+#### Adding default a namespace for global scoped secretstore resource.
+Add a default namespace `global-secretStores` to store the global scoped secretStores.
```
rynowak, line 186: Or we could use `radius-system`.

recipe/2024-01-global-scope-secret-store.md
vishwahiremat: it can be anything

recipe/2024-01-global-scope-secret-store.md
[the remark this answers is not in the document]
rynowak: From reading more of the doc, it looks like you did try to address this feedback as part of the proposal. Let's talk about it 👍

recipe/2024-01-global-scope-secret-store.md
rynowak, line 125: Let's talk about the semantics of this field.

youngbupark: In the initial proposal of secretstore, we proposed using full resource id for Azure KeyVault in `resource` field when it refers to KeyVault Resource.

recipe/2024-01-global-scope-secret-store.md
```diff
@@ -0,0 +1,259 @@
...
++}
+```
+
+typespec/Applications.Core/secretstores.tsp
+
+```diff
+@doc("The properties of SecretStore")
+model SecretStoreProperties {
+-  ...ApplicationScopedResource;
++  ...GlobalScopedResource;
```
lakshmimsft, line 169: since this is new type, pls describe it in detail

recipe/2024-01-global-scope-secret-store.md
```diff
@@ -0,0 +1,261 @@
...
+
+## Objectives
+
+> **Issue Reference:** https://github.com/radius-project/radius/issues/7030
+
+### Goals
+
+- Enable support to extend the capabilities of secretStores to a global scope.
+
+### Non goals
```
kachawla, line 26: @vishwahiremat is https://github.com/radius-project/design-notes/pull/38/files#r1466964211 a non goal for the scope of this document? If so, let's please add it here with the reasoning why.

recipe/2024-01-global-scope-secret-store.md
```diff
@@ -0,0 +1,261 @@
...
+    },
+```
+With this change we can create a secret store resource before an application or environment is created. But, if the user is creating a secretStore of kind kubernetes secret, it is deployed in application/environment namespace. But for secret resource with global scope use the `resource` property which is used to specify the secret ref of the existing secret to provide namespace and secret name details. User is expected to provide these details in `<namespace>/<secretName>` format for secretStore with global scope.
+
+```diff
+resource secretStore 'Applications.Core/secretStores@2023-10-01-preview' = {
+  name: 'github'
+  properties:{
+-  	app: app.id
++   resource: <namespace>/<secretName>
```
kachawla, line 127: @vishwahiremat could you confirm if it was decided to go with `<namespace>/<secretName>` or something more generic that would work with other resource types as well as mentioned in this comment https://github.com/radius-project/design-notes/pull/38/files#r1466963848? Would be great if you could add a quick note in the design on how this would extend for other resource types in the future.

recipe/2024-01-global-scope-secret-store.md
```diff
@@ -0,0 +1,261 @@
...
+      }
+      'username': {
+        value: '<username>'
+      }
+    }
+  }
+}
+```
+
+### API design (if applicable)
```
kachawla, line 141:
```suggestion
### API design
```

recipe/2024-01-global-scope-secret-store.md
vishwahiremat: thats correct, its a non goal for this task. Will add it as part of the doc

recipe/2024-01-global-scope-secret-store.md
vishwahiremat: we decided to go with <namespace>/<secretname> format. and I will update it

kachawla (approved):

recipe/2024-01-global-scope-secret-store.md
```diff
@@ -0,0 +1,261 @@
...
+## Objectives
+
+> **Issue Reference:** https://github.com/radius-project/radius/issues/7030
+
+### Goals
+
+- Enable support to extend the capabilities of secretStores to a global scope.
+
+### Non goals
+- To move secret store resource as part of UCP.
```
kachawla, line 27: @vishwahiremat Please log an issue to revisit this when you get a chance

PR #39, opened 2024-02-07
Add design document to support multiple Terraform Providers

lakshmimsft:


recipe/2024-02-terraform-providers.md
```diff
@@ -0,0 +1,196 @@
...
+
+The system will allow for ability to set up multiple configurations for the same provider using the keyword *alias*. Validation for configuration 'correctness' will be handled by Terraform with calls to terraform init and terraform apply.
+
+```
+...
+ recipeConfig: {
+    terraform: {
+      ...
+      providers: {
+        [azurerm: {
```
rynowak, line 89: Is this meant to be a list? or a map?. The syntax currently looks like it's both/neither. `[]` is used for a list, and `{ key: <value> }` is for maps.

recipe/2024-02-terraform-providers.md
```diff
@@ -0,0 +1,196 @@
...
+        subscriptionid: 1234
+        alias: 'az-paymentservice'
+        secrets: {
+          source: secretStoreAzPayment.id
+        }         
+      }]
+    } 
+...       
+```
+Configuration for providers as described in this document will take precedence over provider credentials stored in UCP (currently these include azurerm, aws, kubernetes providers). So, for eg. In the scenario where credentials for Azure are saved with UCP during Radius install and a Terraform recipe created by an operator declares 'azurerm' under the the *required_providers* block; If there exists a provider configuration under the *providers* block under *recipeConfig*, these would take precedence and used to build the Terraform configuration file instead of Azure credentials stored in UCP.
```
rynowak, line 105: Glad you considered this 👍

recipe/2024-02-terraform-providers.md
```diff
@@ -0,0 +1,196 @@
...
+Containerization work may alleviate this limitations. Further design work will be needed for towards this which is planned for the near future. 
+
+### API design (if applicable) (TBD)
+
+TBD
+
+## Alternatives considered
+
+Mentioned under Limitations above, work to containerize running terraform jobs was considered as a precursor to this work. The time sensitivity for unblocking customers on ability to configure and use providers they use today was given a priority and current design held. Containerization will be taken up as a parallel effort in the near future.
+
```
rynowak, line 175: Do we need a way to specific configuration with specific recipes?

recipe/2024-02-terraform-providers.md
```diff
@@ -0,0 +1,190 @@
...
+**Reference for new type recipeConfig and handling of secrets:** [Design document for Private Terraform Repository](https://github.com/radius-project/design-notes/blob/3644b754152edc97e641f537b20cf3d87a386c43/recipe/2024-01-support-private-terraform-repository.md)
+
+> **Issue Reference:** <!-- (If appropriate) Reference an existing issue that describes the feature or bug. -->
+https://github.com/radius-project/radius/issues/6539
+
+### Goals
+
+1. Enable users to use terraform recipes with multiple providers (including and outside of Azure, Kubernetes, AWS).
+2. Enable users to use terraform recipes with multiple configurations for the same provider.
```
AaronCrawfis, line 28: Does this mean TF aliases?

recipe/2024-02-terraform-providers.md
```diff
@@ -0,0 +1,190 @@
...
+```
+...
+ recipeConfig: {
+    terraform: {
+      ...
+     providers: [...]
+     env: {
+       key: value
+       secrets: {
+        source: secretStoreOCICred.id
```
AaronCrawfis, line 71: How do we address a specific secret inside of the secret store? Is it the same as the value of `key`? What if you want to set a key to a secret not named the same as the value of key?

recipe/2024-02-terraform-providers.md
```diff
@@ -0,0 +1,190 @@
...
+          }
+        },
+        'azurerm': {
+        subscriptionid: 1234
+        alias: 'az-paymentservice'
+        secrets: {
+          source: secretStoreAzPayment.id
+        }         
+      }]
+...
```
AaronCrawfis, line 99: Is this allowed? What would happen if I tried accessing `env.properties.recipeConfig.terraform.providers['azurerm']`?

recipe/2024-02-terraform-providers.md
```diff
@@ -0,0 +1,190 @@
...
+
+1. Enable users to use terraform recipes with multiple providers (including and outside of Azure, Kubernetes, AWS).
+2. Enable users to use terraform recipes with multiple configurations for the same provider. 
+3. Enable provider configurations to be available for all recipies used in an environment.
+
+
+### Non goals
+1. Updates to Bicep Provider configuration.
+2. Authentication for providers hosted in private registries.
+3. Support for .terraformrc files.
```
rynowak, line 35: Can you say more about why (what is a `.terraformrc` file?)

recipe/2024-02-terraform-providers.md
```diff
@@ -0,0 +1,190 @@
...
+
+Largely following security considerations and secret handling described in design for private terraform modules: [Design document for Private Terraform Repository](https://github.com/radius-project/design-notes/blob/3644b754152edc97e641f537b20cf3d87a386c43/recipe/2024-01-support-private-terraform-repository.md)
+The work done here will be to read, decipher values and secrets and environment variables set by user in the *providers* block, to build internal terraform configuration file used by tf setenv, init and apply commands. 
+
+## Monitoring
+
+## Development plan (TBD)
+
+## Open issues/questions
+Do we consider first class support for GCP and other popular providers or continue with a generic approach??
```
AaronCrawfis, line 190: We should start with a generic approach to unlock everything, and then add first-class support where users request it later on. That we support everything, and then we can add convenience later on.

recipe/2024-02-terraform-providers.md
```diff
@@ -0,0 +1,190 @@
...
+     providers: [...]
+     env: {
+       key: value
+       secrets: {
+        source: secretStoreOCICred.id
+      }
+     }
+```
+
+Environment variables apply to all providers configured in the environment. The system cannot set two separate values for the same environment variable for multple provider configurations. In such cases, per provider documentation, there may be alternatives that users can avail (eg. For GCP, users can set  *credentials* field inside each instance of provider config as opposed to using env variable GOOGLE_APPLICATION_CREDENTIALS).
```
rynowak, line 76: Would we recommend that users avoid using env-vars in favor of other techniques?

recipe/2024-02-terraform-providers.md
```diff
@@ -0,0 +1,190 @@
...
+
+
+The system will allow for ability to set up multiple configurations for the same provider using the keyword *alias*. Validation for configuration 'correctness' will be handled by Terraform with calls to terraform init and terraform apply.
+
+```
+...
+ recipeConfig: {
+    terraform: {
+      ...
+      providers: ['azurerm': {
```
rynowak, line 86: Can you clarify whether this is a a list `[<value> ...]` or a map `{ key: <value> ... }`

It feels like it wants to be both.

rynowak: Do you have a diff of the typespec?

recipe/2024-02-terraform-providers.md
```diff
@@ -0,0 +1,190 @@
...
+### Goals
+
+1. Enable users to use terraform recipes with multiple providers (including and outside of Azure, Kubernetes, AWS).
+2. Enable users to use terraform recipes with multiple configurations for the same provider. 
+3. Enable provider configurations to be available for all recipies used in an environment.
+
+
+### Non goals
+1. Updates to Bicep Provider configuration.
+2. Authentication for providers hosted in private registries.
```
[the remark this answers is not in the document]
lakshmimsft: discussed in mtg, not part of scope for this effort.

recipe/2024-02-terraform-providers.md
lakshmimsft: https://developer.hashicorp.com/terraform/cli/config/config-file

recipe/2024-02-terraform-providers.md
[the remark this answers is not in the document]
lakshmimsft: that is correct. options of allowed/possible inputs are provider specific.

recipe/2024-02-terraform-providers.md
lakshmimsft: As discussed, this was considered and we've went ahead with first targetting setting up configuration at environment level before getting more granular at the individual recipe level.

recipe/2024-02-terraform-providers.md
```diff
@@ -0,0 +1,399 @@
...
+
+```
+...
+ recipeConfig: {
+    terraform: {
+      ...
+     providers: [...]
+     env: {
+      key: value
+      envSecrets: [
```
AaronCrawfis, line 70: If we're adding a `valueFrom` property below, where the user can specify a secret store, does the user also have to specify that this is a secret under `envSecrets`? Put another way, what if `valueFrom` was directly underneath `env`?

Also, what is the `key` on line 69? Is that the key for the environment variable? How does that relate to the environment variable key that appears to be on lines 72 and 80?

recipe/2024-02-terraform-providers.md
```diff
@@ -0,0 +1,401 @@
...
+
+**Reference for new type recipeConfig and handling of secrets:** [Design document for Private Terraform Repository](https://github.com/radius-project/design-notes/blob/3644b754152edc97e641f537b20cf3d87a386c43/recipe/2024-01-support-private-terraform-repository.md)
+
+> **Issue Reference:** <!-- (If appropriate) Reference an existing issue that describes the feature or bug. -->
+https://github.com/radius-project/radius/issues/6539
+
+### Goals
+
+1. Enable users to use terraform recipes with multiple providers (including and outside of Azure, Kubernetes, AWS).
+2. Enable users to use terraform recipes with multiple configurations for the same provider using *alias*.
```
kachawla, line 28: `same provider using *alias*` - I think this is a good candidate for `Terms and Definitions` section. Those who haven't spent much time with Terraform likely aren't familiar with this.

recipe/2024-02-terraform-providers.md
```diff
@@ -0,0 +1,401 @@
...
+
+### Goals
+
+1. Enable users to use terraform recipes with multiple providers (including and outside of Azure, Kubernetes, AWS).
+2. Enable users to use terraform recipes with multiple configurations for the same provider using *alias*. 
+3. Enable provider configurations to be available for all recipies used in an environment.
+
+
+### Non goals
+1. Updates to Bicep Provider configuration.
```
kachawla, line 33: Could you add more details on this, what do we mean by bicep provider and why this is a non goal? Is an equivalent feature possible bicep that's not in scope?

recipe/2024-02-terraform-providers.md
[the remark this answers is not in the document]
kachawla: @lakshmimsft can you update the doc to provide more context on why this is out of scope. Others reading the doc in the future may have same question as Ryan.

recipe/2024-02-terraform-providers.md
kachawla: Let's update the doc to include this reference, we don't expect people to come back and look at comments once this is merged :). Also, let's add why this is a non goal.

recipe/2024-02-terraform-providers.md
```diff
@@ -0,0 +1,401 @@
...
+#### User story 2
+As an operator, I would like to manage cloud resources in different geographical regions. To enable this, I need to set up multiple configurations for the same provider for different regions and use an alias to refer to individual provider configurations. 
+
+## Design
+### Design details
+
+Reviewing popular provider configurations (GCP, Oracle, Heroku, DigitalOcean, Docker etc.) a large number of provider configurations can be setup by handling a combination of key value pairs, key value pairs in nested objects, secrets and environment variables.
+
+### Key-value pairs
+Key-value pairs will be parsed and saved to a Terraform configuration file.
```
kachawla, line 52: I think this can be a bit confusing.. Let's add more details about which key value pairs and which Terraform configuration file.

recipe/2024-02-terraform-providers.md
```diff
@@ -0,0 +1,401 @@
...
+          }
+        },
+        {
+          name: 'azurerm',
+          properties: {
+            subscriptionid: 1234,
+            alias: 'az-paymentservice',
+            secrets: {
+              source: secretStoreAzPayment.id
+            }
```
ytimocin, line 123: Is this map going to have any property other than **source**?

recipe/2024-02-terraform-providers.md
```diff
@@ -0,0 +1,401 @@
...
+...
+ recipeConfig: {
+    terraform: {
+      ...
+     providers: [...]
+     env: {
+      key: value
+      envSecrets: [
+        {
+          name: 'SPECIAL_LEVEL_KEY'
```
kachawla, line 72: What is `SPECIAL_LEVEL_KEY`?

recipe/2024-02-terraform-providers.md
```diff
@@ -0,0 +1,401 @@
...
+          valueFrom: {
+            source: secretStoreConfig.id
+            name: 'special-config'
+            key: 'special.level'
```
kachawla, line 76: Could you share details on how are `name` and `key` consumed? Why do we need these?

recipe/2024-02-terraform-providers.md
```diff
@@ -0,0 +1,401 @@
...
+        {
+          name: 'LOG_LEVEL'
+          valueFrom: {
+            source: secretStoreConfig.id
+            name: 'special-config'
+            key: 'special.log'
+          }
+        }
+      ]
+      envSecretsFrom: {                   // All Secrets from Secret Store
```
kachawla, line 88: What's the motivation for adding this another layer of `envSecretsFrom` and `envSecrets`?

recipe/2024-02-terraform-providers.md
```diff
@@ -0,0 +1,401 @@
...
++          {
++             name: 'LOG_LEVEL'
++             valueFrom: {
++               source: secretStoreConfig.id
++               name: 'special-config'
++               key: 'special.log'
++             }
++          }
++        ]
++        envSecretsFrom: {               // All Secrets from Secret Store
```
ytimocin, line 205: Is there a way to merge `envSecretsFrom` and `envSecrets`? Like JavaScript does with the spread operator. Not that important but having two keys (`envSecretsFrom` and `envSecrets`) that look a lot like each other could create confusion.

recipe/2024-02-terraform-providers.md
```diff
@@ -0,0 +1,401 @@
...
+
+### Option 1:
+```
+Addition of new property to TerraformConfigProperties in `recipeConfig` under environment properties.
+
+model TerraformConfigProperties{
+  @doc(Specifies authentication information needed to use private terraform module repositories.)  
+  authentication?: AuthConfig
++ providers?: Array<ProviderConfig>
+}
```
ytimocin, line 314: Should we make this more generic so that we can use this for Bicep later too?

recipe/2024-02-terraform-providers.md
lakshmimsft: Line 187 is more helpful to understand. I need to update the description. will do soon.
Current proposal is to provide 3 ways of setting environment variables:
1. the key would be name of env var and value is the value set (non sensitive values)
2. envSecrets : Individual Secrets from SecretStore
3. envSecretsFrom: All Secrets from Secret Store
All three are optional.

ytimocin (commented):
More general question: Are these configurations be injected to the Terraform recipe or do we need to do some changes in the Terraform Driver too? cc/ @kachawla

recipe/2024-02-terraform-providers.md
```diff
@@ -0,0 +1,401 @@
...
+  @doc("Specifies the terraform config properties")
+  terraform?: TerraformConfigProperties;
++ env?: Env
+}
+
+@doc("Env describes structure enabling environment variables to be set")
+model Env extends Record<string>{
+  envSecrets?: Array<EnvSecret>
+  envSecretsFrom?: Secret
+}
```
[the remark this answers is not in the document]
rynowak: I think a more straightforward design might be like this:

```
env: {
  MY_ENV_VAR: 'my non-secret value'
}
envSecrets: {
  MY_SECRET_VAR: {
    secretStore: 'mysecretstore'
    key: 'mysecretkey'
  }
}
envSecretsFrom: {
  secretStore: 'myothersecretstore';
}
```

recipe/2024-02-terraform-providers.md
```diff
@@ -0,0 +1,401 @@
...
+          valueFrom: {
+            source: secretStoreConfig.id
+            name: 'special-config'
+            key: 'special.log'
+          }
+        }
+      ]
+      envSecretsFrom: {                   // All Secrets from Secret Store
+        secrets: {                        // This exists:Private TF module
+          source: secretStoreOCICred.id
```
kachawla, line 90: We should also discuss development plan for this. I think we can break it down in different iterations, with first iteration being the simplest scenario where fetching env variables from secret store isn't supported.

recipe/2024-02-terraform-providers.md
```diff
@@ -0,0 +1,401 @@
...
++              database_endpoint: "...",
++              java_endpoint: "...",
++              mysql_endpoint: "..."
++            }
++          ]
++        }
++     }
++     env: {
++       key: value
++       envSecrets: [                     // Individual Secrets from SecretStore
```
kachawla, line 266: Can it just be `Secrets`? `env` part is a given since it is inside the env block.

```suggestion
+       secrets: [                     // Individual Secrets from SecretStore
```

kachawla:
> More general question: Are these configurations be injected to the Terraform recipe or do we need to do some changes in the Terraform Driver too? cc/ @kachawla

@lakshmimsft could you add high level details in the design doc on the changes to request payload for recipe engine/driver?

recipe/2024-02-terraform-providers.md
lakshmimsft: schema is updated per mtg discussion

recipe/2024-02-terraform-providers.md
lakshmimsft: updated per discussion. only key is needed

recipe/2024-02-terraform-providers.md
lakshmimsft: updated schema per discussion

recipe/2024-02-terraform-providers.md
lakshmimsft: fyi.. envSecretsFrom was removed from the datamodel

recipe/2024-02-terraform-providers.md
lakshmimsft: TerraformConfigProperties currently exists under recipeConfig, we should be able to extend that later into Bicep specific provider configuration as needed.

recipe/2024-02-terraform-providers.md
lakshmimsft: updated datamodel per discussion

recipe/2024-02-terraform-providers.md
lakshmimsft: with updated dm, we should get a list of provider configs with Option2

recipe/2024-02-terraform-providers.md
```diff
@@ -0,0 +1,376 @@
...
+ recipeConfig: {
+    terraform: {
+      ...
+      providers: [
+      {
+        name: 'azurerm',
+        properties: {
+          subscriptionid: 1234,
+          secrets: {                  // Individual Secrets from SecretStore
+            'my_secret_1': {
```
lakshmimsft, line 99: @kachawla , @AaronCrawfis : question fmi: what do we typically write to the tf config file for provider secrets? trying to understand if the 'my_secret_1' key is needed here
is it:
 my_secret_1: value 
 or 
 secret.one: value 
 or
 my_secret_1 {
  secret.one: value 
 }
 or something else?

recipe/2024-02-terraform-providers.md
```diff
@@ -0,0 +1,376 @@
...
+          alias: 'az-paymentservice'
+       }
+     }]
+...       
+```
+Configuration for providers as described in this document will take precedence over provider credentials stored in UCP (currently these include azurerm, aws, kubernetes providers). So, for eg. In the scenario where credentials for Azure are saved with UCP during Radius install and a Terraform recipe created by an operator declares 'azurerm' under the the *required_providers* block; If there exists a provider configuration under the *providers* block under *recipeConfig*, these would take precedence and used to build the Terraform configuration file instead of Azure credentials stored in UCP. 
+
+
+### Example Bicep Input :
+``` diff
```
rynowak, line 123: @lakshmimsft - can you update this to be reflect the option we chose in the meeting? Looks like it's describing both options but we made a decision.

rynowak: Oh, I see you did that below.

rynowak (changes requested):
Waiting on resolution of @kachawla 's comments. Other than that LGTM, examples for option 2 look right.

recipe/2024-02-terraform-providers.md
lakshmimsft: discussed. we keep same dm as in env variables. my_secret_1: value

recipe/2024-02-terraform-providers.md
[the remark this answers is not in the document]
lakshmimsft: updated.

recipe/2024-02-terraform-providers.md
lakshmimsft: updated

recipe/2024-02-terraform-providers.md
lakshmimsft: updated

recipe/2024-02-terraform-providers.md
lakshmimsft: updated

recipe/2024-02-terraform-providers.md
lakshmimsft: updated

.github/config/en-custom.txt
```diff
@@ -612,3 +612,20 @@ parsers
 Shruthi
 Kumar
 sk
+lakshmimsft
+recipeConfig
+recipies
```
ytimocin, line 617:
```suggestion
recipes
```

recipe/2024-02-terraform-providers.md
```diff
@@ -0,0 +1,690 @@
...
+  }
+}
+// The `required_providers` block for this module specifies that it will need provider configuration for aliases west_region and east_region to deploy successfully.
+```
+
+The bicep configuration for Terraform providers described in the environment resource is typically written to `provider` block in the main.tf.json file. These provider configurations are then passed to the recipe module. For provider configurations without aliases, they are be passed to the module automatically by Terraform.
+
+However, when the recipe module specifies one or more aliases in the `required_providers` block (and the module does not contain the provider configuration details), the aliased provider configuration from the `provider` block needs to be passed explicitly to the module.
+
+This is done by updating the module block in the main.tf.json file with provider aliases.
```
rynowak, line 83: When you say `update` here - do you mean that we're updating code the user wrote? or are you saying we'll change the code we current generate?

recipe/2024-02-terraform-providers.md
```diff
@@ -0,0 +1,690 @@
...
+For a Terraform recipe, aliases needed by the module are declared in the required_providers block, where each alias is a unique provider configuration.
+eg.
+
+```
+terraform {
+  required_providers {
+    mycloud = {
+      source  = "mycorp/mycloud"
+      version = "~> 1.0"
+      configuration_aliases = [ mycloud.west_region, mycloud.east_region ]
```
kachawla, line 72: Is this required to use any provider with an alias or does Terraform make implicit assumptions if this is not set?

recipe/2024-02-terraform-providers.md
```diff
@@ -0,0 +1,690 @@
...
+### Key-value pairs
+Attributes or settings, represented as key-value pairs are often used to define configuration parameters in provider configurations.
+We will parse and save these values in the Terraform configuration file, main.tf in a working directory, which the system creates today. 
+
+### Alias
+
+Aliases in Terraform allow users to manage multiple instances of the same provider with different configurations for a single module. 
+Ref links: 
+- [Alias Multiple Provider Configurations](https://developer.hashicorp.com/terraform/language/providers/configuration#alias-multiple-provider-configurations)
+- [Passing Providers Explicitly](https://developer.hashicorp.com/terraform/language/modules/develop/providers#passing-providers-explicitly)
```
rynowak, line 61: Thanks for including the references 👍

recipe/2024-02-terraform-providers.md
```diff
@@ -0,0 +1,690 @@
...
+}
+// The `required_providers` block for this module specifies that it will need provider configuration for aliases west_region and east_region to deploy successfully.
+```
+
+The bicep configuration for Terraform providers described in the environment resource is typically written to `provider` block in the main.tf.json file. These provider configurations are then passed to the recipe module. For provider configurations without aliases, they are be passed to the module automatically by Terraform.
+
+However, when the recipe module specifies one or more aliases in the `required_providers` block (and the module does not contain the provider configuration details), the aliased provider configuration from the `provider` block needs to be passed explicitly to the module.
+
+This is done by updating the module block in the main.tf.json file with provider aliases.
+
```
rynowak, line 84: It would be helpful to me to understand what need to *change* to support aliases.

- I think we generate `required_providers`. It sounds like that changes.
- We generate a `provider` block for each provider/alias. I think it's new that we're aware of aliases in this code.
- We generate a call to the module, it sounds like this needs to change too.

rynowak: I think going over the example from below in more detail (what's new) will answer my question.

recipe/2024-02-terraform-providers.md
```diff
@@ -0,0 +1,690 @@
...
+          "postgresql.pgdb-test": "postgresql.pgdb-test"
+        }
+      }
+    }
+}
+```
+
+#### Use Case 2: Terraform configuration allows passing of provider configurations with aliases different from those declared in `required_providers` block to the recipe module.
+
+In order to support this use case we will have to delve further to design how we enable users to map aliases in provider configuration to those declared in recipe modules in `required_providers` blocks across different recipes.
```
rynowak, line 138: I don't think I understand what this means or whether it's important.

recipe/2024-02-terraform-providers.md
```diff
@@ -0,0 +1,690 @@
...
+
+The bicep configuration for Terraform providers described in the environment resource is typically written to `provider` block in the main.tf.json file. These provider configurations are then passed to the recipe module. For provider configurations without aliases, they are be passed to the module automatically by Terraform.
+
+However, when the recipe module specifies one or more aliases in the `required_providers` block (and the module does not contain the provider configuration details), the aliased provider configuration from the `provider` block needs to be passed explicitly to the module.
+
+This is done by updating the module block in the main.tf.json file with provider aliases.
+
+#### Use Case 1: Matching aliases between provider configuration and `required_provider` block: sample generated main.tf.json file
+
+With current design, we are able to update the module configuration block with provider aliases when a match is found.
```
kachawla, line 87: This choice of design is not something we can revert back from if we decide to let user control the mappings entirely, so let's make sure we think about long term goals before deciding that this is an appropriate intermediate solution.

recipe/2024-02-terraform-providers.md
lakshmimsft: If this is not set Terraform makes implicit assumptions.

recipe/2024-02-terraform-providers.md
lakshmimsft: added language to clarify : provider configurations in Bicep and the recipe module, which includes the `required_providers` blocks, are authored by the user and transcribed into the .main.tf.json file without updates while the module block, is generated by the Radius code.

recipe/2024-02-terraform-providers.md
lakshmimsft: Use of env-vars would be the users' decision i think. We're outlining details on what the system can support. I'm not sure we want to have a blanket recommendation in this document. Each provider may have their recommended way of setting up configuration.

rynowak (approved):

PR #40, opened 2024-02-09
Add provisioning state to output resources property

nithyatsu:


resources/2024-02-output-resources.md
```diff
@@ -0,0 +1,210 @@
...
+
+## Overview
+
+Output resources are the actual platform specific resources that get deployed in order to deploy a Radius resource. For example to deploy a container as part of radius application deployment on kubernetes, we deploy a kubernetes ServiceAccount, Role, RoleBinding and Deployment Output resources.
+
+There are a couple of problems related to how we manage output resources today -
+
+1. Output resources do not have a ProvisioningStatus property. Instead, if any of the output resource deployment fails, we set the Provisioning state of the corresponding radius resource accordingly and do not store the output resources. With this approach, we lose the information on which output resources failed deployment since we aggregate failures into the owning Radius resource's provisioningStatus property. 
+
+2. We do not record any of the output resources as part of the radius resource whose deployment failed, although we deployed them with an errored status. This leads to orphaned resources which should be manually cleaned up.
```
rynowak, line 14: This can be fixed without an API change right?

resources/2024-02-output-resources.md
```diff
@@ -0,0 +1,210 @@
...
+
+This design proposes addition of ProvisioningStatus to outputResources so that, 
+once we add changes to track these resources in case of both successful and failed deployment to avoid orphaning resources, there will be clearer indication of what lead to failed deployment to the user.
```
rynowak, line 17: I want to push back on one aspect of this. The `operationStatus` and `operationResult` are the things that describe the result of an operation. The resource itself describes the state of the resource. They have different purposes.

I'm saying this because we need to keep the purpose in mind when we're designing. For the example the error messages or other debugging information are not part of the resource, they are part of the `operationStatus`, and it should stay that way.

rynowak: I am ok with adding the provisioningState for output resources, but we should be careful how we're applying this reasoning.

resources/2024-02-output-resources.md
```diff
@@ -0,0 +1,210 @@
...
+
+## Objectives
+
+> **Issue Reference:**
+
+<https://github.com/radius-project/radius/issues/7052>
+
+### Goals
+
+The goal is to record the output resources whether or not deployment succeeds, so that radius can still manage the cleanup of resources it failed to deploy successfully.
```
rynowak, line 38: This is independent of the API change right?

resources/2024-02-output-resources.md
```diff
@@ -0,0 +1,210 @@
...
+}
+
+@doc("Provisioning Status of Output Resource")
+enum ProvisioningStatus {
+  @doc("The resource is successfully provisioned")
+  Succeeded,
+
+  @doc("The resource provisioning failed")
+  Failed,
+}
```
rynowak, line 140: We shouldn't need a separate definition of this, it should be the same as `.properties.provisioningState`.

resources/2024-02-output-resources.md
```diff
@@ -0,0 +1,210 @@
...
+model OutputResource {
+  @doc("The logical identifier scoped to the owning Radius resource. This is only needed or used when a resource has a dependency relationship. LocalIDs do not have any particular format or meaning beyond being compared to determine dependency relationships.")
+  localId?: string;
+
+  @doc("The UCP resource ID of the underlying resource.")
+  id?: string;
+
+  @doc("Determines whether Radius manages the lifecycle of the underlying resource.")
+  radiusManaged?: boolean;
+
```
rynowak, line 128: What would provisioningState be for a non-radius-managed resource?

resources/2024-02-output-resources.md
```diff
@@ -0,0 +1,210 @@
...
+  @doc("Determines whether the resource was provisioned successfully")
+  provisioningStatus: ProvisioningStatus
```
rynowak, line 130: This should be the same as `properties.provisioningState`. The goal is to reuse the concept, so it should be spelled the same.

resources/2024-02-output-resources.md
```diff
@@ -0,0 +1,210 @@
...
+}
+
+@doc("Provisioning Status of Output Resource")
```
vinayada1, line 133: Are these the only possible values? What will happen in the case where 2 output resources are rendered and the deployment of the 1st output resource fails? What will be the provisioningStatus of resource 2?

resources/2024-02-output-resources.md
```diff
@@ -0,0 +1,210 @@
...
+          {
+            "id": "/planes/kubernetes/local/namespaces/allinoneenv-allinoneapp/providers/core/Service/allinonecontainer",
+            "localId": "Service"
+            "provisioningState": "Succeeded"
+          }
+        ]
+      }
+      :
+]
+```
```
vinayada1, line 185: There would be some changes needed in the delete logic as well, right?

resources/2024-02-output-resources.md
```diff
@@ -0,0 +1,210 @@
...
+1. Updating conversion unit tests to make sure the new attribute is handled correctly
+2. Updating some API unit tests to see if the information is being populated on the resource level.
+
+## Monitoring
+
+N/A
+
+## Development plan
+
+There is an in-progress PR which brings in the proposed changes as part of fix for rad app delete.
```
vinayada1, line 203: Would be good to also specify the steps and link to your PR.

vinayada1, line 203: There would be some changes needed in the delete logic as well, right?

resources/2024-02-output-resources.md
nithyatsu: yes, no API changes needed.

resources/2024-02-output-resources.md
nithyatsu: yes.. I was just trying to capture what lead to this.

resources/2024-02-output-resources.md
nithyatsu: Based on @rynowak 's feedback, I will just use the existing provisioning state enum with all its values. 
If only 1st outputResource failed, only its provisioning status will be failed. And the parent radius resource will have a failed provisioningStatus too. resource 2 will have its status based on its own deployment being successful.

resources/2024-02-output-resources.md
nithyatsu: Yes. https://github.com/radius-project/radius/pull/7148 
The PR needs to be updated based on the feedback I receive here, but it tests the idea out and worked in case of container deployment failures  followed by a rad app delete.

nithyatsu:
closing due to low priority on the issue.

PR #42, opened 2024-03-07
Add plane api proposal

rynowak:


ucp/2024-03-planes-apis.md
```diff
@@ -0,0 +1,335 @@
...
+```
+
+This design leverages discriminated unions. The `type` and `location` fields are missing from the response due to bugs.
+
+#### Proposed Design
+
+There are 4 parts to the API new design, because we're separating out one type into 4.
+
+- A "generic" plane resource that's used with `/planes` (list all planes of all types)
+- AWS plane: `System.Planes/aws`
```
kachawla, line 141: How do you envision specifying cloud partition in this? I believe last part of current id allowed providing cloud partition `/planes/aws/aws-cn` for example

ucp/2024-03-planes-apis.md
```diff
@@ -0,0 +1,335 @@
...
+## Open issues
+
+**Naming?**
+
+I'd like some help with naming. I'm not sure that `System.Planes/aws` is right. It feels like we're defining more namespaces than necessary. 
+
+Some other ideas I had:
+
+- `System.Resources/awsPlane`
+- `System.Aws/planes` (we already have `System.Aws/credentials`)
```
rynowak, line 335: From discussion in the design meeting we chose `System.Aws/planes` as the style of the API. This is recommended for consistency with `System.Aws/credentials`. This pattern also scales up as we add new things.

I'll update the design doc with this decision.

kachawla (approved):

ucp/2024-03-planes-apis.md
rynowak: This is a resource type, not a resource ID. The resource type for a scope doesn't appear in the URL. The resource IDs/URLs will be unchanged from this proposal, which is why it wasn't discussed 👍

PR #44, opened 2024-04-19
Cli improvements

nithyatsu:
Redesign cli experience to provide better default behaviors for common patterns.

### DIFF
--- a/cli/cli-improvements.md
+++ b/cli/cli-improvements.md
@@ -265,9 +265,20 @@ The aforementioned commands require a -g flag to specify group.
 
 These commands should support -g flag mandatorily if a -e is used to specify the environment name. If -e is not specified, default environmentID in config file can be used.
 
-In summary, most CLI commands should support an additional mandatory -g flag, making user experience more verbose. 
+## Test plan
 
-We could support setting a default group which would be valid through a session for "read" actions.
+As applicable we would add test cases to support the added -g flag. 
+We also would modify existing tests to remove a default resource group assumption.
+
+
+## Development plan
+
+
+## Question
+
+This approach requires most CLI commands to support an additional mandatory -g flag, making user experience more verbose. 
+
+Would supporting a default group which would be valid through a session for "read" actions to make this experience better?
 
 
 
--- a/cli/cli-improvements.md
+++ b/cli/cli-improvements.md
@@ -25,33 +25,47 @@ There are a few points that can be improved in the way these concepts work toget
 
 | Term | Definition |
 |---|---|
-| workspace | |
-| environment | |
-| scope | |
+| Workspace |Workspaces allow managing multiple Radius environments using a local, client-side, configuration file. |
+| Environment | Environments are server-side resources that exist within the Radius control-plane. Applications deployed to an environment will inherit the container runtime, configuration, Recipes, and other settings from the environment.|
+| Scope | Radius resource groups are used to organize Radius resources, such as applications, environments and portable resources.|
 
 
 ## Objectives
 
 ### Goals
 
-Enhance the user experience with `rad deploy` by making its interaction with various Radius concepts more intuitive.
+* Improve the deploy experience (rad deploy, rad run) with better default behaviors for common patterns (deploy app to multiple environments, deploy multiple apps to save environment).
+* Improve the ability of various CLI commands and the Dashboard (rad resource list) to work with resource groups, and to work across all resource groups.
+* Make resource groups more visible in our documentation, and provide recommendations towards common patterns.
 
 ### Non-Goals
 
-NA
+* (out of scope) Server-side changes (other than Dashboard).
 
 ### User scenarios
 
-* As a developer, I would prefer not to manually enter or set the group to which my application will be deployed each time. It would be beneficial if Radius could automatically select the group name, ensuring all related resources are grouped together and avoiding conflicts with resources from other applications. I should have the option to override this behavior when necessary.
+* As someone new to Radius, I am using the tutorials to make myself familiar with its concepts. I do not want to get into details, yet. 
 
-* As a developer, I would like to avoid the responsibility of selecting unique names for components within my application to prevent conflicts with other deployed applications.
+* I would like to deploy multiple applications to the same environment.
 
+* I would like to deploy a single application to multiple environments.
+  
 ## Design
 
-We aim to redesign the rad CLI commands, adopting a convention-over-configuration approach to determine the Radius resource group to be operated on. Consequently, 
+Today, we establish a default resource group during `rad init` process in `config.yaml`.  All of the `rad` cli commands operate on this resource group, unless a different resource group is specified using option `-g`.
+
+With respect to our user scenarios, this means, 
+
+* A new user has to understand managing Radius resource groups to be able to deploy a simple tutorial application.
+
+* The user has to pick unique resource names across all applications since they would all be deployed to the same resource group (default behavior).
+
+* The user has to create multiple resource groups and manage deploying the same application to different resource groups in order to have multiple copies of the application.
+
+In order to improve the default experience of `rad` commands, we want to adopt a convention-over-configuration approach in determining the Radius resource group to be operated on. This means,
 
 * we will not establish a default resource group during the 'rad init' process in config.yaml.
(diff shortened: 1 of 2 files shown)


cli/cli-improvements.md
```diff
@@ -0,0 +1,315 @@
...
+
+
+*	Today, if we want two copies of same app, we want to create two resource group, have an environment in each of those resource group, have the user explicitly use rad group switch and rad env switch or specify the group and env using flags, to make a desirable deployment. Would it be a better idea to, instead of expecting users specify instructions explicitly, print a default behavior which is intuitive and deploy the app if user confirms?
+
+
+*	Since only ResourceGroup  is part of resource ID, if we deploy multiple application to same resource group (ex: we do this in our functional tests) we have to ensure the resource names across applications are all unique. This limitation has no reason to be present in real world and is a hinderance since the developer has to make sure to choose names that are not already taken across applications. Again, we could create a resource group based on user scenario per instance of application using an agreed upon convention to solve this problem. 
+
```
rynowak, line 23: I think this is a pretty good intro. I'm expecting @willtsai to provide this as part of the feature spec, and you can reference it here. 

These are good examples.

cli/cli-improvements.md
```diff
@@ -0,0 +1,315 @@
...
+| workspace | |
+| environment | |
+| scope | |
+
+
+## Objectives
+
+### Goals
+
+Enhance the user experience with `rad deploy` by making its interaction with various Radius concepts more intuitive.
```
rynowak, line 37: I think the goal is bigger than just `rad deploy`. 

- Improve the deploy experience (`rad deploy`, `rad run`) with better default behaviors for common patterns (deploy app to multiple environments, deploy multiple apps to save environment).
- Improve the ability of various CLI comments (`rad resource list`) to work with resource groups, and to work across all resource groups. 
- Make resource groups more visible in our documentation, and provide recommendations towards common patterns.

cli/cli-improvements.md
```diff
@@ -0,0 +1,315 @@
...
+
+### Non-Goals
+
+NA
```
rynowak, line 41: Suggestion:

- *(out of scope)* Server-side changes (other than Dashboard).

cli/cli-improvements.md
```diff
@@ -0,0 +1,315 @@
...
+
+### User scenarios
+
+* As a developer, I would prefer not to manually enter or set the group to which my application will be deployed each time. It would be beneficial if Radius could automatically select the group name, ensuring all related resources are grouped together and avoiding conflicts with resources from other applications. I should have the option to override this behavior when necessary.
+
+* As a developer, I would like to avoid the responsibility of selecting unique names for components within my application to prevent conflicts with other deployed applications.
+
```
rynowak, line 48: I'd suggest three scenarios:

- Learner doing a tutorial. (Focus is on simplicity, this is basocally what you wrote down) 
- Deploying multiple applications to the same environment.
- Deploying a single application to multiple environments.

cli/cli-improvements.md
```diff
@@ -0,0 +1,315 @@
...
+## Design
+
```
rynowak, line 50: This is a good place to explain the current behavior. Suggestion

- What does `rad` do today with respect to resource groups? (and what's the problem with it)
- What would happen if we took out the idea of "default scope" from `config.yaml`
- Once we do that, many commands become very verbose, how would we simplify things.

cli/cli-improvements.md
```diff
@@ -0,0 +1,315 @@
...
+            connection:
+                context: radius-agent-aks
+                kind: kubernetes
+            environment: /planes/radius/local/resourceGroups/default/providers/Applications.Core/environments/default
+```
+
+All CLI commands will be impacted due to this change.
+
+
+### Example
```
rynowak, line 88: This is a really good place to talk about your three scenarios again. eg: give an example of each. You can make the point that the two non-tutorial scenarios ignore/fight-with one of our built in behaviors (default scope)

cli/cli-improvements.md
```diff
@@ -0,0 +1,315 @@
...
+   Use rad deploy <app-name> -g <group-name> -e <env-name> 
+   to change the behavior. Proceed? [Y/n]
+   :
+
+4. rad deploy cool-app
+   which env would you like to deploy to? [prefilled to production but set to /planes/radius/local/resourcegroups/all-environments/providers/applications.core/environments/qa] 
+   `cool-app` being deployed using environment qa in resource group cool-app-qa.
+   Use rad deploy <app-name> -g <group-name> -e <env-name> 
+   to change the behavior. Proceed? [Y/n] 
+   :
```
rynowak, line 141: Reminder: we want to be really clear and specific about which commands are interactive and not. Ideally we have few interactive commands.

cli/cli-improvements.md
```diff
@@ -0,0 +1,315 @@
...
+### rad group 
+
+The ```rad group``` family of commands have no changes. ```rad group switch``` alone will become obsolete since it deals with setting the current scope as a configuation in config.yaml.
+
+### rad env 
+
+The ```rad env``` family of commands currently operate in the scope that is stored in config.yaml, unless a scope is explicitly specified using -g flag.   
+
+Since the idea of "default" scope is going away, all rad env commands will need a mandatory -g flag to specify the group.
+
```
rynowak, line 233: How would we list environments across all groups? (hint, look at kubectl + listing across all namespaces).

rynowak: When `rad env list` fails to find an envioronment, we might want to print a suggestion to use `-A`.

cli/cli-improvements.md
```diff
@@ -0,0 +1,315 @@
...
+#### rad env list 
+
+This could operate at two levels
+1. list all environments in a specified resource group
+```rad env list -g cool-group```
+2. list all environments across resource groups
+```rad env list```
```
rynowak, line 240: We will have a choice to make here between consistency and simplicity. If each environment is in its own group (by recommendation), then it's tricky how best to design this command.

cli/cli-improvements.md
```diff
@@ -0,0 +1,330 @@
...
+* **Status**: Draft
+* **Author**: Nithya Subramanian (@nithyatsu)
+
+## Overview
+
+When a Radius user wants to deploy an application, a few concepts come into picture
+
+*	Resource group – Radius resource group mimic ARM resource group and are tied to RBAC. 
+*	Environment – Radius environment is what the application uses. The environment can have a set of recipes and is “in” a resource group itself, since it requires RBAC to make sure not everyone can edit it and for example, point a Production environment to use Sandbox recipes.
+*	Workspace – Radius workspace has a "per user" information about the kubernetes context/ cluster where the Radius services are hosted as well as the “default” environment and resource group  of the user.
```
willtsai, line 12: if it's "per user" then does it mean the user can customize these things? what and how can they customize?

cli/cli-improvements.md
```diff
@@ -0,0 +1,330 @@
...
+```
+workspaces:
+    default: default
+    items:
+        default:
+            connection:
+                context: radius-agent-aks
+                kind: kubernetes
+            environment: /planes/radius/local/resourceGroups/default/providers/Applications.Core/environments/default
+```
```
willtsai, line 100: so if `scope` is completely dropped from the config, does it mean that instead of deploying to the `default` (aka "scoped") resource group, new resource groups based on environment and app names per the convention will be created each time if no `-g` is specified?

cli/cli-improvements.md
```diff
@@ -0,0 +1,330 @@
...
+
+All CLI commands will be impacted due to this change. Most commands have to  support taking a mandatory resource group argument.
```
willtsai, line 102: "Most commands have to  support taking a mandatory resource group argument." --> can you elaborate on why they will now require a mandatory `-g` argument? Why didn't it require one before?

cli/cli-improvements.md
```diff
@@ -0,0 +1,330 @@
...
+
+
+### Examples
```
willtsai, line 105: Have we considered giving the user a visual indicator in their terminal regarding which Radius environment and resource group they are currently in? kind of like how Python Flask has an indicator when you activate a virtual environment.

rynowak:
Closing this as stale, @nithyatsu please reopen in the future when you're ready to work on this 👍

PR #45, opened 2024-04-25
Update design doc template

kachawla:
Updating the template to include more details/subsections with higher level of granularity.

### DIFF
--- a/template/YYYY-MM-design-template.md
+++ b/template/YYYY-MM-design-template.md
@@ -65,9 +65,38 @@ link to that issue instead.
 
 #### User story 2
 
+## User Experience (if applicable)
+<!--
+If the change impacts the user experience, provide expected interaction 
+flow we aim to achieve through this proposal.
+
+When users interact with Radius through the CLI, include sample 
+input commands and their corresponding output. Include a bicep/helm code 
+sample, if this proposal involves updates to that experience.
+-->
+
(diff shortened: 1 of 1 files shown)


ytimocin (approved):

template/YYYY-MM-design-template.md
```diff
@@ -65,9 +65,38 @@ link to that issue instead.
 
 #### User story 2
 
+## User Experience (if applicable)
```
sk593, line 68: What is this difference between this and the user scenarios section?

kachawla: Good question!

The user scenario section is expected to provide background on *when and why* a user would need this functionality. For example: https://github.com/radius-project/design-notes/blob/d6b3ea8c44905b5e6eac7b328ab33869e488b41e/recipe/2023-08-garbage-collection.md?plain=1#L37.

The user experience section aims to provide insights into *how* a user would interact with this functionality, outlining the proposed end to end user flow.

Please feel free to suggest any updates if this can be made clearer in the template to avoid any confusion. :)

I expect both of these sections to be covered as part of the product specification. The design doc should primarily copy them over from the product spec document going forward. cc @willtsai

kachawla:
Going to merg this. Feel free to leave comments and I'll address them in a follow up PR.

PR #48, opened 2024-05-06
Add workflow for stale pull requests

rynowak:
Proposing some automation to help us keep design-notes clean. This will give a gentle nudge when a design-notes PR goes stale after 14 days, and then close the pull-request after an additional 7 days with no activity.

### DIFF
--- a/.github/workflows/stale-prs.yaml
+++ b/.github/workflows/stale-prs.yaml
@@ -1,0 +1,42 @@
+# ------------------------------------------------------------
+# Copyright 2023 The Radius Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#    
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# ------------------------------------------------------------
+
+# See: https://github.com/actions/stale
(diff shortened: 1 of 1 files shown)


kachawla (approved):
Thanks for adding this!

14 days could be too aggressive, but could also be a good motivation to prioritize closing design docs before moving forward with the implementation.

rynowak:
> Thanks for adding this!
> 
> 14 days could be too aggressive, but could also be a good motivation to prioritize closing design docs before moving forward with the implementation.

I debated this too, but it's 14 days with **no activity** before the warning shows up, and then another 7 days. Worst case you just reopen the PR.

If folks have strong feedback we can adjust 😁

.github/workflows/stale-prs.yaml
```diff
@@ -0,0 +1,42 @@
...
+    runs-on: ubuntu-latest
+    steps:
+     - uses: actions/stale@v9
+       with:
+        stale-pr-message: |
+          This pull request is stale because it has been open 14 days with no activity. Remove stale label or comment or this will be closed in 7 days.
+        close-pr-message: |
+          This pull request has been closed due to inactivity. Feel free to reopen if you are still working on it.
+        days-before-stale: 14
+        days-before-close: 7
```
ytimocin, line 40: Does **stale** mean PRs with activity but not merged yet?
Any PR that doesn't have activity for 7 days will be deleted even if it is not stale?

ytimocin (approved):

.github/workflows/stale-prs.yaml
rynowak: I think this section of the docs explains this pretty well: https://github.com/actions/stale?tab=readme-ov-file#close-stale-issues-and-prs

rynowak: With our configuration this will only close pull requests after 21 days of no activity.

PR #49, opened 2024-05-10
Add Radius on Dapr proposal

rynowak:


architecture/2024-05-radius-on-dapr.md
```diff
@@ -0,0 +1,324 @@
+# Radius on Dapr
+
+* **Status**: Pending/Approved
+* **Author**: Ryan Nowak (`@rynowak`)
+
+## Overview
+
+The design doc proposes replacing some existing plumbing in Radius with [Dapr](https://dapr.io) and taking an install-time dependency on Dapr. Radius needs a database and a message queue to do its work. Dapr provides implementations of its [building block APIs](https://docs.dapr.io/concepts/building-blocks-concept/) that work across self-hostable open-source technology and managed cloud services. Compared to developing our own data layer, adopting Dapr will accellerate development and production-readiness of Radius because Dapr is already mature and production-hardened. Dapr provides SDKs for most languages, including Go and C# so it can be used in the Deployment Engine.
```
ytimocin, line 8: Are we replacing **etcd** and **in-memory queue**?

architecture/2024-05-radius-on-dapr.md
```diff
@@ -0,0 +1,324 @@
...
+
+Dapr also brings with it a powerful [workflow](https://docs.dapr.io/developing-applications/building-blocks/workflow/) programming model that can be used to write reliable and scalable distributed systems workflows. This is ideal for building more complex background processing in Radius, and also ideal for us to offer extensibility to users.
+
+Dapr is well-supported both inside and outside of Kubernetes. Adopting a technology like Dapr will help Radius work well regardless of the hosting scenario.
```
ytimocin, line 12: If we decide to do **serverless**, would **Dapr** also be able to help there too?

architecture/2024-05-radius-on-dapr.md
ytimocin: We can also use the secret store of **Dapr**, I guess...

architecture/2024-05-radius-on-dapr.md
```diff
@@ -0,0 +1,324 @@
...
+
+
+## Objectives
+
+> **Issue Reference:** <!-- (If appropriate) Reference an existing issue that describes the feature or bug. -->
+
+### Goals
+
+- Adopt Dapr for our data storage and messaging needs. 
+  - Keep `rad init` and `rad install` simple by including Dapr and a default choice of components as part of our Helm chart.
```
ytimocin, line 38: What would happen if my cluster already has **Dapr** installed?

architecture/2024-05-radius-on-dapr.md
```diff
@@ -0,0 +1,324 @@
...
+  - Keep the Helm chart flexible by documenting how to BYO components.
+  - Provide production-guidance for Radius based on Dapr's existing guidance.
+- Remove our existing messaging and storage implementations in favor of Dapr.
+
+### Non goals
+
+- (out of scope) User-facing behavior changes other than installation/dependencies.
+- (out of scope/follow-up) Adoption of Dapr workflows.
+  - While this is highly valuable for us, this
```
ytimocin, line 47: Was this sentence meant to be completed?

architecture/2024-05-radius-on-dapr.md
```diff
@@ -0,0 +1,324 @@
...
+✅ Installing Radius 0.32.0 on kind-kind
+✅ Installing Dapr 1.13.0 on kind-kind
+✅ Create new environment default
+   - Kubernetes namespace: default
+   - Recipe pack: local-dev
+✅ Update local configuration
+```
+
+
+**Sample Recipe Contract:**
```
ytimocin, line 95: Could we keep recipes in a state store of Dapr so that we don't need to download it from ghcr? I am also not sure what advantages that would bring 😀 Just thinking out loud.

architecture/2024-05-radius-on-dapr.md
```diff
@@ -0,0 +1,324 @@
...
+
+*n/a*
+
+### Development Experience
```
kachawla, line 99: 👍

architecture/2024-05-radius-on-dapr.md
```diff
@@ -0,0 +1,324 @@
...
+- Updating the async work worker.
+- Deleting unused code.
+
+#### Setup
+
+For setup (`rad init`/`rad install`) we previously installed Dapr (a long time ago) along with Contour. We should follow the same design approach we currently do for Contour:
+
+- `rad install` will install Dapr.
+- `rad uninstall` will not uninstall Dapr.
+- `rad install --reinstall` will not upgrade Dapr.
```
ytimocin, line 143: Why not? If upgraded, do we lose data?

architecture/2024-05-radius-on-dapr.md
```diff
@@ -0,0 +1,324 @@
...
+
+Whichever options we choose, we'll add them to our Helm chart with a new option `global.dapr.defaultComponents=true` as an opt-in. Our `rad init` and `rad install` commands will set this option. 
+
+I feel better about opt-in vs out-out for the Helm Chart. It's very typical that Helm Charts require additional settings or configuration when dependencies are involved. 
+
+These components should be configured to balance resource usage and complexity versus production readiness. We want to do something reasonable, but we also don't expect users to use our default components for production.
+
+#### State store
+
+Our task is to implement the [`StorageClient` API](https://github.com/radius-project/radius/blob/main/pkg/ucp/store/client.go#L25) on top of the Dapr state store API. Since we'll be keeping the existing store API this change should have no impact on the rest of the codebase.
```
ytimocin, line 182: We can easily switch between CosmosDB and etcd. We can go with something like that.

architecture/2024-05-radius-on-dapr.md
```diff
@@ -0,0 +1,324 @@
...
+Wilford is a developer that's making some contributions to Radius for the first time. To get started, he reads the developer document at in [the repo](https://github.com/radius-project/radius/blob/main/docs/contributing/contributing-code/contributing-code-control-plane/running-controlplane-locally.md). He sees that installing Dapr is part of the prerequisites, so he follows the [getting started instructions](https://docs.dapr.io/getting-started/).
+
+After that he follows the rest of the instructions to set up a Kubernetes cluster, and configure his `rad` CLI to talk to the local instance of Radius. When he's done with the setup instructions, he can launch Radius in the debugger. When he does that Radius manages the lifetime of each Dapr process so that Wilford does not to do that manually or learn any extra commands.
+
+## Design
+
+### High Level Design
+
+For our immediate needs we'll be replacing the Radius database and message queue with Dapr's functionality. This doesn't require an architectural change or a large update to the Radius codebase. We'll keep our existing abstractions like the async worker and state store in place. 
+
```
nithyatsu, line 112: would we also eventually use dapr's observability to monitor radius?

architecture/2024-05-radius-on-dapr.md
```diff
@@ -0,0 +1,324 @@
...
+-- 
+
+I'm also working on an additional proposal to replace the async worker with Dapr workflows. If we think it's better, we could skip the intermediate step of upgrading the current architecture. The path outlined above is the most incremental approach. 
+
+#### Development Experience
+
+Since Dapr is a new required component, we will need to make some investments or our development experience will be poor. 
+
+I'm proposing that we require Dapr to by installed by users to develop Radius. We'll need the `daprd` binary to be present on the user's system.
```
ytimocin, line 217: Should a user be given an option to not install Dapr and do it the way we have now? Do we need to be heavily dependent on Dapr?

architecture/2024-05-radius-on-dapr.md
```diff
@@ -0,0 +1,324 @@
...
+- Deleting unused code.
+
+#### Setup
+
+For setup (`rad init`/`rad install`) we previously installed Dapr (a long time ago) along with Contour. We should follow the same design approach we currently do for Contour:
+
+- `rad install` will install Dapr.
+- `rad uninstall` will not uninstall Dapr.
+- `rad install --reinstall` will not upgrade Dapr.
+
```
lakshmimsft, line 144: if new version of radius brings in a new version of Dapr, then will there be an upgrade?

architecture/2024-05-radius-on-dapr.md
nicolejms, line 8: We spent time last year removing Dapr from installer due to download speed. Is this still a concern?

ytimocin, line 8: Will we give the option to opt-in/opt-out of Dapr during Radius installation?

architecture/2024-05-radius-on-dapr.md
```diff
@@ -0,0 +1,324 @@
...
+- Including default component configurations in our Helm Chart.
+- Updating the database implementation.
+- Updating the async work worker.
+- Deleting unused code.
+
+#### Setup
+
+For setup (`rad init`/`rad install`) we previously installed Dapr (a long time ago) along with Contour. We should follow the same design approach we currently do for Contour:
+
+- `rad install` will install Dapr.
```
nithyatsu, line 141: would it be useful to make dapr optional? What if some users did not want dapr but want rdaius?

architecture/2024-05-radius-on-dapr.md
ytimocin: We can talk about this question above. Would there be a clash? Or are two **Dapr**s (one inside Radius and one running on the cluster) two different things?

architecture/2024-05-radius-on-dapr.md
```diff
@@ -0,0 +1,324 @@
...
+
+#### Wendy tries Radius out
+
+Wendy wants to learn more about Radius, so she fires up a Kubernetes cluster and tries some of our tutorials. She runs `rad init` to install Radius, which includes Dapr and some components that Radius can use for its storage. This is easy for her because she didn't need to learn about Dapr to try Radius out.
+
+*This scenario is here to point out that the simplicity of the current user experience is maintained.*
+
+#### Wendy and Jake put Radius in production
+
+Wendy's been successful experimenting with Radius, and she wants to take it to production. She works with Jake to set up some cloud datastores for Radius to use, to ensure that data is persisted. Jake creates a DynamoDB table and the necessary SNS/SQS resources, following the Dapr documentation. Jake reads the Radius documentation, installs Dapr, and configures the `Component` objects in Kubernetes that are required. Lastly, Jake runs `helm install radius radius -n radius-system` to install Radius without its default component configuration.
```
kachawla, line 59: > Jake creates a DynamoDB table and the necessary SNS/SQS resources, following the Dapr documentation

Shouldn't this be independent of Dapr, why do they need to follow Dapr docs?

> Jake reads the Radius documentation, installs Dapr

I understood dapr installation would be done by default

ytimocin, line 59: So, installing Radius with Helm will not install **Dapr**? Can we talk about this scenario?

architecture/2024-05-radius-on-dapr.md
rynowak: > Are we replacing **etcd** and **in-memory queue**?

Probably yes to this. I think it's better if we can rely on Dapr and not have additional dependencies in our code.

> We can also use the secret store of Dapr, I guess...

The secret store of Dapr is **read only**. Most of our secret operations are read/write so it's not a great fit unfortunately.

rynowak: Yes. The installation time is a drawback. We're probably talking about adding a minute to install time for local kubernetes, and much less for cloud-hosted.

architecture/2024-05-radius-on-dapr.md
rynowak: Yes absolutely.

architecture/2024-05-radius-on-dapr.md
```diff
@@ -0,0 +1,324 @@
...
+
+**Sample Recipe Contract:**
+
+*n/a*
+
+### Development Experience
+
+*Adding Dapr as a dependency has the potential to impact our development experience. We can follow some existing patterns that Dapr users like to simplify this.*
+
+Wilford is a developer that's making some contributions to Radius for the first time. To get started, he reads the developer document at in [the repo](https://github.com/radius-project/radius/blob/main/docs/contributing/contributing-code/contributing-code-control-plane/running-controlplane-locally.md). He sees that installing Dapr is part of the prerequisites, so he follows the [getting started instructions](https://docs.dapr.io/getting-started/).
```
ytimocin, line 103: Why is it a prerequisite? Is Radius going to depend on local **Dapr** installation?

architecture/2024-05-radius-on-dapr.md
rynowak: I think we'd skip it. We already have logic for this for Contour. We should be consistent with that behavior unless we have a reason not to.

architecture/2024-05-radius-on-dapr.md
rynowak: Yes :). I'll take a todo

architecture/2024-05-radius-on-dapr.md
```diff
@@ -0,0 +1,324 @@
...
+
+<p style="text-align: center;font-style: italic;">
+Fig: Diagram depicting Radius hosting in Kubernetes. <a href="https://excalidraw.com/#json=plLzyHWUCb76t3NnOoiQf,kKABl_YcZOObWznIGKj1jQ">link</a>
+</p>
+
+### Detailed Design
+
+We can break the detailed design into four main categories of work:
+
+- Including Dapr in `rad init` and `rad install`.
```
ytimocin, line 131: Would `rad uninstall` uninstall all the data in **Dapr** components?

architecture/2024-05-radius-on-dapr.md
rynowak: That's another good idea. Or could use something like MinIO as a cache. I'm treating this as out of scope for now since this plan is focused on the minimal requirements for adoption.

architecture/2024-05-radius-on-dapr.md
rynowak: I think it's an augment not a replacement.

architecture/2024-05-radius-on-dapr.md
ytimocin: What if there is a new version by the time we are running `--reinstall`? Should we ask the user if they want to upgrade?

rynowak: Simplicitly. `rad init` & `rad install` are for non-production usecases primarily. We expect users in production to manage Dapr.

The upgrade will not result in data loss. It's not Dapr holding our data, it's some other technology like postgres.

architecture/2024-05-radius-on-dapr.md
rynowak: 👍

architecture/2024-05-radius-on-dapr.md
rynowak: We either take a dependency or not. We can't have the benefits without being required, because we'll be building an abstraction layer for an abstraction layer.

architecture/2024-05-radius-on-dapr.md
```diff
@@ -0,0 +1,324 @@
...
+
+#### State store
+
+Our task is to implement the [`StorageClient` API](https://github.com/radius-project/radius/blob/main/pkg/ucp/store/client.go#L25) on top of the Dapr state store API. Since we'll be keeping the existing store API this change should have no impact on the rest of the codebase.
+
+The storage API in Radius is already oriented around key-value stores, so it's mostly a 1:1 replacement. 
+
+The challenging portion is that Dapr lacks a widely-supported query API. The state store building block includes an optional query feature, but it's deprecated and not widely supported. We shouldn't use it.
+
+The alternative is that we build our own "index" feature when inserting or deleting records. This would be used when querying resources at a given scope.
```
willdavsmith, line 188: is this a well-known pattern? or a hacky kinda thing

kachawla, line 188: Not sure if I follow this - dapr doesn't support the underlying database features?

architecture/2024-05-radius-on-dapr/hosting-diagram.png
kachawla, line 1: What is being replaced in DE through dapr as a part of this design?

kachawla: Nvm, just noticed this called out in the development plan - `Replace DE in-memory data layer with Dapr state store`

architecture/2024-05-radius-on-dapr.md
```diff
@@ -0,0 +1,324 @@
...
+
+Dapr is also easy to use in integration tests because it provides in-memory components. We can test more of Radius in integration tests by using this support.
+
+## Security
+
+Adopting Dapr as described in this document does not change our existing security posture. However Dapr comes with many security features we can leverage in the future, and will make improving our security posture easier.
+
+## Compatibility (optional)
+
+We don't currently support upgrades from one release to another. Adopting Dapr and mature database technology like Postgres will help us support upgrades in the future.
```
kachawla, line 269: > Adopting Dapr and mature database technology like Postgres will help us support upgrades in the future

Curious to hear more on how this will help with upgrades

architecture/2024-05-radius-on-dapr.md
rynowak: If Dapr is optional that would mean that our use of Dapr is also optional. We can't have the benefits without taking a dependency. 

I'm not worried about this, in practice lots of similar projects have dependencies.

architecture/2024-05-radius-on-dapr.md
rynowak: > Shouldn't this be independent of Dapr, why do they need to follow Dapr docs?

Some of the Dapr components require/recommend specific settings in cloud resources. The point I want to make is that we're not responsible for this in Radius, it's part of an existing project. Any ideas to make this more clear?

> I understood dapr installation would be done by default

There are two scenarios:

- `rad init`/`rad install kubernetes` will include Dapr & Contour (not recommended for production)
- `helm install radius ...` will NOT include Dapr & Contour (recommended for production)

This is consistent with our existing behavior for a dependency (Contour)

architecture/2024-05-radius-on-dapr.md
rynowak: Answered elsewhere in the doc.

architecture/2024-05-radius-on-dapr.md
rynowak: Yes, what do you want to talk about?

architecture/2024-05-radius-on-dapr.md
rynowak: Yes. If Dapr is required, then it's required. Do you see this as a problem?

architecture/2024-05-radius-on-dapr.md
rynowak: This is already answered elsewhere in the doc.

architecture/2024-05-radius-on-dapr.md
rynowak: This is a pretty well-known pattern. It's basically what RDBMS'es do when you define an index.

ytimocin: With RDBMS, it could be easier to keep track of all resources of an environment and application. I think we were talking about building something like this many months ago.

rynowak: There are basically three options:

- Implement an "index" on the client-side and use the Dapr state store
- Take a dependency on the "query" functionality in Dapr state store (which is deprecated)
- Don't use Dapr to store our data - requires us to define another solution

architecture/2024-05-radius-on-dapr.md
rynowak: I'd like to dig into this more in the detailed design doc for state storage.

architecture/2024-05-radius-on-dapr.md
ytimocin: If I don't know about **Dapr**, then to use **Radius** I will need to go and install **Dapr**. It feels like it is one additional step for the user. Not a big problem if we expect/assume users of Radius to also know about Dapr.

Another thought is that if I am installing a software and, to use it, it asks me to install another one. Basically, I need to install 2 software. 🤔

We can discuss further.

architecture/2024-05-radius-on-dapr.md
ytimocin: After all the discussion it makes sense to me. We don't need to talk on this more.

architecture/2024-05-radius-on-dapr.md
rynowak: Would it be ok to answer this as part of the detailed design doc for storage?

kachawla: Strange that Dapr's support is semi-deprecated, since most NoSQL databases support secondary indexes and complex queries at this point.

Yeah, totally ok to discuss this in detailed design.

rynowak: That would imply that all services use the same database, and read/write each-other's data. This is the opposite of microservices, and should be avoided. We should use the public API.

rynowak: I'll take this up as a topic as part of a follow up doc.

rynowak: > Strange that Dapr's query support is semi-deprecated, since most NoSQL databases support secondary indexes and complex queries at this point.


Yup. I think they started building query support and realized it was hard to implement well across all of the stores they support. There have been some additional proposals to revive it or solve the problem in a different way, but none have gotten serious traction.

architecture/2024-05-radius-on-dapr.md
rynowak: Let's discuss this again next time we meet. It's the primary topic of the document.

We're going to add **some** kind of dependency to Radius, that's not up for debate.

architecture/2024-05-radius-on-dapr.md
```diff
@@ -0,0 +1,324 @@
...
+
+Dapr is well-supported both inside and outside of Kubernetes. Adopting a technology like Dapr will help Radius work well regardless of the hosting scenario.
+
+This document will outline and adoption plan for us to leverage Dapr inside Radius' implementation.
+
+## Terms and definitions
+
+*These terms are part of the Dapr project, not defined by us as part of Radius.*
+
+**This dociment assumes some level of familiarity with Dapr. I'd suggest anyone new to Dapr stop reading this document and read the [Dapr conceptual overview](https://docs.dapr.io/concepts/overview/) before proceeding.**
```
Reshrahim, line 20:
```suggestion
**This document assumes some level of familiarity with Dapr. I'd suggest anyone new to Dapr stop reading this document and read the [Dapr conceptual overview](https://docs.dapr.io/concepts/overview/) before proceeding.**
```

ytimocin, line 20: nit: typo
```suggestion
**This document assumes some level of familiarity with Dapr. I'd suggest anyone new to Dapr stop reading this document and read the [Dapr conceptual overview](https://docs.dapr.io/concepts/overview/) before proceeding.**
```

architecture/2024-05-radius-on-dapr/hosting-diagram.png
ytimocin, line 1: Does the Deployment Engine write or read from the **pubsub**? I think, for async work, we are calling the DE based on the message in the queue.

rynowak: FYI I've got some ideas for how we can improve the async worker. For now that topic is out of scope for this doc. 

> Does the Deployment Engine write or read from the pubsub? I think, for async work, we are calling the DE based on the message in the queue.

That's right, we'll use pub/sub to trigger the async worker.

architecture/2024-05-radius-on-dapr.md
```diff
@@ -0,0 +1,381 @@
...
+
+*These are topics in-scope of this document, where we've had enough discussion to make a decision.*
+
+**Q: Can we use the Dapr state store as a database? What are the tradeoffs?**
+
+The Dapr state store building block provides the capabilities of a generic key-value-store. This is lower level than a document store like MongoDB or a RDBMS like PostgreSQL. 
+
+This isn't bad, we designed Radius around a key-value-store abstraction. Our existing database abstraction is similar to Dapr's state store. It's a good fit.
+
+There's one minor issue to address which is how to implement "list" operations (eg: list all containers in resource group X). Dapr state store doesn't have list functionality (though most key-value-stores do). The well-known solution for this is to maintain our own "index" as it's own key. This is a common enough solution that we shouldn't be concerned.
```
ytimocin, line 323: Do we have this functionality right now? 🧐 I thought we didn't. Both ways, it could be a good challenge to solve.

kachawla, line 323: Are there any examples we can refer to around how other Dapr customers have adopted it and any gotchas when their requirements grew?

architecture/2024-05-radius-on-dapr.md
```diff
@@ -0,0 +1,381 @@
...
+
+*Follow-up: Why don't we use the Query building block in Dapr?*
+
+The query building-block is [deprecated](https://github.com/dapr/dapr/issues/7338#issuecomment-2132282783), and supported by a limited set of state stores. The query building-block will not see more investment, we should not take a dependency.
+
+----
+
+*Follow-up: What do we miss out on since we're not using an RDBMS?*
+
+RDBMS'es are useful when a system needs rich and flexible query capabilities, and the ability to enforce referential integrity across entities. These aren't requirements that we have in Radius. It's very hard to predict the future, and this design-document does not try.
```
ytimocin, line 335: Did we say that as long as we use Dapr and provide the proxy layer, we can choose anything we want that the Dapr will use behind the scenes? Whether it is a key-value store or RDBMS.

architecture/2024-05-radius-on-dapr.md
```diff
@@ -0,0 +1,381 @@
...
+- Using Dapr pub/sub for notifications.
+- Using Dapr workflows inside our RPs and UCP to implement complex background tasks.
+
+If we wanted Dapr to be optional we'd have to have replacements for these features. Dapr is itself an abstraction-layer, and we'd be building an abstraction-layer for the abstraction-layer. If we did enough work to make Dapr an optional dependency of Radius we wouldn't add capabilities, we'd just add work.
+
+**Q: What if users don't want Dapr?**
+
+We're **definitely** going to add some dependencies on our way to production. Taking a bet on Dapr is a way to scope down the number of different dependencies we add. We should not be concerned about this, other projects with a similar flavor to Radius also include their own dependencies.
+
+The only scenario where a dependency on Dapr impacts users is when:
```
rynowak, line 354: Nit: make it clear that this is an "and".

architecture/2024-05-radius-on-dapr.md
rynowak: Yes we have this functionality now.

architecture/2024-05-radius-on-dapr.md
rynowak: Yes.

rynowak:
Oops, this is ready to be merged!

architecture/2024-05-radius-on-dapr.md
rynowak: I don't know of a Dapr specific example. 

The following description is based on examples I've seen of other KVSes. Keep in mind, I don't want to get deep into it here, this is going to have a follow-up design document.

For an "index", assuming that we're using resource ids as keys:

- Choose a key that won't collide with other valid keys.
- For example `/planes/radius/local/resourcegroups/my-group/providers/applications.core/applications/.index`. The `.index` part ensures we don't collide with the keyspace of valid key names.
- The `.../.index` key will store an array of resource ids that are contained within the scope.
- Each upsert or delete operation on a resource will use a transaction to update both the `.../.index` and the destination resource id.
- List operations will do an N+1 query. 
  - First retrieve the `.../.index` key and read the list of resource ids.
  - Then use the "bulk state" API to read all keys of the matching resource ids.
  - This is kinda an N+1 query, but it's only two roundtrips with the database.

rynowak: We can continue the discussion about this in the next design doc when we pick up the work to integrate Dapr. I don't think we're totally closed on the topic, I just wanted to explain my thinking.

rynowak: Adding more of this info to the doc 👍

lakshmimsft (approved):

kachawla (approved):

PR #53, opened 2024-06-25
Design doc for private bicep registries

vishwahiremat:


recipe/2024-06-private-bicep-registries.md
```diff
@@ -0,0 +1,258 @@
...
+| -------- | ---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| OCI compliant registry | An OCI-compliant registry refers to a container registry that adheres to the specifications set by the Open Container Initiative  |
+| ORAS | ORAS is the de facto tool for working with OCI Artifacts.(https://oras.land/) |
+| Private Bicep Registry | A private Bicep registry is a dedicated repository for storing and managing Bicep modules which is not accessible publicly need to authenticate into it for registry operations.| 
+
+## Objectives
+> **Issue Reference:** https://github.com/radius-project/radius/issues/6917
+
+### Goals
+- Add support to publish bicep recipes to a private bicep registry.
```
rynowak, line 24: Does `rad bicep publish` support this?

recipe/2024-06-private-bicep-registries.md
```diff
@@ -0,0 +1,258 @@
...
+- Enable support to use bicep recipes stored in private bicep registries.
+- Support all OCI compliant private registries.
+
+### Non goals
+
+- Workload identity based authentication(Will be covered as part of different design)
```
rynowak, line 30: Good note on this one. Do we have clarity on what forms of authentication we will support?

recipe/2024-06-private-bicep-registries.md
```diff
@@ -0,0 +1,258 @@
...
+
+### Goals
+- Add support to publish bicep recipes to a private bicep registry.
+- Enable support to use bicep recipes stored in private bicep registries.
+- Support all OCI compliant private registries.
+
+### Non goals
+
+- Workload identity based authentication(Will be covered as part of different design) 
+- Configure private OCI registry credentials in Radius via CLI and API.
```
rynowak, line 31: Can you explain why it's a non-goal?

recipe/2024-06-private-bicep-registries.md
```diff
@@ -0,0 +1,258 @@
...
+```
+
+**Sample Output:**
+
+No change in the output
+
+
+## Design
+### Design details
+Currently, OCI-compliant registries are used to store Bicep recipes, with the ORAS client facilitating operations like publish and pull these recipes from the registries. ORAS provides package auth, enabling secure client authentication to remote registries. Private Registry credentials information i.e username and password must be stored in a Application.Core/secretStores resource and the secret resource ID is added to the recipe configuration.
```
rynowak, line 53:
```suggestion
Currently, OCI-compliant registries are used to store Bicep recipes, with the ORAS client facilitating operations like publish and pull these recipes from the registries. ORAS provides package auth, enabling secure client authentication to remote registries. Private Registry credentials information i.e username and password must be stored in an `Application.Core/secretStores` resource and the secret resource ID is added to the recipe configuration.
```

recipe/2024-06-private-bicep-registries.md
```diff
@@ -0,0 +1,258 @@
...
+
+During the recipe deployment bicep driver checks for secrets information for that container registry the recipe deploying is stored on,  and creates an oras auth client to authenticate the private bicep registries.
+
```
rynowak, line 56: It sounds like there are different parts of Radius that will need to authenticate. 

- The CLI (`rad bicep publish`)
- The Bicep recipe driver 

Do these have different requirements? or the same requirements?

recipe/2024-06-private-bicep-registries.md
```diff
@@ -0,0 +1,258 @@
...
++  @doc(Configuration for Bicep Recipes. Controls how Bicep plans and applies templates as part of Recipe deployment.)
++  bicep?: BicepConfigProperties;
+
+  @doc("Environment variables injected during Terraform Recipe execution for the recipes in the environment.")
+  env?: EnvironmentVariables;
+}
+
++@doc(Configuration for Bicep Recipes. Controls how Bicep plans and applies templates as part of Recipe deployment.)
++model BicepConfigProperties {
++  @doc("Authentication information used to access private bicep registries, which is a map of registry hostname to secret config that contains credential information.")
```
rynowak, line 157: Great docs 👍

recipe/2024-06-private-bicep-registries.md
```diff
@@ -0,0 +1,258 @@
...
+},
+```
+
+### CLI Design
+
+`rad bicep publish` need to be updated to ask the user to provide credential information(i.e username and password) for the private bicep registry user is trying to publish the recipe to.
+
+```
+rad bicep publish --file ./redis-test.bicep --target br:ghcr.io/myregistry/redis-test:v1 --username <user> --password <password>
+```
```
rynowak, line 201: Let's make sure to cover this in the discussion 😆

recipe/2024-06-private-bicep-registries.md
```diff
@@ -0,0 +1,258 @@
...
+    func (d *bicepDriver) FindSecretIDs(ctx context.Context, envConfig recipes.Configuration, definition recipes.EnvironmentDefinition) (secretStoreIDResourceKeys map[string][]string, err error) {
+        ...
+    }
+    ```
+- Update Bicep driver apis i.e execute, delete and getMetadata to use oras auth package for private registry authentication.
+
+### Error Handling
+<!--
+Describe the error scenarios that may occur and the corresponding recovery/error handling and user experience.
+-->
```
rynowak, line 221: This is an important section for a feature like this. Any thoughts?

recipe/2024-06-private-bicep-registries.md
```diff
@@ -0,0 +1,258 @@
...
+-   Update environment controller unit tests to add recipe config.
+-   Adding new unit tests in bicep driver validating recipe config changes and retrieving secrets.
+
+#### Functional tests
+- Add e2e test to verify recipe deployment using a bicep stored in a private bicep registry.
+    - publish a recipe to private ghcr.io/radius-project/private-recipes/<recipe> 
+    - Deploy the recipe as part of the functional test using github app token to authenticate ghcr. 
+
+## Security
+With this design we enable username-password based authentication for OCI compliant registries, we let the users manage secrets. For secret rotation users need to re deploy the Applications.Core/secretStore resource with updated credentials.
```
rynowak, line 236:
```suggestion
With this design we enable username-password based authentication for OCI compliant registries, we let the users manage secrets. For secret rotation users need to re deploy the `Applications.Core/secretStores` resource with updated credentials.
```

recipe/2024-06-private-bicep-registries.md
```diff
@@ -0,0 +1,258 @@
...
+    }
+  }
+  "env": {
+    ...
+  }
+},
+```
+
+#### Different container registry providers provide different authenticate methods to login.
+
```
rynowak, line 97: I'd like to understand more about how this is going to work, and what you think the right scope is. Some questions:

- What's going to be supported in the first iteration of this feature? All of these registries? All of the authentication methods discussed here?
- Do we have to special case any of these registries? Put another way do we need to have "if is ACR" in our code anywhere?
- What do we expect in the secrets for these registries? `username` and `password`? Something else?

rynowak (commented):
Thanks @vishwahiremat - looks like this is off to a good start.

recipe/2024-06-private-bicep-registries.md
vishwahiremat: Here are the 2 forms of authentication we want to support:
1) Username-password based auth, (which i supported by most of the OCI complaint registries)
2) Federated authentication(Azure workload identity/ AWS IRSA) for ACR and ECR.

recipe/2024-06-private-bicep-registries.md
vishwahiremat: today `rad bicep publish` uses the exisitng dockerfile(`~/.dockerfile/config.json`) to authenticate private registry.

recipe/2024-06-private-bicep-registries.md
vishwahiremat: we decided this as future scenario and not in the scope of the initial release.

recipe/2024-06-private-bicep-registries.md
vishwahiremat: Yes its different, for the `recipe driver` we expect the users to have bicep recipe configured with `secretstore` information holding private registry credentials. But for `rad bicep publish` we expect username and password as command line args.

recipe/2024-06-private-bicep-registries.md
vishwahiremat: Updated this section

recipe/2024-06-private-bicep-registries.md
vishwahiremat: - What's going to be supported in the first iteration of this feature? All of these registries? All of the authentication methods discussed here?
        - First iteration of feature will have support for username-password for authentication which enables all the methods discussed here and most of the OCI complaint registries.

-Do we have to special case any of these registries? Put another way do we need to have "if is ACR" in our code anywhere?
         - No,  we don't have to write registry provider specific code with username-password based auth, i foresee we may have azure/aws specific code when we have support for federated auth.
-What do we expect in the secrets for these registries? username and password? Something else?
          - yes its just username, password for now.

recipe/2024-06-private-bicep-registries.md
rynowak: So do we need to add/change anything? or is the current functionality good enough?

recipe/2024-06-private-bicep-registries.md
rynowak: Is 2. going to be a fast-follow after this? eg: are we going to enable username/password and then immediately enable AZWI/IRSA?

recipe/2024-06-private-bicep-registries.md
rynowak: Maybe I misunderstood. Do you have an example of that this "non-goal" would include?

recipe/2024-06-private-bicep-registries.md
```diff
@@ -0,0 +1,269 @@
...
+- Enable support to use bicep recipes stored in private bicep registries.
+- Support all OCI compliant private registries.
+
+### Non goals
+
+- Workload identity based authentication(Will be covered as part of different design) 
+- Configure private OCI registry credentials in Radius via CLI and API.
+
+### User scenarios (optional)
+As an operator I am responsible for maintaining Bicep recipes for use with Radius. Bicep recipes used contains sensitive information, proprietary configurations, or data that should not be shared publicly and its intended for internal use within our organization and have granular control over who can access, contribute to, or modify bicep recipes. So I store the recipes in a private bicep registry. And I would like to use these private registries as template-path while registering a bicep recipe.
```
rynowak, line 34: Ideally not sensitive information 😉. The rest of the explanation is great.

recipe/2024-06-private-bicep-registries.md
```diff
@@ -0,0 +1,269 @@
...
+-   Update environment controller unit tests to add recipe config.
+-   Adding new unit tests in bicep driver validating recipe config changes and retrieving secrets.
+
+#### Functional tests
+- Add e2e test to verify recipe deployment using a bicep stored in a private bicep registry.
+    - publish a recipe to private ghcr.io/radius-project/private-recipes/<recipe> 
+    - Deploy the recipe as part of the functional test using github app token to authenticate ghcr. 
+
+## Security
+With this design we enable username-password based authentication for OCI compliant registries, we let the users manage secrets. For secret rotation users need to re deploy the `Applications.Core/secretStores` resource with updated credentials.
```
sk593, line 247: Are there any security concerns with passing the username/password through the CLI during rad bicep publish

recipe/2024-06-private-bicep-registries.md
kachawla: Would be great to also call out why it will not be covered in initial support.

recipe/2024-06-private-bicep-registries.md
rynowak: These are all good answers 👍 - please make sure they make it into the doc with the same level of clarity.

recipe/2024-06-private-bicep-registries.md
```diff
@@ -0,0 +1,269 @@
...
+@doc("Configuration for Recipes. Defines how each type of Recipe should be configured and run.")
+model RecipeConfigProperties {
+  @doc("Configuration for Terraform Recipes. Controls how Terraform plans and applies templates as part of Recipe deployment.")
+  terraform?: TerraformConfigProperties;
+
++  @doc(Configuration for Bicep Recipes. Controls how Bicep plans and applies templates as part of Recipe deployment.)
++  bicep?: BicepConfigProperties;
+
+  @doc("Environment variables injected during Terraform Recipe execution for the recipes in the environment.")
+  env?: EnvironmentVariables;
```
rynowak, line 152: Unrelated, but .... Is there a reason why `env` is part of the "recipe config" instead of the "terraform recipe config"

recipe/2024-06-private-bicep-registries.md
```diff
@@ -0,0 +1,269 @@
...
+environment.bicep
+```
+"recipeConfig": {
+  "terraform": {
+    ...
+  },
+  "bicep": {
+    "authentication":{
+      "test.azurecr.io":{
+        "secret": "/planes/radius/local/resourcegroups/default/providers/Applications.Core/secretStores/acr-secret"
```
rynowak, line 180: Will anything in this design need to change when we add AZWI or IRSA support? How will we tell the difference between the password case and AZWI casE?

recipe/2024-06-private-bicep-registries.md
```diff
@@ -0,0 +1,269 @@
...
+As an operator I am responsible for maintaining Bicep recipes for use with Radius. Bicep recipes used contains sensitive information, proprietary configurations, or data that should not be shared publicly and its intended for internal use within our organization and have granular control over who can access, contribute to, or modify bicep recipes. So I store the recipes in a private bicep registry. And I would like to use these private registries as template-path while registering a bicep recipe.
+
+
+## User Experience (if applicable)
+**Sample Input:**
+Command to publish recipe to private bicep registry.
+```diff
+-rad bicep publish --file ./redis-test.bicep --target br:ghcr.io/myregistry/redis-test:v1
+
++rad bicep publish --file ./redis-test.bicep --target br:ghcr.io/myregistry/redis-test:v1 --username <user> --password <password>
```
kachawla, line 43: Maybe I'm misunderstanding the context here, but doesn't this command already support publishing to private registries? Why do we need to add username/password here?

recipe/2024-06-private-bicep-registries.md
willtsai: I believe the plan of record is for workload and service identities to be in scope for the initial release.

recipe/2024-06-private-bicep-registries.md
```diff
@@ -0,0 +1,269 @@
...
+
+- Implement FindSecretIDs() function in bicep driver, to retrieve the secret store resource IDs associated with private bicep registries.
+    ```
+    // FindSecretIDs is used to retrieve a map of secretStoreIDs and corresponding secret keys.
+    // associated with the given environment configuration and environment definition.
+    func (d *bicepDriver) FindSecretIDs(ctx context.Context, envConfig recipes.Configuration, definition recipes.EnvironmentDefinition) (secretStoreIDResourceKeys map[string][]string, err error) {
+        ...
+    }
+    ```
+- Update Bicep driver apis i.e execute, delete and getMetadata to use oras auth package for private registry authentication.
```
rynowak, line 217: Found any relevant examples in the ORAS code?

recipe/2024-06-private-bicep-registries.md
```diff
@@ -0,0 +1,269 @@
...
+
+## Objectives
+> **Issue Reference:** https://github.com/radius-project/radius/issues/6917
+
+### Goals
+- Add support to publish bicep recipes to a private bicep registry.
+- Enable support to use bicep recipes stored in private bicep registries.
+- Support all OCI compliant private registries.
+
+### Non goals
```
willtsai, line 28: suggest specifying which non-goals are part of initial feature release and which are future ones

recipe/2024-06-private-bicep-registries.md
```diff
@@ -0,0 +1,269 @@
...
+```
+
+
+### Implementation Details
+#### Portable Resources / Recipes RP (if applicable)
+***Bicep Driver***
+
+- Implement FindSecretIDs() function in bicep driver, to retrieve the secret store resource IDs associated with private bicep registries.
+    ```
+    // FindSecretIDs is used to retrieve a map of secretStoreIDs and corresponding secret keys.
```
sk593, line 211: This is also used by the Terraform driver. Will there need to be changes on the driver/engine side to support the Bicep additions?

recipe/2024-06-private-bicep-registries.md
```diff
@@ -0,0 +1,269 @@
...
+
+### Goals
+- Add support to publish bicep recipes to a private bicep registry.
+- Enable support to use bicep recipes stored in private bicep registries.
+- Support all OCI compliant private registries.
+
+### Non goals
+
+- Workload identity based authentication(Will be covered as part of different design) 
+- Configure private OCI registry credentials in Radius via CLI and API.
```
kachawla, line 31: Without this, how does Radius connect to private registry at recipe deploy time?

recipe/2024-06-private-bicep-registries.md
```diff
@@ -0,0 +1,269 @@
...
+
+  This error can occur if users provide incorrect credential information or omit the Bicep recipe configuration for their private registry, we can add a new error code for this scenario `BicepRegistryAuthFailed`
+  
+  Incorrect credential information:
+  ```
+  NewRecipeError("BicepRegistryAuthFailed", fmt.Sprintf("failed to authenticate the bicep registry %s: %s", <private-oras-registry-name>, <error returned by the oras client>))
+  ```
+  Missing Bicep Recipe Config:
+  ```
+  NewRecipeError("BicepRegistryAuthFailed", fmt.Sprintf("missing recipe config for registry %s: %s", <private-oras-registry-name>, <error returned by the oras client>))
```
rynowak, line 231: Are you confident that we can tell the difference between these two cases?

- You tried to log in, but the credentials were wrong.
- You tried to access the registry anonymously, but credentials are required.

recipe/2024-06-private-bicep-registries.md
```diff
@@ -0,0 +1,269 @@
...
+### Non goals
+
+- Workload identity based authentication(Will be covered as part of different design) 
+- Configure private OCI registry credentials in Radius via CLI and API.
+
+### User scenarios (optional)
+As an operator I am responsible for maintaining Bicep recipes for use with Radius. Bicep recipes used contains sensitive information, proprietary configurations, or data that should not be shared publicly and its intended for internal use within our organization and have granular control over who can access, contribute to, or modify bicep recipes. So I store the recipes in a private bicep registry. And I would like to use these private registries as template-path while registering a bicep recipe.
+
+
+## User Experience (if applicable)
```
willtsai, line 37: there is also a user experience in configuring the secret into the environment, correct?

recipe/2024-06-private-bicep-registries.md
```diff
@@ -0,0 +1,269 @@
...
+
+### Error Handling
+
+- Authorization failure error:
+
+  This error can occur if users provide incorrect credential information or omit the Bicep recipe configuration for their private registry, we can add a new error code for this scenario `BicepRegistryAuthFailed`
+  
+  Incorrect credential information:
+  ```
+  NewRecipeError("BicepRegistryAuthFailed", fmt.Sprintf("failed to authenticate the bicep registry %s: %s", <private-oras-registry-name>, <error returned by the oras client>))
```
rynowak, line 227: nit: we should include the username in the error message. 

When something fails, users try to figure out the difference between:

- I messed up the configuration of Radius.
- I configured Radius correctly but the username/password is wrong.

recipe/2024-06-private-bicep-registries.md
vishwahiremat: here is the oras code block to auth private registries using username, password. 
```
	hostname := url.Hostname()
	repo.Client = &auth.Client{
		Client: retry.DefaultClient,
		Credential: auth.StaticCredential("orasregistry.azurecr.io", auth.Credential{
			Username: "<username>",
			Password: "password",
		}),
	}
	repo.Client = client
```

recipe/2024-06-private-bicep-registries.md
willtsai: I think what's missing is the functionality to configure the secret into the environment so that you don't need the creds in the cli command -- this is how it should work rather than requiring the user to insert creds into their cli

recipe/2024-06-private-bicep-registries.md
```diff
@@ -0,0 +1,269 @@
...
+
+
+## Design
+### Design details
+Currently, OCI-compliant registries are used to store Bicep recipes, with the ORAS client facilitating operations like publish and pull these recipes from the registries. ORAS provides package auth, enabling secure client authentication to remote registries. Private Registry credentials information i.e username and password must be stored in an `Application.Core/secretStores` resource and the secret resource ID is added to the recipe configuration.
+
+During the recipe deployment bicep driver checks for secrets information for that container registry the recipe deploying is stored on,  and creates an oras auth client to authenticate the private bicep registries.
+
+
+e.g: creating a secret store with keys "username" and "password".
```
kachawla, line 58: Is it always required to be a username/password or do we envision other forms of auth with different key/values expected as well?

recipe/2024-06-private-bicep-registries.md
```diff
@@ -0,0 +1,269 @@
...
+`rad bicep publish` need to be updated to ask the user to provide credential information(i.e username and password) for the private bicep registry user is trying to publish the recipe to.
+
+```
+rad bicep publish --file ./redis-test.bicep --target br:ghcr.io/myregistry/redis-test:v1 --username <user> --password <password>
+```
+
+
+### Implementation Details
+#### Portable Resources / Recipes RP (if applicable)
+***Bicep Driver***
```
kachawla, line 207: Will there be any changes required in the engine? Anything currently hardcoded for Terraform that needs to be updated?

recipe/2024-06-private-bicep-registries.md
```diff
@@ -0,0 +1,269 @@
...
+    - Updating unit tests.
+- Task 2:
+    - Adding changes to bicep driver to retrieve secrets.
+    - Adding changes to bicep driver to use oras auth package to authenticate private bicep registries
+    - Update/Add unit tests
+- Task 3:
+    - Adding cli changes for `rad bicep publish` to support private bicep registries.
+    - Update the cli unit tests.
+- Task 3:
+    - Manual Validation and adding e2e tests to verify using private bicep registries
```
willtsai, line 262: also need to update and add developer documentation.

recipe/2024-06-private-bicep-registries.md
```diff
@@ -0,0 +1,269 @@
...
+
++rad bicep publish --file ./redis-test.bicep --target br:ghcr.io/myregistry/redis-test:v1 --username <user> --password <password>
+```
+
+**Sample Output:**
+
+No change in the output
+
+
+## Design
```
kachawla, line 51: Can you please also add a high-level overview summarizing what's changing along with an architecture diagram for any new interactions with internal and external components being added?

recipe/2024-06-private-bicep-registries.md
vishwahiremat: as discussed, not making changed to `rad bicep publish`

recipe/2024-06-private-bicep-registries.md
vishwahiremat: updated the doc

recipe/2024-06-private-bicep-registries.md
vishwahiremat: as discussed, this was about managing configuring secrets via CLI/ API . We are doing the same through bicep

recipe/2024-06-private-bicep-registries.md
vishwahiremat: added it here

recipe/2024-06-private-bicep-registries.md
vishwahiremat: as discussed, not adding changes to `rad bicep publish`

recipe/2024-06-private-bicep-registries.md
vishwahiremat: Added the high level flow and an arch diagram

recipe/2024-06-private-bicep-registries.md
vishwahiremat: yes, it is always username and password

recipe/2024-06-private-bicep-registries.md
vishwahiremat: no, we mostly have added all the required changes as part of private terraform repository support.

recipe/2024-06-private-bicep-registries.md
vishwahiremat: updated it

recipe/2024-06-private-bicep-registries.md
vishwahiremat: added it

recipe/2024-06-private-bicep-registries.md
vishwahiremat: yes, the first case if credentials provided are incorrect. and secod is when secrets are not configured.

recipe/2024-06-private-bicep-registries.md
```diff
@@ -0,0 +1,266 @@
...
+- Bicep driver uses ORAS auth client to authenticate the private container registry and fetches the bicep file contents and calls UCP to deploy recipe.
+
+
+
+### Design details
+Currently, OCI-compliant registries are used to store Bicep recipes, with the ORAS client facilitating operations like publish and pull these recipes from the registries. ORAS provides package auth, enabling secure client authentication to remote registries. Private Registry credentials information i.e username and password must be stored in an `Application.Core/secretStores` resource and the secret resource ID is added to the recipe configuration.
+
+During the recipe deployment bicep driver checks for secrets information for that container registry the recipe deploying is stored on,  and creates an oras auth client to authenticate the private bicep registries.
+
+#### Different container registry providers provide different authenticate methods to login.
```
rynowak, line 91: It might be more accurate to say that different registry providers use the username and password differently.

recipe/2024-06-private-bicep-registries.md
```diff
@@ -0,0 +1,266 @@
...
+```
+Username: Use '00000000-0000-0000-0000-000000000000' (a placeholder indicating token-based authentication).
+Password: Use the token obtained from az acr login.
+```
+
+You can obtain the token using the Azure CLI:
+```
+az acr login --name <registry-name> --expose-token
+```
+***Github***:
```
rynowak, line 120: nit: there's a mix of headings `### Foo` and bold text `***Github***:`

recipe/2024-06-private-bicep-registries.md
```diff
@@ -0,0 +1,266 @@
...
+`bicep` recipe config is added in the environment to stores bicep recipe configurations, update it to point to the `secretStore` with credential information.
+```
+"recipeConfig": {
+  "terraform": {
+    ...
+  },
+  "bicep": {
+    "authentication":{
+      "test.azurecr.io":{
+        "secret": "/planes/radius/local/resourcegroups/default/providers/Applications.Core/secretStores/acrsecrets"
```
rynowak, line 65: Will we need to make changes to this when we add AZWI support?

rynowak: How will our code tell the difference between AZWI and username+password?

recipe/2024-06-private-bicep-registries.md
```diff
@@ -0,0 +1,266 @@
...
+Username: AWS
+Password: output of `aws ecr get-login-password --region region`
+```
+While there are various container registry providers, each offering multiple authentication methods, by leveraging ORAS, we abstract away the need for provider-specific code to handle authentication. 
+
+ORAS auth client example to authenticate private registry using username and password,
+```
+	repo.Client = &auth.Client{
+		Client: retry.DefaultClient,
+		Credential: auth.StaticCredential("orasregistry.azurecr.io", auth.Credential{
```
ytimocin, line 138: Are we going to change the `auth.StaticCredential` based on the authentication type (mentioned above)?

recipe/2024-06-private-bicep-registries.md
```diff
@@ -0,0 +1,266 @@
...
+
+environment.tsp
+```diff
+@doc("Configuration for Recipes. Defines how each type of Recipe should be configured and run.")
+model RecipeConfigProperties {
+  @doc("Configuration for Terraform Recipes. Controls how Terraform plans and applies templates as part of Recipe deployment.")
+  terraform?: TerraformConfigProperties;
+
++  @doc(Configuration for Bicep Recipes. Controls how Bicep plans and applies templates as part of Recipe deployment.)
++  bicep?: BicepConfigProperties;
```
ytimocin, line 156: Should we have a different name like recipeConfig or something like that? **Bicep** may be too general.

recipe/2024-06-private-bicep-registries.md
lakshmimsft: fyi..updating comment on this to reflect env is part of Recipe config in upcoming PR

recipe/2024-06-private-bicep-registries.md
vishwahiremat: yes we will be using `auth.StaticCredential` for username-password based and federated token based auth

recipe/2024-06-private-bicep-registries.md
vishwahiremat: as discussed , the parent property is already named as recipeConfig, so naming the child only as `Bicep`

recipe/2024-06-private-bicep-registries.md
vishwahiremat: Updated the design to handle different registry auth types.

recipe/2024-06-private-bicep-registries.md
```diff
@@ -0,0 +1,384 @@
...
+| awsIRSA | arn | 
+
+
+```
+resource secretStore 'Applications.Core/secretStores@2023-10-01-preview' = {
+  name: 'acrsecrets'
+  properties:{
+    type: 'generic'
+    resource: 'registry-secrets/acr'
+    data: {
```
ytimocin, line 49: Have we always had `data` here?

recipe/2024-06-private-bicep-registries.md
```diff
@@ -0,0 +1,384 @@
...
+      "grant_type":   {"access_token"},
+      "service":      <acr-url>,
+      "tenant":       <tenant-id>,
+      "access_token": <aad-token>,
+    }
+    
+    // jsonResponse contains the refresh token from ACR.
+    jsonResponse, err := http.PostForm("https://<acr-url>/oauth2/exchange", formData)
+  ```
+- Use `RefreshToken` in ORAS auth client to authenticate private ACR.
```
kachawla, line 175: Which component will be responsible for these steps?

recipe/2024-06-private-bicep-registries.md
```diff
@@ -0,0 +1,384 @@
...
++
++  @doc("azureFederatedIdentity type is used to represent registry authentication using azure federated identity and the secretstore resource is expected to have the keys 'clientId' and 'tenantId'")
++  azureFederatedIdentity
++
++  @doc("awsIRSA type is used to represent registry authentication using AWS IRSA.")
++  awsIRSA
+}
+
+```
+I propose adding `type` property to the `RegistrySecretConfig` i.e Option 1, as with the current implementation of the secrets flow between engine-driver,  Option 2 lead to either having driver specific code in engine/secretLoader or driver an making api call to corerp to get secrets information.
```
willtsai, line 279: I think presenting the full pros and cons of each option can help the reader fully understand your rationale for recommending option 1.

recipe/2024-06-private-bicep-registries.md
```diff
@@ -0,0 +1,384 @@
...
+
+**AWS**:
+
+Similar to Azure Federated Identity, users can leverage AWS IAM Roles for Service Accounts (IRSA) to authenticate seamlessly to Amazon Elastic Container Registry (ECR). Users are expected to provide `arn` as secret in secretStore resources referenced in the recipe config.
+- Retrieve the ECR Authorization token using the `arn` configured with AWS IRSA.
+- And use that as a access token with ORAS auth client
+  ```
+  repo.Client = &auth.Client{
+		Client: retry.DefaultClient,
+		Credential: auth.StaticCredential(<ecr-url>, auth.Credential{
```
kachawla, line 193: Could you please add a link to relevant api docs for oras, just for reference?

kachawla, line 193: What are the values for `ecr-url` and `acr-url`? Are they constant or change based on region/partition etc? How do we plan to manage them?

recipe/2024-06-private-bicep-registries.md
```diff
@@ -0,0 +1,384 @@
...
+		}),
+	}
+	repo.Client = client
+```
+
+#### Authentication using Federated identity
+**Azure**:
+
+Users can leverage Azure Federated Identity to authenticate seamlessly. When Azure Federated Identity is set up on a Kubernetes cluster, it enables secure and managed authentication, allowing workloads running on the cluster to access ACR without needing to manage service principal credentials directly. This approach enhances security by eliminating the need for hardcoded credentials and simplifies the authentication process by relying on Azure Active Directory (AAD) token issuance for the cluster's managed identity.
+
```
willtsai, line 159: can you elaborate a bit on the specific user experience? from the content here i can see that they need to configure some federated credential to their environment, anything else they'll need to do on the radius side to allow for the authentication to succeed?

recipe/2024-06-private-bicep-registries.md
```diff
@@ -0,0 +1,384 @@
...
+***Bicep Example***
+
+environment.bicep
+```
+"recipeConfig": {
+  "terraform": {
+    ...
+  },
+  "bicep": {
+    "authentication":{
```
willtsai, line 290: should these actually be classified as secrets or can they be a little less strict to allow for better troubleshooting for the user?

recipe/2024-06-private-bicep-registries.md
```diff
@@ -0,0 +1,384 @@
...
+  certificate,
+
++  @doc("basicAuthentication type is used to represent username and password based authentication and the secretstore resource is expected to have the keys 'username' and 'password'")
++  basicAuthentication
++
++  @doc("azureFederatedIdentity type is used to represent registry authentication using azure federated identity and the secretstore resource is expected to have the keys 'clientId' and 'tenantId'")
++  azureFederatedIdentity
++
++  @doc("awsIRSA type is used to represent registry authentication using AWS IRSA.")
++  awsIRSA
```
kachawla, line 275: Could you expand on what would a user specify within these types?

kachawla, line 275: Have we considered either making it more generic - use `federatedIdentity` for both aws and azure, or make it more specific `azureWorkloadIdentity` because that's what we are supporting. @willtsai @reshmarahim thoughts?

Reshrahim: I think making it more specific to the cloud providers like `azureWorkloadIdentity` and `awsIRSA` would be easier for the users to interpret. Also expanding IRSA to IAM Roles for Service accounts in product/code docs would be helpful.

recipe/2024-06-private-bicep-registries.md
vishwahiremat: yes, the secrets key-values are added as part of data.

recipe/2024-06-private-bicep-registries.md
vishwahiremat: updated the doc with pros and cons of each option

recipe/2024-06-private-bicep-registries.md
vishwahiremat: updated the doc on steps users need to do on the radius side to enable workload identity/ aws irsa

recipe/2024-06-private-bicep-registries.md
vishwahiremat: updated the doc with details on secret keys for awsIRSA

recipe/2024-06-private-bicep-registries.md
```diff
@@ -0,0 +1,384 @@
...
+      "service":      <acr-url>,
+      "tenant":       <tenant-id>,
+      "access_token": <aad-token>,
+    }
+    
+    // jsonResponse contains the refresh token from ACR.
+    jsonResponse, err := http.PostForm("https://<acr-url>/oauth2/exchange", formData)
+  ```
+- Use `RefreshToken` in ORAS auth client to authenticate private ACR.
+  ```
```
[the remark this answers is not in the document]
vishwahiremat: as discussed in the design meeting, the `type` parameter will be used to determine if we are using token or usernam-password properties while creating auth client

recipe/2024-06-private-bicep-registries.md
kachawla: Can you please update the architecture diagram to reflect this call flow?

recipe/2024-06-private-bicep-registries.md
```diff
@@ -0,0 +1,425 @@
...
+  "env": {
+    ...
+  }
+},
+```
+
+
+## Design
+![Alt text](./2024-06-private-bicep-registries/overview.png)
+
```
rynowak, line 83: Good diagram.

recipe/2024-06-private-bicep-registries.md
```diff
@@ -0,0 +1,425 @@
...
+      "grant_type":   {"refresh_token"},
+      "service":      test.azurecr.io,
+      "tenant":       <tenant-id>,
+      "access_token": <aad-token>,
+      "expires_in": <5 mins in seconds> 
+    }
+    
+    // jsonResponse contains the refresh token from ACR.
+    jsonResponse, err := http.PostForm("https://test.azurecr.io/oauth2/exchange", formData)
+  ```
```
rynowak, line 180: A few questions about this ... does MSAL provide a library for this? I'm a little bit anxious about doing the raw HTTP calls ourselves if we don't have to since it's security sensitive.

Is this based on a sample the ACR team provided? or just something we figured out

recipe/2024-06-private-bicep-registries.md
```diff
@@ -0,0 +1,425 @@
...
+		Credential: auth.StaticCredential(test.azurecr.io, auth.Credential{
+			RefreshToken: <refresh-token>,
+		}),
+	}
+  ```
+
+**AWS**:
+
+Similar to Azure Federated Identity, users can leverage AWS IAM Roles for Service Accounts (IRSA) to authenticate seamlessly to Amazon Elastic Container Registry (ECR). Users are expected to provide `rolearn` as secret in secretStore resources referenced in the recipe config.
+
```
rynowak, line 194: Same question for AWS - any bespoke client code we have to write? or do their SDKs do what we need?

recipe/2024-06-private-bicep-registries.md
```diff
@@ -0,0 +1,425 @@
...
+  ```
+  rad install kubernetes --set global.awsIRSA.enabled=true
+  ```
+  or by configuring aws provider with AWS Credential kind `IRSA` using `rad init --full` command.
+- Retrieve the ECR Authorization token using the `roleARN` configured with AWS IRSA.
+- And use that as a access token with ORAS auth client
+  ```
+  repo.Client = &auth.Client{
+		Client: retry.DefaultClient,
+		Credential: auth.StaticCredential(123456789012.dkr.ecr.us-west-2.amazonaws.com, auth.Credential{
```
kachawla, line 205: `123456789012.dkr.ecr.us-west-2.amazonaws.com` - how do we get this value, recipe's template path?

recipe/2024-06-private-bicep-registries.md
```diff
@@ -0,0 +1,425 @@
...
++
++  @doc("Authentication information used to access private bicep registries, which is a map of registry hostname to secret config that contains credential information.")
++  authentication?: Record<RegistrySecretConfig>;
+}
+```
+`RegistrySecretConfig` contains registry credential information, we also want users to specify the type of credentials, e.g. username-password, federated credentials.
+
+Option 1: 
+
+Adding a `type` property to `RegistrySecretConfig`
```
rynowak, line 239: We'd usually call this `kind` to avoid overloading the term `type`. Every resource has a `type` that returns it's resource type, so we try to avoid using that field name for other purposes.

recipe/2024-06-private-bicep-registries.md
```diff
@@ -0,0 +1,425 @@
...
++  @doc("awsIRSA type is used to represent registry authentication using AWS IRSA.")
++  awsIRSA
+}
+```
+Pros:
+- Leveraging the existing property `type` in `Application.core/secretStores` to specify the kind of authentication.
+- Incorrect configurations(e.g: clientId is not provided as a secret key for type azureWorkloadIdentity) can be detected during the creation of the `Applications.Core/secretStore`.
+
+Cons:
+- This options lead to either having driver specific code in engine/secretLoader or driver an making api call to corerp to get secrets information.
```
rynowak, line 300: Can you explain this more? What's the driver-specific code that needs to go into the engine?

Isn't the API call to corerp needed for both cases?

recipe/2024-06-private-bicep-registries.md
```diff
@@ -0,0 +1,425 @@
...
+
+I propose adding `type` property to the `RegistrySecretConfig` i.e Option 1, as with the current implementation of the secrets flow between engine-driver,  Option 2 lead to either having driver specific code in engine/secretLoader or driver an making api call to corerp to get secrets information. 
+
```
rynowak, line 303: I don't think that either option is bad, but I'm curious to hear what was discussed last week.

recipe/2024-06-private-bicep-registries.md
```diff
@@ -0,0 +1,425 @@
...
+  azureWorkloadIdentity
+
+  @doc("awsIRSA type is used to represent registry authentication using AWS IRSA(IAM Roles for Service accounts) and the secretstore resource is expected to have the keys 'roleARN'")
+  awsIRSA
+}
+
+@doc("Registry Secret Configuration used to authenticate to private bicep registries.")
+model RegistrySecretConfig {
+  @doc("The ID of an Applications.Core/SecretStore resource containing credential information used to authenticate private container registry.The keys in the secretstore depends on the type.")
+  secretSource?: string;
```
rynowak, line 257: Did we call this `secretSource` for Terraform? We should be consistent if possible.

recipe/2024-06-private-bicep-registries.md
vishwahiremat: yes from the templatepath, we get the domain from template path and check for this domain in bicep recipe config to see if secrets are configured.

recipe/2024-06-private-bicep-registries.md
rynowak: Yeah ... we called it `secret` for Terraform: https://github.com/radius-project/design-notes/pull/53/files#diff-12354a738cc93ef8bfee787de4ebb16befdb3c7dccce0df0a0f4e29e523b779dR314

recipe/2024-06-private-bicep-registries.md
```diff
@@ -0,0 +1,425 @@
...
+
+### Goals
+- Enable support to use bicep recipes stored in private bicep registries.
+- Support all OCI compliant private registries.
+
+### Non goals
+- Support to manage(configure/view) OCI registry credentials in Radius environment via CLI. (this is a future priority scenario and would not be in scope for the initial release)
+
+### User scenarios (optional)
+As an operator I am responsible for maintaining Bicep recipes for use with Radius. Bicep recipes used contains proprietary configurations, or data that should not be shared publicly and its intended for internal use within our organization and have granular control over who can access, contribute to, or modify bicep recipes. So I store the recipes in a private bicep registry. And I would like to use these private registries as template-path while registering a bicep recipe.
```
brooke-hamilton, line 31:
```suggestion
As an operator I am responsible for maintaining Bicep recipes for use with Radius. My Bicep recipes contain proprietary configurations, or data that should not be shared publicly, and are intended for internal use within our organization and have granular control over who can access, contribute to, or modify bicep recipes. So I store the recipes in a private bicep registry. And I would like to use these private registries as template-path while registering a bicep recipe.
```

recipe/2024-06-private-bicep-registries.md
```diff
@@ -0,0 +1,425 @@
...
+}
+```
+Pros:
+- Leveraging the existing property `type` in `Application.core/secretStores` to specify the kind of authentication.
+- Incorrect configurations(e.g: clientId is not provided as a secret key for type azureWorkloadIdentity) can be detected during the creation of the `Applications.Core/secretStore`.
+
+Cons:
+- This options lead to either having driver specific code in engine/secretLoader or driver an making api call to corerp to get secrets information.
+
+I propose adding `type` property to the `RegistrySecretConfig` i.e Option 1, as with the current implementation of the secrets flow between engine-driver,  Option 2 lead to either having driver specific code in engine/secretLoader or driver an making api call to corerp to get secrets information.
```
willtsai, line 302: So this means that the user continues to set up their registry auth secrets as `generic` type like they do today? If so, I'm good with this approach.

recipe/2024-06-private-bicep-registries.md
```diff
@@ -0,0 +1,425 @@
...
++      123456789012.dkr.ecr.us-west-2.amazonaws.com:{
++        "secretSource": "/planes/radius/local/resourcegroups/default/providers/Applications.Core/secretStores/aws-secret",
++        "type":"awsIRSA"
++      },
++      ghcr.io:{
++        "secretSource": "/planes/radius/local/resourcegroups/default/providers/Applications.Core/secretStores/ghcr-secret",
++        "type": "basicAuthentication"
++      },
++    }
++  }
```
willtsai, line 336: will this become a breaking change for users who have already set up [private Terraform repos](https://docs.radapp.io/guides/recipes/terraform/howto-private-registry/)?

recipe/2024-06-private-bicep-registries.md
```diff
@@ -0,0 +1,425 @@
...
+
+  @doc("Environment variables injected during Terraform Recipe execution for the recipes in the environment.")
+  env?: EnvironmentVariables;
+}
+
++@doc(Configuration for Bicep Recipes. Controls how Bicep plans and applies templates as part of Recipe deployment.)
++model BicepConfigProperties {
++
++  @doc("Authentication information used to access private bicep registries, which is a map of registry hostname to secret config that contains credential information.")
++  authentication?: Record<RegistrySecretConfig>;
```
kachawla, line 232: Have we considered using the format used by [TerrafromConfig auth](https://github.com/radius-project/radius/blob/main/typespec/Applications.Core/environments.tsp#L92), to reduce the cognitive load on the users to learn different property names?

recipe/2024-06-private-bicep-registries.md
willtsai, line 239: Will this be Bicep-specific? And is `type` going to be a required field?

recipe/2024-06-private-bicep-registries.md
kachawla: I see, also noticed it here https://github.com/radius-project/design-notes/blob/fc93b704215ff14da8f1e3ab789673ac727dc878/recipe/2024-06-private-bicep-registries.md?plain=1#L327.. is the part `123456789012` account id? If it is then do we want the bicep config to specified at the this granularity level or a level above?

recipe/2024-06-private-bicep-registries.md
```diff
@@ -0,0 +1,425 @@
...
+
+
+## User Experience (if applicable)
+Credential information is stored in `Application.Core/secretstore` resource, users are expected to provide keys based on the `type` of authentication used,  with the appropriate values in the secret store that will be used to authenticate private registry.
+| type     | Expected Keys                                                                                                                                                                                                 |
+| -------- | ---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| basicAuthentication | username, password  |
+| azureFederatedIdentity | clientid, tenantId |
```
lakshmimsft, line 39: Setting this to azureWorkloadIdentity per the typespec changes below. Pls confirm case of required keys. slight differences  eg. clientid versus clientId

recipe/2024-06-private-bicep-registries.md
lakshmimsft, line 302: Document needs to be updated to reflect decision taken between Option 1 & 2

recipe/2024-06-private-bicep-registries.md
```diff
@@ -0,0 +1,497 @@
...
+- Updated the return type of secrets to include the secrets type, and also today in loadSecrets() we would return empty secrets map if the key list is empty, but we update it it return all the secret key-value pairs from secretStore resource if the key list is empty.
+
+```diff
+type SecretsLoader interface {
+
+// LoadSecrets loads secrets from secret stores based on input map of provided secret store IDs and secret keys.
+// It returns a map of secret data, where the keys are the secret store IDs and the values are maps of secret keys and their corresponding values.
+-	LoadSecrets(ctx context.Context, secretStoreIDs map[string][]string) (secretData map[string]map[string]string, err error)
+
++	LoadSecrets(ctx context.Context, secretStoreIDs map[string][]string) (secretData map[string]secretData, err error)
```
lakshmimsft, line 420: LoadSecrets(ctx context.Context, secretStoreKeysFilter map[string][]string) (secretData map[string]SecretData, err error)

recipe/2024-06-private-bicep-registries.md
```diff
@@ -0,0 +1,497 @@
...
+type SecretsLoader interface {
+
+// LoadSecrets loads secrets from secret stores based on input map of provided secret store IDs and secret keys.
+// It returns a map of secret data, where the keys are the secret store IDs and the values are maps of secret keys and their corresponding values.
+-	LoadSecrets(ctx context.Context, secretStoreIDs map[string][]string) (secretData map[string]map[string]string, err error)
+
++	LoadSecrets(ctx context.Context, secretStoreIDs map[string][]string) (secretData map[string]secretData, err error) 
+}
+
++type secretData struct{
```
lakshmimsft, line 423: nit: SecretData

lakshmimsft (approved):

PR #57, opened 2024-07-26
Feature specs: `Applications.Core/secretStores` extended use cases

willtsai:
Here are the feature specs for extending the use cases of `Applications.Core/secretStores`


resources/2024-07-secretstore-feature-spec.md
```diff
@@ -0,0 +1,229 @@
+# Extend the use-cases of `Applications.Core/secretStores`
+
+* **Status**: Pending
+* **Author**: Will Tsai (@willtsai)
+
+## Target customers
```
rynowak, line 6: `s/customer/user/g` please 😁

resources/2024-07-secretstore-feature-spec.md
```diff
@@ -0,0 +1,229 @@
...
+```bicep
+resource demo 'Applications.Core/containers@2023-10-01-preview' = {
+  name: 'demo'
+  properties: {
+    application: application
+    container: {
+      image: 'ghcr.io/radius-project/samples/demo:latest'
+      env:{
+        // validate accuracy of referencing the values from the secret store
+        DB_CONNECTION: authcreds.properties.data.connectionString
```
rynowak, line 97: I might be missing something, is this a description of the current state?

resources/2024-07-secretstore-feature-spec.md
willtsai: good catch, edited!

resources/2024-07-secretstore-feature-spec.md
```diff
@@ -0,0 +1,229 @@
...
+… -->
+
+Step 1: Operator creates and deploys `Applications.Core/secretStores` resources containing the secret data required for authenticating into resources that their developers may use.
+
+```bicep
+resource authcreds 'Applications.Core/secretStores@2023-10-01-preview' = {
+  name: 'authcreds'
+  properties:{
+    application: application
+    type: 'credentials' // this can change based on detailed tech design
```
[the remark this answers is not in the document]
willtsai: Today the implementation is actually an ENUM with possible values `["generic", "certificate"]`

There was some related discussion in https://github.com/radius-project/design-notes/pull/53 where the decision was to extend this ENUM with additional values for their use cases. We have an opportunity to propose a change here to make this `type` field free form to enable better extensibility going forward so that we don't have to keep updating the API each time. What are your thoughts here @superbeeny?

resources/2024-07-secretstore-feature-spec.md
willtsai: This is a description of what the propsed state should be, let me update this to make it more apparent.

resources/2024-07-secretstore-feature-spec.md
```diff
@@ -0,0 +1,229 @@
...
+
+* **Status**: Pending
+* **Author**: Will Tsai (@willtsai)
+
+## Target customers
+**Developers and operators** who are building and managing microservice applications, are using Radius to deploy applications and wish to effortlessly and securely manage the secrets for use in their environments and applications.
+
+## Existing customer problem
+
+Currently, Radius provides an `Applications.Core/secretStores` resource type that allows developers to store and retrieve secrets in a secure and reliable way. The `secretStores` resource type can be referenced and used by the following Radius resources today:
```
rynowak, line 11: One thing that's on my mind is whether `Applications.Core/secretStores` is right. Specifically whether it should be part of `Applications.Core` or whether it should be part of UCP (something like `System.Resources/secretStores`). 

My concern is that secret storage and rotation is a general platform/control-plane concern and not specific to Radius. For example the design of `System.Azure/credentials` and `System.AWS/credentials` isn't aligned with our other usage of secrets and maybe it should be. I see this as an artifact of the order in which we built the features.

I'm also OK to consider this again in the future if we're OK with the current layering.

resources/2024-07-secretstore-feature-spec.md
```diff
@@ -0,0 +1,229 @@
...
+
+## Desired customer experience outcome
+<!-- <Write this as an “I statement” that expresses the new capability from customer perspective … i.e. After this scenario is implemented “I can do, x, y, z, steps in cloud native app developer and seamlessly blah blah blah …. As a result <summarize positive impact on your work / business>  -->
+
+As an operator, I can define an `Applications.Core/secretStores` resource and deploy it along with an Environment so that the developers I support can securely leverage secrets I manage on their behalf for use in their application resources.
+
+As a developer, I can reference an `Applications.Core/secretStores` in my `Applications.Core/containers` resource definition so that Radius will inject secrets as environment variables into my container at deploy time so that credentials can be provided to the container for authentication, etc.
+
+As a developer, I can reference an `Applications.Core/secretStores` in my `Applications.Extenders` or `Applications.Volumes` resource definition so that I no longer have to store secrets as plain text in the `properties.secrets` field of my resource for authentication, etc.
```
rynowak, line 26: Let's make sure to cover the difference between:

- My application code needs some secret values set in environment variables.
- My application code needs to read some secret values from the disk.

These are different ways to accomplish the same thing. The reason why I'm calling it out is that the developer of the code already made a decision which one they want, our job is to do what they say.

resources/2024-07-secretstore-feature-spec.md
```diff
@@ -0,0 +1,229 @@
...
+
+As an application developer, I cannot reference a `secretStores` resource in the `Applications.Core/containers`, `Applications.Core/extenders`, `Applications.Core/volumes`, `Applications.Datastores/*`, and `Applications.Messaging/*` resource types to securely manage secrets for use in my application. Instead, I'm having to store my secrets in plain text within the `properties.secrets` field for each resource. This is a critical gap in the secret management capabilities of Radius that is hindering my ability to securely manage secrets for use in my containers.
```
rynowak, line 17: > Instead, I'm having to store my secrets in plain text within the `properties.secrets` field for each resource.

We actually do this correctly for extender today. It's the Dapr resources that are wrong.

rynowak: I'm hoping we change the design for secret handling all-up for Radius as part of this effort and the UDT effort.

resources/2024-07-secretstore-feature-spec.md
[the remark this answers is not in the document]
rynowak: We picked `generic` over `opaque` because we found it much clearer.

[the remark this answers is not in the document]
rynowak: Something else that's needed here is define a naming scheme for the values of `type`. What should the names be for AzWI? for AWS IRSA? For the secret created by an `Applications.Datastores/redisCaches` resource?

resources/2024-07-secretstore-feature-spec.md
rynowak: Update please 👍 this doesn't look like @superbeeny 's PR.

resources/2024-07-secretstore-feature-spec.md
willtsai: I think this scenario will mostly focus on reading some secret values from disk, right? For the case of needing secret values stored as env variables, the scenario covered here will be storing secrets securely on disk and then reading those secrets from disk to either use them directly for auth or else injecting them into env vars. In other words, for "application code needs some secret values set in environment variables", the ability to inject env vars into a container and then for an app to subsequently read those env vars exists already regardless of whether or not we extend the `secretStores` use cases.

resources/2024-07-secretstore-feature-spec.md
willtsai: good catch, updated this and all other proposed examples to align with @superbeeny PR (https://github.com/radius-project/radius/pull/7744)

resources/2024-07-secretstore-feature-spec.md
```diff
@@ -0,0 +1,286 @@
...
+      }
+      'version': {
+        value: 1
+      }
+      'encoding': {
+        value: 'base64'
+      }
+      'alias': {
+        value: secretalias
+      }
```
kachawla, line 75: I couldn't tell how it references from azure keyvault, could you please elaborate on that?

resources/2024-07-secretstore-feature-spec.md
```diff
@@ -0,0 +1,286 @@
...
+resource twilio 'Applications.Core/extenders@2023-10-01-preview' = {
+  name: 'twilio'
+  properties: {
+    application: application
+    environment: environment
+    recipe: {
+      name: 'twilio'
+    }
+    secrets: {
++        password: {
```
kachawla, line 124: Are you proposing a new property "secrets" to specify recipe's input parameters that contain sensitive data? If so, I think this should be nested under recipe->parameters to make it more intuitive and consistent with rest of the parameters.

resources/2024-07-secretstore-feature-spec.md
```diff
@@ -0,0 +1,286 @@
...
+    host: substring(cosmosAccount.properties.documentEndpoint, 0, lastIndexOf(cosmosAccount.properties.documentEndpoint, ':'))
+    port: int(split(substring(cosmosAccount.properties.documentEndpoint,lastIndexOf(cosmosAccount.properties.documentEndpoint, ':') + 1), '/')[0])
+    database: cosmosAccount::database.name
+    username: ''
+    resources: [
+      { id: cosmosAccount.id }
+    ]
+    secrets: {
++        connectionString: {
++          valueFrom: {
```
kachawla, line 202: We should revisit the overall plan for provisioning types we plan to support on portable resources before implementing this change. There are some changes with UDT that haven't been solidified yet.

resources/2024-07-secretstore-feature-spec.md
```diff
@@ -0,0 +1,286 @@
...
+<!-- Dependency Name – summary of dependency.  Issues/concerns/risks with this dependency -->
+<!-- Risk Name – summary of risk.  Mitigation plan if known. If it is not yet known, no problem. -->
+
+**Dependency: ability to reference values from `Applications.Core/secretStores`.** This feature is dependent on the ability to reference values from `Applications.Core/secretStores`, which should be doable given the existing functionality of referencing secret values for TLS Termination in `Applications.Core/gateways`.
+
+> The TLS certificate data secret in `Applications.Core/gateways` is referenced today by `tls: { certificateFrom: secretstore.id }`. As a part of implementation we should evaluate if the `valueFrom: { secretRef: { ... } }` pattern proposed here is an acceptable deviation from the previous pattern implemented for `gateways`.
+
+**Risk: use of secrets in core and portable resources.** This pattern of referencing and leveraging secrets in core and portable resources might not be a common or desirable pattern for users. We will validate this by opening up this feature spec for discussion with the community.
```
kachawla, line 275: Not sure if I follow this, are there alternatives that we think might be more desirable?

resources/2024-07-secretstore-feature-spec.md
```diff
@@ -0,0 +1,286 @@
...
+
+As an application developer, I cannot reference a `secretStores` resource in the `Applications.Core/containers`, `Applications.Core/extenders`, `Applications.Core/volumes`, `Applications.Datastores/*`, and `Applications.Messaging/*` resource types to securely manage secrets for use in my application. Instead, I'm having to store my secrets in plain text within the `properties.secrets` field for each resource. This is a critical gap in the secret management capabilities of Radius that is hindering my ability to securely manage secrets for use in my containers.
+
```
lakshmimsft, line 18: Requesting more info (as discussed) on adding more information on customers bringing in their own secrets, secrets stores resource supporting only k8s resource.

kachawla:
@willtsai looks like this got auto closed. Are there any pending updates or is this PR mainly pending approval at this point?

resources/2024-07-secretstore-feature-spec.md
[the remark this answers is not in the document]
willtsai: this has been defined here already: https://github.com/radius-project/radius/pull/7816

resources/2024-07-secretstore-feature-spec.md
willtsai: @kachawla yes, good catch, in this case I missed adding the reference to an external resource name, so I've added `resource: 'secret-app-existing-secret'` to the example. Hopefully that clears it up.

resources/2024-07-secretstore-feature-spec.md
```diff
@@ -0,0 +1,286 @@
+# Extend the use-cases of `Applications.Core/secretStores`
```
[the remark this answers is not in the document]
willtsai: good question - that's currently a missing scenario, I'll be sure to add it!

resources/2024-07-secretstore-feature-spec.md
willtsai: added this as a note to the feature spec

resources/2024-07-secretstore-feature-spec.md
willtsai: Brooke pointed out some similar feedback, which is something I hadn't considered. I'll add it as an additional scenario (referencing secrets in recipes)

resources/2024-07-secretstore-feature-spec.md
```diff
@@ -0,0 +1,286 @@
...
+## Key investments
+<!-- List the features required to enable this scenario. -->
+
+### Feature 1: Add functionality to reference `Applications.Core/secretStores` in `Applications.Core/containers` resources
+<!-- One or two sentence summary -->
+Add the ability for developers to reference values from their `Applications.Core/secretStores` resources in their `Applications.Core/containers` resource definitions under the `properties.container.env` field so that secrets can be injected as environment variables into their container at deploy time.
+
+### Feature 2: Add functionality to reference `Applications.Core/secretStores` in `Applications.Core/*` resources
+<!-- One or two sentence summary -->
+Add the ability for developers to reference values from their `Applications.Core/secretStores` resources in their core resources (namely `Applications.Core/extenders` and `Applications.Core/volumes`) so that secrets can be securely managed for use in their application to authenticate into those resources.
```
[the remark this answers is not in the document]
willtsai: I bucketed it under Core resources because of its `Application.Core` namespace, do you think it's misnamed in this case?

resources/2024-07-secretstore-feature-spec.md
willtsai: no, there are no more desirable alternatives that we can think of, but I just added this as a risk so that we can explicitly call it out and source feedback on it

resources/2024-07-secretstore-feature-spec.md
willtsai: @lakshmimsft - does this address your concern? https://github.com/radius-project/design-notes/pull/57/files#diff-b9ff2f984ca28909884ce7f981fca555ab16b7c3f52afae2bb14bdd05c83a6f0R63-R89

resources/2024-07-secretstore-feature-spec.md
```diff
@@ -0,0 +1,417 @@
...
+
+As an operator, I can create a `secretStores` resource to securely manage secrets for use in the Radius Environments I provide for my developers, which is handy to allow for authentication into private Recipe registries and custom Terraform Providers. However, the `secretStores` resource type is not yet supported by the `Applications.Core/containers`, `Applications.Core/extenders`, `Applications.Core/volumes`, `Applications.Datastores/*`, and `Applications.Messaging/*` resource types. Thus, I have to do extra work to inject secrets as environment variables by configuring each custom Recipe in order to propagate secrets to my application developers.
+
+As an operator, I cannot reference a `secretStores` resource in the Recipes I create in order to securely manage secrets to be used in the Recipe configurations. Instead, I'm having to store my secrets in plain text within the `secrets` field of my Recipe's `output` object. This is a critical gap in the secret management capabilities of Radius that is hindering my ability to securely manage secrets for use in my Recipes.
```
rynowak, line 16: We're planning to address this as part of the UDT work. 

Today there's a large set of resources that _directly_ store secret data. When we're done it will all be stored in `secretStores` resources and _referenced_ from the places that need it.

resources/2024-07-secretstore-feature-spec.md
```diff
@@ -0,0 +1,417 @@
...
+
+As an operator, I can reference an `Applications.Core/secretStores` in my Recipes so that I no longer have to store secrets as plain text in the `secrets` field of my Recipe's `output` object for authentication, etc.
+
+As a developer, I can reference an `Applications.Core/secretStores` in my `Applications.Core/containers` resource definition so that Radius will inject secrets as environment variables into my container at deploy time so that credentials can be provided to the container for authentication, etc.
+
+As a developer, I can reference an `Applications.Core/secretStores` in my `Applications.Extenders` or `Applications.Volumes` resource definition so that I no longer have to store secrets as plain text in the `properties.secrets` field of my resource for authentication, etc.
+
+As a developer, I can reference an `Applications.Core/secretStores` in my `Applications.Datastores/*` or `Applications.Messaging/*` resource definition so that I no longer have to store secrets as plain text in the `properties.secrets` field of my resource for authentication, etc.
```
ytimocin, line 30: I know this is for later, but do we want the user-defined providers and types to be able to user **secretStores**?

resources/2024-07-secretstore-feature-spec.md
```diff
@@ -0,0 +1,417 @@
...
+
+As an application developer, I cannot reference a `secretStores` resource in the `Applications.Core/containers`, `Applications.Core/extenders`, `Applications.Core/volumes`, `Applications.Datastores/*`, and `Applications.Messaging/*` resource types to securely manage secrets for use in my application. Instead, I'm having to store my secrets in plain text within the `properties.secrets` field for each resource. This is a critical gap in the secret management capabilities of Radius that is hindering my ability to securely manage secrets for use in my containers.
+
+## Desired user experience outcome
+
+As an operator, I can define an `Applications.Core/secretStores` resource and deploy it along with an Environment so that the developers I support can securely leverage secrets I manage on their behalf for use in their application resources.
+
+As an operator, I can reference an `Applications.Core/secretStores` in my Recipes so that I no longer have to store secrets as plain text in the `secrets` field of my Recipe's `output` object for authentication, etc.
```
rynowak, line 24: Does this document include the design for **how** users will do this?

resources/2024-07-secretstore-feature-spec.md
```diff
@@ -0,0 +1,417 @@
...
+
+As a developer, I can reference an `Applications.Core/secretStores` in my `Applications.Core/containers` resource definition so that Radius will inject secrets as environment variables into my container at deploy time so that credentials can be provided to the container for authentication, etc.
+
+As a developer, I can reference an `Applications.Core/secretStores` in my `Applications.Extenders` or `Applications.Volumes` resource definition so that I no longer have to store secrets as plain text in the `properties.secrets` field of my resource for authentication, etc.
```
rynowak, line 28: Does this document include the design for each of the places we need to update?

rynowak: I think the volumes scenario is different. The use-case for extenders belongs in the next paragraph.

resources/2024-07-secretstore-feature-spec.md
```diff
@@ -0,0 +1,417 @@
...
+
+As an operator, I cannot reference a `secretStores` resource in the Recipes I create in order to securely manage secrets to be used in the Recipe configurations. Instead, I'm having to store my secrets in plain text within the `secrets` field of my Recipe's `output` object. This is a critical gap in the secret management capabilities of Radius that is hindering my ability to securely manage secrets for use in my Recipes.
+
+As an application developer, I cannot reference a `secretStores` resource in the `Applications.Core/containers`, `Applications.Core/extenders`, `Applications.Core/volumes`, `Applications.Datastores/*`, and `Applications.Messaging/*` resource types to securely manage secrets for use in my application. Instead, I'm having to store my secrets in plain text within the `properties.secrets` field for each resource. This is a critical gap in the secret management capabilities of Radius that is hindering my ability to securely manage secrets for use in my containers.
+
+## Desired user experience outcome
+
```
rynowak, line 21: I think we can simplify this into three cases:

- Secrets are stored and then mounted as environment-variables in a container.
- Secrets are stored and then mounted as files-on-disk in a container.
- Secrets are stored and used internally by the Radius infrastructure or some other piece of automation (not application code).

resources/2024-07-secretstore-feature-spec.md
```diff
@@ -0,0 +1,417 @@
...
+Step 1: Operator creates or references an existing `Applications.Core/secretStores` resource containing the secret data required for authenticating into resources that their developers may use.
+
+Creating a new `secretStore` resource: 
+
+```bicep
+resource authcreds 'Applications.Core/secretStores@2023-10-01-preview' = {
+  name: 'authcreds'
+  properties:{
+    application: application
+    // today the type is an enum, see https://github.com/radius-project/radius/pull/7816
```
rynowak, line 43: Is this something addressed by the proposal?

resources/2024-07-secretstore-feature-spec.md
sylvainsf, line 24: Right now it reads like it could be two things: a recipe containing a single instance of a secret used by all applications (simple), just want to make sure we don't paint ourselves into a corner if there's any plans for a recipe containing a class of secret that resolves to an application instance specific secret(complex).

resources/2024-07-secretstore-feature-spec.md
```diff
@@ -0,0 +1,417 @@
...
+```
+
+Referencing an existing `secretStore` resource:
+
+```bicep
+resource azurekeyvaultsecrets 'Applications.Core/secretStores@2023-10-01-preview' = {
+  name: 'existing-azurekeyvault-secrets'
+  properties:{
+    application: application
+    resource: 'secret-app-existing-secret'
```
rynowak, line 70: I'm confused by the difference between defining and referencing. It looks like in both cases we're setting the secret values.

resources/2024-07-secretstore-feature-spec.md
```diff
@@ -0,0 +1,417 @@
...
+      }
+      'alias': {
+        value: secretalias
+      }
+    }
+  }
+}
+```
+
+Step 2: Operator references the `Applications.Core/secretStores` resource in their Recipe's `output` object to securely pass secrets as outputs to Radius that can be used in provisioning and deploying the resource using the Recipe.
```
rynowak, line 91: How will a user do this from Terraform?

resources/2024-07-secretstore-feature-spec.md
kachawla: It might be helpful to use a more real example instead of a string like `secret-app-existing-secret`. For example, if a user wanted to reference an azure keyvault, would they need to provide a fully qualified resource id?

resources/2024-07-secretstore-feature-spec.md
rynowak: Users can create a secret from Bicep, how can they do it from Terraform?

resources/2024-07-secretstore-feature-spec.md
```diff
@@ -0,0 +1,417 @@
...
+  properties: {
+    application: application
+    container: {
+      image: 'ghcr.io/radius-project/samples/demo:latest'
+      env:{
++        DB_CONNECTION: {
++          valueFrom: {
++            secretRef: {
++              source: authcreds.id
++              key: 'username'
```
rynowak, line 177: Where does the user get `authcreds.id`?

resources/2024-07-secretstore-feature-spec.md
```diff
@@ -0,0 +1,417 @@
...
+        web: {
+          containerPort: 3000
+        }
+      }
+    }
+  }
+}
+```
+
+Step 7: Developer deploys the resources to Radius and the secrets required are either injected into the container as environment variables or used as credentials for authentication into the extender, volume, datastore, or messaging resource at deploy time. If the resource is deployed using a Recipe, the secrets are securely passed as outputs to Radius by the Recipe for use in provisioning and deploying the resource.
```
rynowak, line 359: I'm really confused about this walkthrough because the steps are numbered 1-7, but they appear unrelated, or to address different usecases.

resources/2024-07-secretstore-feature-spec.md
```diff
@@ -0,0 +1,417 @@
...
+
+## Key investments
+<!-- List the features required to enable this scenario. -->
+
+### Feature 1: Add functionality to reference `Applications.Core/secretStores` in `Applications.Core/containers` resources
+<!-- One or two sentence summary -->
+Add the ability for developers to reference values from their `Applications.Core/secretStores` resources in their `Applications.Core/containers` resource definitions under the `properties.container.env` field so that secrets can be injected as environment variables into their container at deploy time.
```
rynowak, line 366: Did we do this already?

resources/2024-07-secretstore-feature-spec.md
sylvainsf: Further thought I think the complex case isn't one we're trying to solve.

resources/2024-07-secretstore-feature-spec.md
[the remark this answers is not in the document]
rynowak: I'd bucket these based on the user's goal/task. I think there are three scenarios:

- Application code needs to read secrets as env-vars.
- Application code needs to read secrets as files.
- Radius needs to operate on secrets for infrastructural purposes (unrelated to user-code). 

What's good about this framing is that the user has already decided what they need to accomplish. We can talk clearly about needs and wants without getting stuck in implementation.

resources/2024-07-secretstore-feature-spec.md
```diff
@@ -0,0 +1,417 @@
...
+> We should revisit the overall plan for provisioning types we plan to support on portable resources before implementing this change. There are some changes with UDT that haven't been solidified yet.
+
+### Feature 4: Add functionality to reference `Applications.Core/secretStores` objects as a parameter to a Bicep resource
+<!-- One or two sentence summary -->
+Add the ability for developers to specify the `Applications.Core/secretStores` resource as a parameter of the `object` type to the Bicep resource that requires the secret so that secrets can be passed as parameters into a Radius Bicep resource.
+
+### Feature 5: Add functionality to reference `Applications.Core/secretStores` in Recipes
+<!-- One or two sentence summary -->
+Add the ability for operators to reference values from their `Applications.Core/secretStores` resources in their Recipe's `output` object so that secrets can be securely passed as outputs to Radius that can be used in provisioning and deploying the resource using the Recipe.
+
```
rynowak, line 385: It feels like this design doc is trying to accomplish a lot. Which of these should we do first?

resources/2024-07-secretstore-feature-spec.md
```diff
@@ -0,0 +1,417 @@
...
+  values: {
+    host: redis.properties.hostName
+    port: redis.properties.port
+    username: ''
+  }
+  secrets: {
++    password: {
++      valueFrom: {
++        secretRef: {
++          source: authcreds.id
```
kachawla, line 128: I missed this before. I don't believe bicep would allow you to reference a resource that is not defined within this file.

resources/2024-07-secretstore-feature-spec.md
```diff
@@ -0,0 +1,417 @@
...
+  value = {
+    values = {
+      host = memory_db.cluster_endpoint_address
+      port = memory_db.cluster_endpoint_port
+    }
+    secrets = {
++      password = {
++        valueFrom = {
++          secretRef = {
++            source = authcreds.id
```
kachawla, line 151: same here

resources/2024-07-secretstore-feature-spec.md
```diff
@@ -0,0 +1,417 @@
...
+<!-- What big questions must we answer in order to clarify our plan for this scenario.  When and how do you plan to answer those questions (prototype feature x, user research, competitive research, etc) -->
+
+**Assumption: operators can create and deploy `Applications.Core/secretStores` resources.** We assume that operators create and manage secrets on behalf of developers, which are encapsulated in `Applications.Core/secretStores` resources. The developers can subsequently reference these resources in their application resources to securely manage secrets for use in their applications. We will validate this assumption by opening up this feature spec for discussion with the community.
+
+**Assumption: the `Applications.secretStores` resource type can be referenced as a parameter to a Bicep resource.** We assume that the `Applications.secretStores` resource type can be referenced as a parameter to a Bicep resource so that secrets can be passed as object parameters into a Radius Bicep resource. We will validate this assumption by investigating the feasibility of this approach during tech design.
```
kachawla, line 404: Ah this assumption answers my confusion above. I don't think this assumption is technically feasible.

resources/2024-07-secretstore-feature-spec.md
willtsai: removing this as it was deemed out of scope for this feature specification

resources/2024-07-secretstore-feature-spec.md
willtsai: good point, i've updated the example to reference that it's all within a single `app.bicep` file

resources/2024-07-secretstore-feature-spec.md
willtsai: removing this as it was deemed out of scope for this feature specification

resources/2024-07-secretstore-feature-spec.md
willtsai: They get it from Step 1 after they create the secret

resources/2024-07-secretstore-feature-spec.md
willtsai: updated steps structuring to avoid this confusion

resources/2024-07-secretstore-feature-spec.md
willtsai: yes, removing

resources/2024-07-secretstore-feature-spec.md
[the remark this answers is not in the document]
willtsai: good suggestion, modified

resources/2024-07-secretstore-feature-spec.md
willtsai: shrunk down the scope to 3 core features (inject as env vars in containers, mounted on volume, use by Radius for auth)

resources/2024-07-secretstore-feature-spec.md
willtsai: removed this as out of scope for this feature spec

resources/2024-07-secretstore-feature-spec.md
```diff
@@ -0,0 +1,329 @@
+# Extend the use-cases of `Applications.Core/secretStores`
+
+* **Author**: Will Tsai (@willtsai)
+
+## Topic Summary
+<!-- A paragraph or two to summarize the topic area. Just define it in summary form so we all know what it is. -->
+
+This feature spec outlines the desired user experience for extending the use-cases of `Applications.Core/secretStores` to allow for referencing secret stores in `Applications.Core/containers`, `Applications.Core/extenders`, `Applications.Core/volumes`, `Applications.Datastores/*`, and `Applications.Messaging/*` resources. The goal is to enable developers to securely manage secrets for use in their applications by referencing `Applications.Core/secretStores` in their resources.
```
ytimocin, line 8: Would we be able to reference them also in Dapr and other UDTs?

nithyatsu (commented):
not sure if I missed the conversation earlier, but with this feature, would there be changes to Credentials type that we use today for storing AZ and AWS secrets? If so, Then does secret store have to move into UCP's resource type?

resources/2024-07-secretstore-feature-spec.md
```diff
@@ -0,0 +1,329 @@
...
+        web: {
+          containerPort: 3000
+        }
+      }
+    }
+  }
+}
+```
+
+(b) Developer references the `Applications.Core/secretStores` resource in their `Applications.Core/volumes` resource definition within the `app.bicep` application definition to write the secret into a file on disk and mount that volume into their container at deploy time.
```
sylvainsf, line 169: I suspect we will also need to define filename (or infer from the key) and path. Often secrets are mounted in a relative way such as a mount being /etc/nginx/conf and then relative paths for certs.

resources/2024-07-secretstore-feature-spec.md
willtsai: good point. added some dapr resource specific features.

resources/2024-07-secretstore-feature-spec.md
willtsai: Addressed this and modified how the volumes use case would work.

willtsai:
> not sure if I missed the conversation earlier, but with this feature, would there be changes to Credentials type that we use today for storing AZ and AWS secrets? If so, Then does secret store have to move into UCP's resource type?

This feature spec should not affect how we store AZ and AWS secrets, thanks for checking though!

willtsai:
still waiting on another round of reviews

brooke-hamilton (approved):

resources/2024-07-secretstore-feature-spec.md
```diff
@@ -0,0 +1,426 @@
...
+
+[proposed feature] As a developer, I can reference an `Applications.Core/secretStores` in my `Applications.Datastores/*`, `Applications.Messaging/*`, or `Applications.Extenders` resource definition so that Radius will use the password internally to authenticate into the resource at deploy time. I no longer have to store secrets as plain text in the `properties.secrets` field of my resource for authentication, etc.
+
+[proposed feature] As a developer, I can reference an `Applications.Core/secretStores` in my `Applications.Dapr/*` resource definition so that Radius will use the Radius secret store to manage secrets for use in my Dapr components. I no longer have to create a Dapr secret store and manage the secrets in the Dapr secret store to be able to reference it in my `Applications.Dapr/*` resource.
+
+[future feature] As a developer, I can specify a `Applications.Core/secretStores` in my custom user-defined resource once the [user-defined types](https://github.com/radius-project/design-notes/blob/main/architecture/2024-07-user-defined-types.md) feature is implemented.
```
kachawla, line 114: Should we create feature requests in github issues/on the roadmap just to track them, might also help get community inputs if someone tries to look up these features?

resources/2024-07-secretstore-feature-spec.md
```diff
@@ -0,0 +1,426 @@
...
++           }
++         }
+        }
+    }
+  }
+}
+```
+> Note: the`secrets`, `filePath`, `fileName` properties and default values are just proposals and may be changed during tech design. They should also be optional for the `volumes` property in `Applications.Core/containers` to write the secret into a file on disk/memory with a specified file path and name, with the proposed defaults being `/var/run/secrets/` and `secretfile.txt` respectively.
+
+(c) Developer references the `Applications.Core/secretStores` resource in their `Applications.Extenders`, `Applications.Datastores/*`, `Applications.Messaging/*`, or `Applications.Dapr/*` resource definition to securely manage secrets for use in their application. Radius then uses the secret to authenticate into the resource at deploy time. The secrets might be referenced in the resources within the `app.bicep` application definition as follows:
```
kachawla, line 235: > Radius then uses the secret to authenticate into the resource at deploy time.

Apologies if this was already discussed during discussion - I didn't fully understand what do we mean by Radius authenticating into the resource?

resources/2024-07-secretstore-feature-spec.md
```diff
@@ -0,0 +1,426 @@
...
+    secrets: {
++        password: {
++          valueFrom: {
++            secretRef: {
++              source: authcreds.id
++              key: 'password'
++            }
++          }
++        }
+    }
```
kachawla, line 256: I understand the scenarios below for manual resource provisioning.. but usage of secrets with a recipe isn't clear. In case of recipes based deployments, secrets property is auto populated by Radius currently from the outputs returned by the recipe. So this input from the user will be overwritten.

resources/2024-07-secretstore-feature-spec.md
```diff
@@ -0,0 +1,426 @@
...
+    env:{
++        URI: {
++          valueFrom: {
++            secretRef: {
++              source: authcreds.id
++              key: 'uri'
++            }
++          }
++        }
+      }
```
kachawla, line 371: We generally expect env variables on container. How/where do envision this being used?

kachawla (commented):
Thanks for the updates - looks great! Left a few clarifying comments.

kachawla (approved):

PR #61, opened 2024-08-17
Add Resource Type Registration doc

rynowak:


architecture/2024-08-resource-types-registration.md
```diff
@@ -0,0 +1,555 @@
...
+
+##### Bootstrap registration
+
+UCP provides a *bootstrap* implementation that registers planes instances during startup using its configuration file. This will be used to register the resource providers and resource types that Radius includes by default. The *bootstrap* implementation will be enhanced as needed to implement this functionality.
+
+### Error Handling
+
+The important error handling scenarios for this change are related to the nested resource lifecyle.
+
+- Rejecting an operation of a resource type when the parent resource provider resource is processing an operation.
```
rynowak, line 493: I'm not sure if this is needed and I've debated taking it out of the doc. Let's discuss.

architecture/2024-08-resource-types-registration.md
```diff
@@ -0,0 +1,580 @@
...
+<center>
+
+![architecture diagram](./2024-08-user-defined-types-registration/registration.png)
+
+**Fig: Diagram of resource type registration and querying** [link](https://excalidraw.com/#json=FzQQaxv1d-vlmmDBAeOfc,UUXIasjO8p27hN0AB8qeiw)
+
+</center>
+
+### Detailed Design
+
```
rynowak, line 140: Something I realize as I'm rereading ... I should summarize the current state - how things work today. I expect not everyone is familiar with this 😆

architecture/2024-08-resource-types-registration.md
```diff
@@ -0,0 +1,580 @@
...
+| Resource Provider  | A resource namespace and implementation of the resource namespace. Resource providers are a new resource type introduced by this proposal.                                                                                                                                                   |
+| Resource Type      | An API entity with CRUDL operations. Resource Types are part of a resource provider. Resource types are a new resource type introduced by this proposal.                                                                                                                                     |
+| OpenAPI/Schema     | A format for documenting HTTP-based APIs. We use OpenAPI v2 currently in Radius for client-sdk generation and validation of inbound requests. OpenAPI v3 is more widely adopted. In this document we're primarily concerned with the parts of OpenAPI pertaining to request/response bodies. |
```
nicolejms, line 24: How much work would it be to move to OpenAPI v3? what benefits would we/users garner from it?

architecture/2024-08-resource-types-registration.md
```diff
@@ -0,0 +1,580 @@
...
+
+She can use tools like Bicep, GitOps, or the `rad` CLI to register the new resource types in any tenant of Radius.
+
+*Note: the exact set of tooling implementations are out of scope for this document.*
+
+#### Removing default resource types
+
+Suzy is a platform engineer at Contoso working on the internal developer platform. She wants to avoid confusion about the set of resource types supported by the platform.
+
+The application teams are Contoso are allowed to use `Applications.Datastores/redisCaches` and `Applications.Datastores/postgreSQLDatabases`. Suzy uses the `rad` CLI to unregister other types in the `Applications.Datastores` namespace.
```
nicolejms, line 56: Would we allow users to unregister the core Radius built in types?

architecture/2024-08-resource-types-registration.md
```diff
@@ -0,0 +1,580 @@
...
+| ------------------ | -------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| Resource Namespace | A group of related or similar resources that share a resource provider. eg: In `Applications.Core/containers` the resource namespace is `Applications.Core`.                                                                                                                                 |
+| Resource Provider  | A resource namespace and implementation of the resource namespace. Resource providers are a new resource type introduced by this proposal.                                                                                                                                                   |
```
ytimocin, line 22: What is an example of this? Like the whole process that runs to serve to the requests for this resource namespace?

architecture/2024-08-resource-types-registration.md
```diff
@@ -0,0 +1,580 @@
...
+- The `resourceTypes` body will be be large due to the inclusion of an OpenAPI schema. 
+- Users need granular registration and unregistration of individual resource types. (See scenario 2 above).
+
+---
+
+These new resource types are scoped to the Radius plane. 
+
+> eg: `/planes/radius/local/providers/System.Resources/resourceProviders`
+
+The rationale for this is to support registration/unregistration per-tenant of Radius rather than per-installation.
```
nicolejms, line 90: Would users want more granular control? Was this vetted with our partners yet?

architecture/2024-08-resource-types-registration.md
kachawla, line 56: nit: I think you meant *not* allowed to use?

```suggestion
The application teams at Contoso aren't allowed to use `Applications.Datastores/redisCaches` and `Applications.Datastores/postgreSQLDatabases`. Suzy uses the `rad` CLI to unregister other types in the `Applications.Datastores` namespace.
```

ytimocin, line 56: Are they going to be able to remove the ones that come out-of-box with the Radius installation?

architecture/2024-08-resource-types-registration.md
```diff
@@ -0,0 +1,580 @@
...
+
+#### Defining new types
+
+Suzy is a platform engineer at Contoso working on the internal developer platform. She defines a resource type for an internal service `ContosoBus` named `Contoso.Platform/contosoBuses`. 
+
+She can use tools like Bicep, GitOps, or the `rad` CLI to register the new resource types in any tenant of Radius.
+
+*Note: the exact set of tooling implementations are out of scope for this document.*
+
+#### Removing default resource types
```
lakshmimsft, line 52: Is a more probabilistic use case to unregister custom resource types than default types?

architecture/2024-08-resource-types-registration.md
```diff
@@ -0,0 +1,580 @@
...
+
+These APIs can also be implemented consistently across all planes in the future. For example Azure doesn't support user-defined types, but it does support looking up the set of resource types.
+
+Lastly, there will still be some system-defined types. For example `System.Resources/resourceProviders` is a system defined type. The query API can return these, but the `System.Resources/resourceProviders` cannot. Imagine for a minute what would happen if users could remove the capability to specify a type.
+
+**Open Question:** Is *query* the right term for this? The term will show up in docs and in the TypeSpec + generated client + our code. Let's pick the term that's least confusing to us. I'm open to suggestions but I'll just roll with *query* unless there's a better idea.
```
brooke-hamilton, line 105: Query works if we provide some filter/parameter capability. Otherwise, *list* also works if we are providing a simple list capability.

architecture/2024-08-resource-types-registration.md
ytimocin, line 90: So, `rad install kubernetes --reinstall` will not bring back the unregistered resource types.

architecture/2024-08-resource-types-registration.md
kachawla: I'd lean towards *list* as well unless we are doing more than listing

architecture/2024-08-resource-types-registration.md
```diff
@@ -0,0 +1,580 @@
...
+The proposal also defines two new **non-resource** APIs:
+
+- `GET /planes/radius/{plane}/providers/`: list all resource providers.
+- `GET /planes/radius/{plane}/providers/{provider}` shows a specific resource provider.
+
+These APIs are optimized for queries and are consistent with similar APIs in ARM. To understand why this is important, remember that the OpenAPI schema for a resource type can be quite large. Some functionality in Radius will need to query resource types frequently, consider the need for the `rad recipe register` to validate a resource type. Building a separate API for queries allows us to optimize the behavior.
+
+These APIs can also be implemented consistently across all planes in the future. For example Azure doesn't support user-defined types, but it does support looking up the set of resource types.
+
+Lastly, there will still be some system-defined types. For example `System.Resources/resourceProviders` is a system defined type. The query API can return these, but the `System.Resources/resourceProviders` cannot. Imagine for a minute what would happen if users could remove the capability to specify a type.
```
ytimocin, line 103: > The query API can return these, but the `System.Resources/resourceProviders` cannot

The latter can still return these but trying to unregister these could throw an exception.

architecture/2024-08-resource-types-registration.md
ytimocin, line 105: The name makes sense to me, but the user will be able to do **Command** and **Query**, right? By **Command** I mean adding a new Resource Type. Why not call it Resource Type API or Custom Resource Type API?

architecture/2024-08-resource-types-registration.md
lakshmimsft, line 103: I missed understanding this sentence.: For example `System.Resources/resourceProviders` is a system defined type. The query API can return these, but the `System.Resources/resourceProviders` cannot.
I thought we can query system-defined types too.

architecture/2024-08-resource-types-registration.md
```diff
@@ -0,0 +1,580 @@
...
+
+**Open Question:** Is *query* the right term for this? The term will show up in docs and in the TypeSpec + generated client + our code. Let's pick the term that's least confusing to us. I'm open to suggestions but I'll just roll with *query* unless there's a better idea.
+
+#### Locations
+
+UCP is a multi-location control-plane like the control-plane that inspired it (ARM). We don't use UCP/Radius in any multi-location scenarios right now but it's important we don't design APIs that lock us into a single-region design. ARM has a mature multi-location design but doesn't expose to users the ability to register types. However we can learn a lot from the scenarios supported by ARM for rolling out new API versions and new types.
```
brooke-hamilton, line 109: In addition to location, ARM allows resources/providers to be registered to specific subscriptions. Is there an analogy for that in Radius? For example, would a user context be applied in which specific resource types/providers are only available to a specific user? 

A valid scenario for this would be white glove onboarding of a specific team to a new resource type, or a controlled rollout of a new version of a type in which the operator wants to control who has access to the type and provider.

ytimocin, line 109: Can we talk about what `multi-location` means? Are we talking about actual cloud locations/regions? Or are we talking something like multi-cluster? And why are the resource providers deployed in a single location? If we have a multi-location UCP and register a resource type using UCP, why is that resource type not available to all the locations? An example could be nice here.

ytimocin: And you gave the example below 🤣

ytimocin: But why are we limiting the resource types to a single location?

ytimocin: I would also like to learn more about UCP being multi-location.

architecture/2024-08-resource-types-registration.md
```diff
@@ -0,0 +1,580 @@
...
+
+Some examples:
+
+> Resource provider `Contoso.Example` is only available in the `westus-1` location. UCP must reject creation requests to other locations.
+
+> Resource type `Contoso.Example/widgets` is available in all locations, but new API version `2024-08-01-preview` is only in the `eastus-2` location. UCP should block all requests for unsupported API version x location pairs.
+
+The design proposal includes axes for configuring location affinity:
+
+- `System.Resources/resourceProviders` includes a list of supported locations with an optional `address` in each location for the resource provider.
```
ytimocin, line 122: Are the available locations going to be an array for a resource type?

architecture/2024-08-resource-types-registration.md
```diff
@@ -0,0 +1,580 @@
...
+### Detailed Design
+
+Since this is our first nested resource type, a basic primer on nested types is needed. A nested resource type exists *inside* another resource type. A resource type is that is not nested is *top-level*. These relationships are also sometimes referred to as a parent and child. 
+
+The following statements are all accurate:
+
+- The `System.Resources/resourceProviders` type is the parent of `System.Resources/resourceProviders/resourceTypes`. 
+- The `System.Resources/resourceProviders/resourceTypes` type is the child of `System.Resources/resourceProviders`.
+- The `System.Resources/resourceProviders/resourceTypes` type is nested within `System.Resources/resourceProviders`.
+- The `System.Resources/resourceProviders` type is a top-level resource type.
```
ytimocin, line 148: We can't have a resource type being a parent of another resource type, right? Like going deeper in different types of **redis** for example.

architecture/2024-08-resource-types-registration.md
```diff
@@ -0,0 +1,580 @@
...
+Nesting a resource type also nests the lifecyle of the resource. Using our examples above, the `/planes/radius/{plane}/providers/System.Resources/resourceProviders/resourceTypes/Contoso.Example/contosoBuses` type cannot exist unless `/planes/radius/{plane}/providers/System.Resources/resourceProviders/Contoso.Example` already exists.
+
+There are a few simple rules for this:
+
+- For operations on parent resources...
+  - A `DELETE` operation on a parent resource deletes all children.
+
+- For operations on child resources...
+  - A *CRUL* operation (anything but delete) on a child resource must return a `404` when the parent does not exist.
+  - A `DELETE` operation on a child resource never returns a `404`.
```
ytimocin, line 168: Unless someone has already deleted that maybe?

architecture/2024-08-resource-types-registration.md
rynowak: This is probably most relevant to one of the upcoming docs, but I'll answer anyway 👍. 

It would be a small amount of work for us to move because we're using TypeSpec to abstract this away and to generate the OpenAPIv2 doc. Generating OpenAPIv3 would be a different command line parameter and possibly some additional annotations in TypeSpec. We could possibly generate both and use them for different purposes. The Azure/ARM toolchain for generating clients and validating RP APIs is based on v2.

The benefit for us is that we're not maintaining an old thing that the world has moved on from. The benefit for users is that they are much more likely to be familiar with v3. v3 has a bunch of minor design improvement compared to v2.

architecture/2024-08-resource-types-registration.md
rynowak: I think they are both pretty important. I went with the "unregister built-in stuff" as a scenario because it's less obvious. 

We've already validated with users that this is important to them.

architecture/2024-08-resource-types-registration.md
rynowak: Yes, definitely. We've already gotten this feedback.

rynowak: Yes, thank you 👍

architecture/2024-08-resource-types-registration.md
rynowak: Let's discuss in the meeting and I'll post conclusions here.

architecture/2024-08-resource-types-registration.md
```diff
@@ -0,0 +1,580 @@
...
+
+Resource providers:
+
+- Locations: the set of logical locations where the resource provider is supported. Each location has:
+  - An optional address/URL of the resource provider. This is optional because it won't be needed for declarative user-defined types.
+
+
+Resource types:
+
+- Locations: the set of logical locations where the resource type is supported.
```
ytimocin, line 184: I can understand this if we are using cloud resources but if we are not, how do we use the locations for UCP and/or resource providers and/or resource types?

architecture/2024-08-resource-types-registration.md
```diff
@@ -0,0 +1,580 @@
...
+- The `System.Resources/resourceProviders` type is a top-level resource type.
+
+Any resource type with two segments (separated by `/`) is a top-level resource type. Any resource type with 3 or more segments (separate by `/`) is a nested resource type.
+
+The resource ids (and URLs) are written like the following:
+
+- `/planes/radius/{plane}/providers/System.Resources/resourceProviders` - List all resource providers.
+- `/planes/radius/{plane}/providers/System.Resources/resourceProviders/Contoso.Example` - Get the Contoso.Example resource provider.
+- `/planes/radius/{plane}/providers/System.Resources/resourceProviders/Contoso.Example/resourceTypes` - List all resource types inside the `Contoso.Example` resource provider/namespace.
+- `/planes/radius/{plane}/providers/System.Resources/resourceProviders/Contoso.Example/resourceTypes/contosoBuses` - Gets the `contosoBuses` resource type inside the `Contoso.Example` resource provider/namespace.
```
kachawla, line 157: I think you have explained it above, but I'm struggling a bit to understand the need for `System.Resources/resourceProviders/<name>/resourceTypes` in the overall id. Would be great to discuss this..

architecture/2024-08-resource-types-registration.md
rynowak: We've got strong validation on the idea of registering/unregistering at the plane scope. 

The idea of registering/unregistering at the resource group scope is pretty complex. It would be hard for us to get feedback on this right now. Just explaining the idea to most users would be complex.

architecture/2024-08-resource-types-registration.md
rynowak: Yes, the query API can return these. The `System.Resources/resourceProviders` API cannot, or at least it can't be modified through `System.Resources/resourceProviders`.

architecture/2024-08-resource-types-registration.md
rynowak: > If this capability is separate from the CRUDL feature mentioned above, then the term query makes sense.

It's a separate API from `System.Resources/resourceProviders`. This is why it feels like it needs a distinct term. 

Let's make sure to cover this in the next design discussion. I'm fairly confident that we need a separate API for "queries", but I think we need to go through the discussion and see if the rest of the room is convinced.

architecture/2024-08-resource-types-registration.md
rynowak: Great questions.

This is covered in the alternatives section below. My opinion  is that I *hope* this complexity isn't needed.

- Azure doesn't have the ability to define your own resource providers or types, so the registration functionality is needed.
- Radius/UCP does have the ability to define your own resource types so the registration functionality would be redundant. 

---

> A valid scenario for this would be white glove onboarding of a specific team to a new resource type

In the Azure context is this is a combination of the registration feature + feature flagging. 

---

The registration feature provides a "callback" to the resource provider so it can initialize itself (mostly identities) and do checks it needs to do. 


If we need a "callback" in UCP/Radius in the future we could add it, and we'd tie to to resource provider creation rather than registration.

architecture/2024-08-resource-types-registration.md
rynowak: An example of this would be `Applications.Core` + `applications-rp` in Radius today. It's both the set of supported types and the implementation.

architecture/2024-08-resource-types-registration.md
rynowak: Yes definitely. See scenario 2

architecture/2024-08-resource-types-registration.md
rynowak: Yes exactly.

architecture/2024-08-resource-types-registration.md
rynowak: That's another option we could consider. Modeling them as `System.Resources/resourceProviders` instances, but making them read-only.

architecture/2024-08-resource-types-registration.md
rynowak: Both of these terms are pretty ambiguous IMO. 

- Resource provider query API - what I'm proposing. 
- Resource Type API - doesn't feel distinct from `System.Resources/resourceProviders/resourceTypes`.
- Custom Resource Type - I don't think the word custom helps us tell these two APIs apart.

architecture/2024-08-resource-types-registration.md
rynowak: The concept of `location` in UCP/ARM answers the question:

> What instance of the control-plane is responsible for this resource?

Remember that `location` does not constrain what users can do. Users **can** mix and match the location of the control plane with the location of the resources they create. 

> Or are we talking something like multi-cluster?

Not exactly. Each location is a "logical instance" of Radius that's working together with others. Like a player on a soccer field, they move independently but they are all playing the same game.

For naive users they will use a single location for everything 👍.

---

`location` provides a lot of value when you start to optimize around it. Being multi-location allows users to deploy UCP/Radius next to the resources they want it to manage. This limits the amount of cross-zonal network traffic required and removes single points of failure.

Imagine you deployed Radius in `westus3`, but you're deploying applications in `westus3`, `canadaeast` and `westeurope`. You *could* do this and it would work.  However if `westus3` has an outage then you can't manage **any** of your applications. 

Now imagine you create a multi-location deployment of Radius. You create a location for each logical location where you want to run applications (`westus3`, `canadaeast` and `westeurope`). Now if `westus3` has an outage you can still manage applications in `canadaeast` and `westeurope`.

architecture/2024-08-resource-types-registration.md
rynowak: Please see the API design.

architecture/2024-08-resource-types-registration.md
rynowak: That's a great question and something we need to support. Let me think about how we're going to represent that.

architecture/2024-08-resource-types-registration.md
rynowak: Discussed in the meeting. My action item is to provide more examples. 

Answer to this question:

`/planes/radius/{plane}/providers/System.Resources/resourceProviders/Contoso.Example/resourceTypes/contosoBuses` represents the **type definition** for `Contoso.Example/contosoBuses`.

The resource IDs and paths of `Contoso.Example/contosoBuses` will look like our other resources: eg. `/planes/radius/{plane}/resourceGroups/foo/providers/Contoso.Example/contosoBuses/my-resource`

architecture/2024-08-resource-types-registration.md
rynowak: `DELETE `operations in UCP never return a 404. `DELETE` is the "make it gone" button.

architecture/2024-08-resource-types-registration.md
rynowak: Discussed in the meeting. My action item is to make a more concrete description of what location **means**.

Short answer for now: location determines which instance of the control-plane is responsible. It does not constrain what users can do. 

eg: a `westus3` instance of Radius can manage Azure resources in any Azure location.

architecture/2024-08-resource-types-registration.md
ytimocin: Interesting. So, if `application-rp` is deleted, we are not going reinstall that pod. The `reinstall` will need to know that the types it needs to install were deleted before. We can have a key called `deletedByUser` or something like that and if that is true, **reinstall** will not install that.

rynowak: I was talking about the `System.Resources/resourceProviders/resourceTypes` resources. We don't want users removing the `applications-rp` microservice. 

I think it's totally valid for users to remove the type for `Applications.Datastores/redisCaches`.

architecture/2024-08-resource-types-registration.md
rynowak: I changed this to `resource provider summary`. Now that the other APIs are more complex, it makes sense to call it a summary. It's not **all** of the info, just the most commonly-used parts.

architecture/2024-08-resource-types-registration.md
rynowak: I added more definitions and information.

rynowak:
Ready for another review.

rynowak:
Still waitng for a chance to have this reviewed.... That'll teach me to miss monday/tuesday for three weeks in a row! 😆

architecture/2024-08-resource-types-registration.md
```diff
@@ -0,0 +1,641 @@
...
+
+| Resource Type      | An API entity that provides CRUDL operations. Resource Types are implemented by a resource provider and belong to a resource provider namespace. Resource types are a new resource type introduced by this proposal. This proposal introduces an API for registering and unregistering resource resource types.                                                                                                                                     |
+| OpenAPI/Schema     | A format for documenting HTTP-based APIs. In this document we're primarily concerned with the parts of OpenAPI pertaining to request/response bodies. |
+| Location/Region | An instance of a loosely-connected deployment of UCP. Regions of a UCP can collaborate to serve route requests between each other. A multi-regional deployment of UCP provides fault-isolation. In a multi-regional deployment of UCP, each region is responsible for processing requests that correspond to its region based on the `location` field. |
+
```
rynowak, line 27: Looks like I broke the formatting :)

architecture/2024-08-resource-types-registration.md
```diff
@@ -0,0 +1,641 @@
...
+* **Author**: `Ryan Nowak` (`@rynowak`)
+
+## Overview
+
+*This document describes the resource type registration APIs. This is a detailed design doc focused on one aspect of the overall user-defined types feature area. Please see the overall user-defined-types plan for more context.*
+
+Resources are the unit of exchange in the Radius API. All of our user-experiences like the CLI, Dashboard, and Bicep authoring are oriented around the catalog of resources. Because Radius is strongly-typed, the set of resource types and their schemas are also defined as a primary concept in Radius. In order for Radius to support each additional kind of resources, for example a PostgreSQL database, we must first have a type definition.
+
+The registration APIS for resource types determine how users add, change, or remove types from Radius. The APIs defined in this document define the *cabilities* of the Radius/UCP implementation with respect to registration of reosurce providers, resource types, and locations. The vast majority of Radius users will not interact with these APIs in detail.
```
ytimocin, line 11:
```suggestion
The registration APIS for resource types determine how users add, change, or remove types from Radius. The APIs defined in this document define the *capabilities* of the Radius/UCP implementation with respect to registration of resource providers, resource types, and locations. The vast majority of Radius users will not interact with these APIs in detail.
```

architecture/2024-08-resource-types-registration.md
```diff
@@ -0,0 +1,641 @@
...
+
+## Objectives
+
+> **Issue Reference:** https://github.com/radius-project/radius/issues/6688
+
+### Goals
+
+- Support granular registration of resource providers and resource types (add/change/remove) by users.
+- Support efficient querying of resource type data for any API consumer.
+- Register resource types for existing Radius functionality like Deployment Engine and `Applications.*` resource types.
```
ytimocin, line 36: Can you give an example of adding a resource type to a Deployment Engine? I understand that users can add to **Applications.Core**, for example, but how can they do the same for Deployment Engine?

architecture/2024-08-resource-types-registration.md
```diff
@@ -0,0 +1,641 @@
...
+- Support the future of UCP as a multi-regional control-plane by avoiding breaking changes in the future.
+  - For now we ensure the `location` field and concept is surfaced in the API by assuming that multi-regional deployments are possible. In practice only a single region will be present.
+
+### Non goals
+
+- (out of scope): Complete the user-defined type implementation. This design document is scoped to registration of resource types.
+- (out of scope): Tooling for resource type registration (CLI, Bicep). This work is tracked by another design document.
+- (non-goal): Provider non-user-facing mechanisms for resource type registration.
```
ytimocin, line 44:
```suggestion
- (non-goal): Provide non-user-facing mechanisms for resource type registration.
```

architecture/2024-08-resource-types-registration.md
nicolejms, line 27: The table not rendering correctly

architecture/2024-08-resource-types-registration.md
```diff
@@ -0,0 +1,641 @@
...
+
+For anyone skeptical about this, the concept of resource provider is already present in **every** resource id/URL. This is the meaning of the `providers` segment. Currently URLs like `/planes/radius/local/providers` don't have any functionality associated with them, and return a 404 - this proposal addresses this gap.
+
+One of these lessons-learned is the ability to register multiple regions/locations where a resource provider can operate. In particular users will need the ability to gradually roll out functionality across regions rather than updating all regions at once. This is complex, and doesn't help us in the short term. The approach taken here is aligned with our current strategy for region/location in Radius. The APIs are designed for multi-region but the implementation assumes a single region is used.
```
ytimocin, line 79: Will they still have the option to roll out functionality to all regions it they want? Like passing in **/*** for the location parameter in the create query for example.

architecture/2024-08-resource-types-registration.md
```diff
@@ -0,0 +1,641 @@
...
+
+#### Resource Types
+
+The registration APIs add a few new resource types:
+
+- `System.Resources/resourceProviders` for resource providers.
+- `System.Resources/resourceProviders/resourceTypes` for resource types.
+- `System.Resources/resourceProviders/resourceTypes/apiVersions` for API versions of resource types.
+- `System.Resources/resourceProviders/locations` for the configuration of a resource provider in a specific location/region.
```
ytimocin, line 88: Very nice summary. Thanks! Makes a lot of sense.

architecture/2024-08-resource-types-registration.md
```diff
@@ -0,0 +1,641 @@
...
+
+These new resource types define the full CRUDL set of APIs. These APIs are optimized for operations that need to modify or deeply understand resource types. 
+
+These are our first nested resource type in UCP/Radius. This is a good usecase for a nested resource types because:
+
+- The `apiVersions` body will be be large due to the inclusion of an OpenAPI schema.
```
ytimocin, line 94:
```suggestion
- The `apiVersions` body will be large due to the inclusion of an OpenAPI schema. 
```

architecture/2024-08-resource-types-registration.md
willtsai, line 11:
```suggestion
The registration APIS for resource types determine how users add, change, or remove types from Radius. The APIs defined in this document define the *capabilities* of the Radius/UCP implementation with respect to registration of resource providers, resource types, and locations. The vast majority of Radius users will not interact with these APIs in detail. 
```

architecture/2024-08-resource-types-registration.md
```diff
@@ -0,0 +1,641 @@
...
+- Users need granular registration and unregistration of individual resource types and API versions. (See scenario 2 above).
+- Modeling `location` as a resource optimizes for per-region rollouts and validation of new functionality.
+
+---
+
+These new resource types are scoped to the Radius plane. 
+
+> eg: `/planes/radius/local/providers/System.Resources/resourceProviders`
+
+The rationale for this is to support registration/unregistration per-tenant of Radius rather than per-installation.
```
ytimocin, line 104: At some point, we may probably need to give users the ability to back-up their Radius installation so that they don't lose data.

architecture/2024-08-resource-types-registration.md
```diff
@@ -0,0 +1,641 @@
...
+  - UCP can query a single `locations` resource to check for support and find the downstream API address.
+- Dynamic-RP Validation:
+  - Dynamic-RP knows the resource type and API version.
+  - Dynamic-RP can query a single `apiVersions` resource to fetch its schema for validation. 
+
+#### Resource Type Summaries
+
+The proposal also defines two new **non-resource** APIs:
+
+- `GET /planes/radius/{plane}/providers/`: list the most commonly used fields all resource providers.
```
ytimocin, line 123:
```suggestion
- `GET /planes/radius/{plane}/providers/`: lists the most commonly used fields all resource providers.
```

architecture/2024-08-resource-types-registration.md
```diff
@@ -0,0 +1,641 @@
...
+
+> **Issue Reference:** https://github.com/radius-project/radius/issues/6688
+
+### Goals
+
+- Support granular registration of resource providers and resource types (add/change/remove) by users.
+- Support efficient querying of resource type data for any API consumer.
+- Register resource types for existing Radius functionality like Deployment Engine and `Applications.*` resource types.
+- Support the future of UCP as a multi-regional control-plane by avoiding breaking changes in the future.
+  - For now we ensure the `location` field and concept is surfaced in the API by assuming that multi-regional deployments are possible. In practice only a single region will be present.
```
willtsai, line 38: it defaults to `global` today, correct?

architecture/2024-08-resource-types-registration.md
ytimocin: How will the system decide on what the most commonly used fields are? Are we talking about returning all the fields of all resource providers? What are the most commonly used fields?

architecture/2024-08-resource-types-registration.md
zachcasper, line 88: I think about location as an attribute of an environment rather than a resource provider. By tying a location to a resource provider, are we possibly, we are leaking the implementation? e.g., the location needs a controller/microservice to be running in each location. 

I'm seeing it as:
- resourceProvider as a set of globally scoped objects
- location as a set of globally scoped objects
- environments as an instance of a location and a set of available resource providers within that environment

architecture/2024-08-resource-types-registration.md
```diff
@@ -0,0 +1,641 @@
...
+
+### Non goals
+
+- (out of scope): Complete the user-defined type implementation. This design document is scoped to registration of resource types.
```
willtsai, line 42: Type definitions/instantiations are out of scope here, correct? This doc focuses on registration of a custom resource type that has been assumed to have been defined or instantiated correctly?

architecture/2024-08-resource-types-registration.md
willtsai, line 44: can you elaborate on what this is?

architecture/2024-08-resource-types-registration.md
```diff
@@ -0,0 +1,641 @@
...
+
+Suzy is a platform engineer at Contoso working on the internal developer platform. She defines a resource type for an internal service `ContosoBus` named `Contoso.Platform/contosoBuses`. 
+
+She can use tools like Bicep, GitOps, or the `rad` CLI to register the new resource types in any tenant of Radius.
+
+*Note: the exact set of tooling implementations are out of scope for this document.*
+
+#### Removing default resource types
+
+Suzy is a platform engineer at Contoso working on the internal developer platform. She wants to avoid confusion about the set of resource types supported by the platform.
```
willtsai, line 59: Recommend clarification of how she wants to go about with avoiding confusion, i.e. removing registered types that are not applicable to her users.

architecture/2024-08-resource-types-registration.md
```diff
@@ -0,0 +1,641 @@
...
+Since this is our first nested resource type, a basic primer on nested types is needed. A nested resource type exists *inside* another resource type. A resource type is that is not nested is *top-level*. These relationships are also sometimes referred to as a parent and child. 
+
+The following statements are all accurate:
+
+- The `System.Resources/resourceProviders` type is the parent of `System.Resources/resourceProviders/resourceTypes`. 
+- The `System.Resources/resourceProviders/resourceTypes` type is the child of `System.Resources/resourceProviders`.
+- The `System.Resources/resourceProviders/resourceTypes` type is nested within `System.Resources/resourceProviders`.
+- The `System.Resources/resourceProviders` type is a top-level resource type.
+
+Any resource type with two segments (separated by `/`) is a top-level resource type. Any resource type with 3 or more segments (separate by `/`) is a nested resource type.
```
ytimocin, line 177: The first one (two segments) is always a resource provider, right?

architecture/2024-08-resource-types-registration.md
```diff
@@ -0,0 +1,641 @@
...
+
+##### Structure
+
+The resource ids (and URLs) are written like the following:
+
+- `/planes/radius/{plane}/providers/System.Resources/resourceProviders` - List all resource providers.
+- `/planes/radius/{plane}/providers/System.Resources/resourceProviders/Contoso.Example` - Get the Contoso.Example resource provider.
+- `/planes/radius/{plane}/providers/System.Resources/resourceProviders/Contoso.Example/resourceTypes` - List all resource types inside the `Contoso.Example` resource provider/namespace.
+- `/planes/radius/{plane}/providers/System.Resources/resourceProviders/Contoso.Example/resourceTypes/contosoBuses` - Gets the `contosoBuses` resource type inside the `Contoso.Example` resource provider/namespace.
```
ytimocin, line 186: Can we have another resource type underneath this? Something like: ``/planes/radius/{plane}/providers/System.Resources/resourceProviders/Contoso.Example/resourceTypes/contosoBuses/resourceTypes/contosoBusConfigurations``?

architecture/2024-08-resource-types-registration.md
```diff
@@ -0,0 +1,641 @@
...
+- The `System.Resources/resourceProviders/resourceTypes` type is nested within `System.Resources/resourceProviders`.
+- The `System.Resources/resourceProviders` type is a top-level resource type.
+
+Any resource type with two segments (separated by `/`) is a top-level resource type. Any resource type with 3 or more segments (separate by `/`) is a nested resource type.
+
+##### Structure
+
+The resource ids (and URLs) are written like the following:
+
+- `/planes/radius/{plane}/providers/System.Resources/resourceProviders` - List all resource providers.
```
brooke-hamilton, line 183: Defining the term "plane" would add good context here.

architecture/2024-08-resource-types-registration.md
zachcasper: I just read the locations section below. I think the discussion is how environments fit into the resource provider (RP) and location configuration, if at all. 

If I wanted to control the roll-out of a RP by location, it seems like I would register a new RP, or a new version of an existing RP. Then update the environment by adding the new RP or incrementing the version of an existing RP.

architecture/2024-08-resource-types-registration.md
```diff
@@ -0,0 +1,641 @@
...
+- `/planes/radius/{plane}/providers/System.Resources/resourceProviders/Contoso.Example` - Get the Contoso.Example resource provider.
+- `/planes/radius/{plane}/providers/System.Resources/resourceProviders/Contoso.Example/resourceTypes` - List all resource types inside the `Contoso.Example` resource provider/namespace.
+- `/planes/radius/{plane}/providers/System.Resources/resourceProviders/Contoso.Example/resourceTypes/contosoBuses` - Gets the `contosoBuses` resource type inside the `Contoso.Example` resource provider/namespace.
+
+---
+**Note: These ids refer to the type definition of the `Contoso.Example/contosoBuses` resource type . The ids of `Contoso.Example/contosoBuses` resources are structured like our existing resource types.**
+
+- `/planes/radius/{plane}/resourceGroups/{resourceGroup}/providers/Contoso.Example/contosoBuses` - Lists the `Contoso.Example/contosoBuses` resources inside the `/planes/radius/{plane}/resourceGroups/{resourceGroup}` scope.
```
ytimocin, line 191: Would the provider not belong to the scope?

architecture/2024-08-resource-types-registration.md
```diff
@@ -0,0 +1,641 @@
...
+Nesting a resource type also nests the lifecycle of the resource. Using our examples above, the `/planes/radius/{plane}/providers/System.Resources/resourceProviders/resourceTypes/Contoso.Example/contosoBuses` type cannot exist unless `/planes/radius/{plane}/providers/System.Resources/resourceProviders/Contoso.Example` already exists.
+
+There are a few simple rules for this:
+
+- For operations on parent resources...
+  - A `DELETE` operation on a parent resource deletes all children.
+
+- For operations on child resources...
+  - A *`CRUL`* operation (anything but delete) on a child resource must return a `404` when the parent does not exist.
+  - A `DELETE` operation on a child resource never returns a `404`, because `DELETE` operations never return `404` in our API.
```
ytimocin, line 209: Would it return `Deleted` even if the resource doesn't exist?

architecture/2024-08-resource-types-registration.md
```diff
@@ -0,0 +1,641 @@
...
+@maxLength(63)
+@pattern("^([A-Za-z]([-A-Za-z0-9]*[A-Za-z0-9]))\\.([A-Za-z]([-A-Za-z0-9]*[A-Za-z0-9]))?$")
+scalar ResourceProviderNamespaceString extends string;
+
+@doc("The resource type name")
+@maxLength(63)
+@pattern("^([A-Za-z]([-A-Za-z0-9]*[A-Za-z0-9]))$")
+scalar ResourceTypeNameString extends string;
+
+@doc("The resource type name")
```
ytimocin, line 251:
```suggestion
@doc("The API version name")
```

architecture/2024-08-resource-types-registration.md
```diff
@@ -0,0 +1,641 @@
...
+@maxLength(63)
+@pattern("^([A-Za-z]([-A-Za-z0-9]*[A-Za-z0-9]))$")
+scalar LocationNameString extends string;
+
+#suppress "@azure-tools/typespec-azure-resource-manager/arm-resource-path-segment-invalid-chars"
+model ResourceProviderResource is TrackedResource<ResourceProviderProperties> {
+  @key("resourceProviderName")
+  @doc("The resource provider name.")
+  @path
+  @segment("providers/System.Resources/resourceproviders")
```
ytimocin, line 266:
```suggestion
  @segment("providers/System.Resources/resourceProviders")
```

architecture/2024-08-resource-types-registration.md
```diff
@@ -0,0 +1,641 @@
...
+
+My proposal is that we escape this be replacing `/` with `_`. We do this transformation on resource types elsewhere in code where `/` is not allowed, for example Kubernetes labels. So the result would be:
+
+- `/planes/radius/local/providers/System.Resources/resourceProviders/Contoso.Platform/resourceTypes/contosoBuses_queues`
+
+HTTP defines `%`-encoding to escape significant characters like `%` as part of the URL. So if we follow the HTTP convention this would be:
+
+- `/planes/radius/local/providers/System.Resources/resourceProviders/Contoso.Platform/resourceTypes/contosoBuses%25queues`
+
+I don't like this idea. It's less obvious to users to figure out why `%25` is the right replacement. It's also problematic in implementations. Many HTTP server implementations will decode `%25` -> `/` as part of a defense-in-depth approach to protecting against path-navigation exploits. This behavior varies per-HTTP server and is usually not configurable. It's also unclear whether we'll support other percent-encodings in the future, which we shouldn't do.
```
willtsai, line 554: +1

architecture/2024-08-resource-types-registration.md
```diff
@@ -0,0 +1,641 @@
...
+Some examples:
+
+> Resource provider `Contoso.Example` is only available in the `westus-1` location. UCP must reject creation requests to other locations.
+
+> Resource type `Contoso.Example/widgets` is available in all locations, but new API version `2024-08-01-preview` is only in the `eastus-2` location. UCP should block all requests for unsupported API version x location pairs.
+
+The design proposal includes axes for configuring location affinity:
+
+- `System.Resources/resourceProviders/locations` allows separate configuration of each location with an optional `address` in each location for the resource provider.
+- `System.Resources/resourceProviders/locations` includes the list of supported resource types and supported API versions.
```
lakshmimsft, line 146: clarifying "includes the list of supported resource types and supported API versions".. for each location supported by a resource provider ?

architecture/2024-08-resource-types-registration.md
```diff
@@ -0,0 +1,641 @@
...
+
+- The `apiVersions` body will be be large due to the inclusion of an OpenAPI schema. 
+- Users need granular registration and unregistration of individual resource types and API versions. (See scenario 2 above).
+- Modeling `location` as a resource optimizes for per-region rollouts and validation of new functionality.
+
+---
+
+These new resource types are scoped to the Radius plane. 
+
+> eg: `/planes/radius/local/providers/System.Resources/resourceProviders`
```
Reshrahim, line 102: I have a basic question here. what does `/local/` mean in the context of planes?

architecture/2024-08-resource-types-registration.md
rynowak: This section describes something else. We (Radius contributors) will register resource types for the resource types that already exist, and ship them in the box.

architecture/2024-08-resource-types-registration.md
rynowak: `global` is the sentinel we currently use to mean "there is only a single region". 

This is also a default we set in Bicep. The API doesn't have a default, it's just an API.

architecture/2024-08-resource-types-registration.md
rynowak: Correct. This doc is just about defining/registering the types. Creating resources of those types is one of the next ones 👍

architecture/2024-08-resource-types-registration.md
rynowak: It's covered in the FAQ.

architecture/2024-08-resource-types-registration.md
rynowak: The wording is intentional, and it's consistent with the scenario doc. We take the position in that doc that platform engineers are the curators based on the technologies supported by the platform team.

architecture/2024-08-resource-types-registration.md
rynowak: I think you're on the right track. Exactly how users do that is out of scope for this doc.

architecture/2024-08-resource-types-registration.md
rynowak: Local is the sentinel value we use for "this is a single-tenant installation of Radius".

architecture/2024-08-resource-types-registration.md
rynowak: That sounds cool 👍

architecture/2024-08-resource-types-registration.md
rynowak: We will decide. We're really smart 😉 

This will be driven by scenarios/needs. For example, what does the `rad` CLI need to know to give a good experience? Consider that question times 3-4 other scenarios and you have the idea.

architecture/2024-08-resource-types-registration.md
rynowak: The first segment is the resource provider namespace.

architecture/2024-08-resource-types-registration.md
rynowak: I think this is covered here: https://docs.radapp.io/concepts/technical/api/

architecture/2024-08-resource-types-registration.md
rynowak: I'm proposing we allow 3 levels total. It's a good idea to have some arbitrary limits in place to prevent abuse. ARM supports 3 levels. If we have a **good reason** to allow more in the future we can add it.

architecture/2024-08-resource-types-registration.md
rynowak: What is deleted?

architecture/2024-08-resource-types-registration.md
rynowak: Hat-tip to you fellow HTTP-enjoyer 🎩

architecture/2024-08-resource-types-registration.md
rynowak: The way we're handling OpenAPI unfortunately doesn't support this. I'm not a fan, but it has to be like this. 

Someone could work on that bug if they're motivated 😆

architecture/2024-08-resource-types-registration.md
rynowak: Resolving. If you still have questions please clarify.

rynowak:
Updated doc and addressed feedback. Ready for final review.

kachawla (approved):

PR #68, opened 2024-10-15
Specifications for Dapr Binding Implementation

SoTrx:
Hey,

Here are the proposed specifications for the Dapr Binding implementation.


resources/2024-10-dapr-bindings.md
```diff
@@ -0,0 +1,276 @@
+# Title
```
lakshmimsft, line 1: missing title.

resources/2024-10-dapr-bindings.md
```diff
@@ -0,0 +1,276 @@
...
+
+* **Author**: Lucas Peirone (@SoTrx)
+
+## Overview
+
+Currently, Radius support for Dapr is incomplete. This design document aims to add support for another Dapr Building Block: Bindings.
```
lakshmimsft, line 7: The first sentence may raise more questions rather than help provide an overview for the design. Adding Dapr Bindings will be great but will not complete support for Dapr. Maybe we avoid the first sentence and add more information on how adding Dapr Bindings will ,say, help resources interact with external systems in a decoupled manner.

resources/2024-10-dapr-bindings.md
```diff
@@ -0,0 +1,276 @@
...
+```	
+
+#### Sequence Diagram
+
+( No changes in existing interactions)
+
+### Detailed Design
+
+This design will require adding a new type to the Dapr RP. This includes:
+- Adding Application.Dapr/Bindings to the TypeSpec frontend.
```
lakshmimsft, line 185: 'frontend' doesn't seem right. Suggestion: Extending the TypeSpec schema definition to include the new Application.Dapr/Bindings resource

resources/2024-10-dapr-bindings.md
```diff
@@ -0,0 +1,276 @@
...
+- Adding the corresponding DaprBinding type to the internal representation (DaprRP)
+- Allowing a new type of Dapr component (binding) to be emitted in the backend Kubernetes cluster.
+
+#### Advantages (of each option considered)
+
+The main advantage of this implementation approach is that it is purely additive and will be non-breaking for existing users.
+
+#### Disadvantages (of each option considered)
+
+The main disadvantage of this approach is that it will lead to some code duplication.
```
lakshmimsft, line 195: Can you pls elaborate more.

resources/2024-10-dapr-bindings.md
```diff
@@ -0,0 +1,276 @@
...
+  @doc("A collection of references to resources associated with the binding")
+  resources?: ResourceReference[];
+
+  ...RecipeBaseProperties;
+}
+```
+
+### Implementation Details
+
+As for every Dapr Building Block, the implementation will need to create in the Dapr RP:
```
lakshmimsft, line 230: we should rename 'Dapr RP' everywhere in the document to 'Applications.Dapr RP/resource provider'

resources/2024-10-dapr-bindings.md
```diff
@@ -0,0 +1,276 @@
...
+
+This implementation will also have some side-effects outside of the Dapr RP:
+- Updates to the allowed resource type in the Portable Resource Renderer
+- Updates to the `getResourceDataByID` function in the Core RP to take the new type into account 
+- Updates to the `ResourceTypesList` variable in the CLI (cli/clients/managements.go) to be able to list the new type
+
+### Error Handling
+
+No new error handling is required. The error handling will remain the same as for other Dapr Building Blocks.
+
```
lakshmimsft, line 242: Can we describe briefly how the system will handle errors when communicating with the external systems in case  they are down or unresponsive?

resources/2024-10-dapr-bindings.md
```diff
@@ -0,0 +1,276 @@
...
+- converter functions
+- processor functions
+
+Functional tests proposed:
+- Using a binding in a sample application (the direction, input/output, of the binding is not important for this test). The external system can be something simple like Redis or a [CRON binding](https://docs.dapr.io/reference/components-reference/supported-bindings/cron/) that doesn't require an external system.
+- Using a binding with secret-store indirection, which will require creating and using a secret store in the binding.
+
+## Security
+
+N/A
```
lakshmimsft, line 255: We may need some details here on how resources can/should connect and communicate securely with external resources with, say, using secret stores.

resources/2024-10-dapr-bindings.md
```diff
@@ -0,0 +1,276 @@
...
+        Application
+        Dapr_Sidecar
+    end
+
+    subgraph Kubernetes
+        Pod
+        Dapr_Binding
+    end
+```	
+
```
lakshmimsft, line 177: Thanks a lot for the diagram! If possible pls add more details on calls eg. input versus output binding request

resources/2024-10-dapr-bindings.md
```diff
@@ -0,0 +1,276 @@
...
+  ...EnvironmentScopedResource;
+  ...DaprResourceProperties;
+
+  @doc("A collection of references to resources associated with the binding")
+  resources?: ResourceReference[];
+
+  ...RecipeBaseProperties;
+}
+```
+
```
lakshmimsft, line 227: 👍

resources/2024-10-dapr-bindings.md
```diff
@@ -0,0 +1,276 @@
...
+  properties: {
+    connections: {
+      queue: {
+        source: jobsQueue.id
+      }
+    }
+  ...
+  }
+}
+```
```
lakshmimsft, line 72: 👍 thanks for the detailed code snippets and examples!

lakshmimsft:
Hello @SoTrx. Big thanks for the design doc PR. I've given some initial comments mainly requesting more details in some areas but looking good overall. Thanks a lot!

resources/2024-10-dapr-bindings.md
```diff
@@ -0,0 +1,276 @@
...
+> **Issue Reference:** radius-project/radius#7960
+
+### Goals
+
+- Enable users to create, update, and delete Dapr Bindings using Radius.
+
+### Non goals
+
+- (out-of-scope) Managing resources used by Dapr Bindings. This includes databases, queues, and storage systems. By definition, Dapr Bindings are intended to be used for external resources.
+- (out-of-scope) Default recipes for Dapr Bindings. Each binding will have different requirements, so it doesn't make sense to have a single default recipe.
```
lakshmimsft, line 31: Let's discuss this in the design meeting. I can imagine that recipes may be valuable here.

resources/2024-10-dapr-bindings.md
SoTrx: I've added some information on that topic, but it may be out of scope as it describes how the Dapr sidecar reacts.

resources/2024-10-dapr-bindings.md
SoTrx: Sure !

rynowak (commented):
Hey @SoTrx - just wanted to drop by and say thanks for doing this. I'll be traveling during the discussion on Monday. I think the template is well filled out 👍 and will lead to the right discussion.

resources/2024-10-dapr-bindings.md
```diff
@@ -0,0 +1,326 @@
...
+
+### User scenarios (optional)
```
willtsai, line 33: I think these focus on user scenarios for the Dapr Bindings themselves, which is helpful, but I'd recommend adding some scenarios on how a user would use Radius to bring a Dapr Binding into their application

kachawla (commented):
Thanks @SoTrx, this is a well written doc! Main area that I would like to discuss is the proposal to support manual provisioning instead of recipe based provisioning, pros and cons that led to this decision.

ytimocin:
> Thanks @SoTrx, this is a well written doc! Main area that I would like to discuss is the proposal to support manual provisioning instead of recipe based provisioning, pros and cons that led to this decision.

Just to better understand @SoTrx, we are going to be supporting both types of provisioning by default, right?

resources/2024-10-dapr-bindings.md
```diff
@@ -0,0 +1,326 @@
...
+
+> **Issue Reference:** radius-project/radius#7960
+
+### Goals
+
+- Enable users to create, update, and delete Dapr Bindings using Radius.
+
+### Non goals
+
+- (out-of-scope) Managing resources used by Dapr Bindings. This includes databases, queues, and storage systems. By definition, Dapr Bindings are intended to be used for external resources.
```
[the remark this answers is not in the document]
SoTrx: (Recap meeting)
Typically, there are two entities to manage with a Dapr building block: the Dapr CRD and the associated resource it connects to.

Bindings are unique because the resource they connect to is external to the deployed application. Therefore, bindings only create the CRD and do not manage or provision the external resource.

The highlighted sentence will be corrected to reflect this distinction

resources/2024-10-dapr-bindings.md
```diff
@@ -0,0 +1,326 @@
...
+  // Process the message
+  // ...
+
+  // Acknowledge the message
+  w.WriteHeader(http.StatusOK)     
+}
+
+func main() {
+	r := mux.NewRouter()
+  // Each time a request is received via the input binding
```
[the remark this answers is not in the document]
SoTrx: (Recap meeting)
Each binding has a direction metadata, indicating whether it is an output binding or an input binding. Some bindings can function as both input and output bindings.

resources/2024-10-dapr-bindings.md
```diff
@@ -0,0 +1,326 @@
...
+
+#### Architectural components 
+
+##### Input Binding
+```mermaid
+graph LR
+    Client -->|Request| Engine
+    Engine -->|Deploy| Kubernetes
+    Dapr_Sidecar -->|1 - Resolves| Dapr_Binding
+    Dapr_Sidecar <--> |2 - Long polling / Subscription | Service_X
```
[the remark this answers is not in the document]
SoTrx: (Recap meeting)
A Dapr subscription is indeed another CRD, used to create what's called a declarative subscription.

However, this is not what's reflected in the mermaid diagram. In this case, "subscription" refers to the Dapr sidecar using its integrated SDKs to establish a connection to an external resource. Depending on the resource being listened to, this connection could involve long polling or a subscription to a topic (e.g., subscribing to a RabbitMQ queue using the SDK).

This will be corrected in the diagram, as it is misleading

resources/2024-10-dapr-bindings.md
```diff
@@ -0,0 +1,326 @@
...
+
+Similarly, the converter code is nearly identical for all Dapr building blocks, differing only in the resource type being converted.
+
+Example:
+- [Pub/Sub converter](https://github.com/radius-project/radius/blob/main/pkg/daprrp/api/v20231001preview/pubsubbroker_conversion.go)
+- [Configuration store converter](https://github.com/radius-project/radius/blob/main/pkg/daprrp/api/v20231001preview/configurationstore_conversion.go)
+
+#### Proposed Option
+
+This option is the most straightforward and efficient in terms of development time. Since the Dapr Building Block implementation is not yet complete, it would be better to have a simple implementation that can be improved later, even if it introduces some code duplication. Additionally, future implementations of User-Defined Types may lead to a refactor of the entire Applications.Dapr resource provider.
```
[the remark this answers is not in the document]
SoTrx: (Recap meeting)
My proposal is to wait either until all the building blocks are implemented to do some refactoring or to wait for User Defined Types to assess the consequences.

resources/2024-10-dapr-bindings.md
SoTrx: (Recap meeting)
The user story will be updated to reflect that

SoTrx:
> > Thanks @SoTrx, this is a well written doc! Main area that I would like to discuss is the proposal to support manual provisioning instead of recipe based provisioning, pros and cons that led to this decision.
> 
> Just to better understand @SoTrx, we are going to be supporting both types of provisioning by default, right?

In my view, recipes aren't necessary for bindings, as they only create the CRD and could lead to a two-step process for creating a single resource.

However, since this indeed contradicts Radius' design policy, how about introducing "common usage binding recipes"?

For instance, we could define an "object-storage binding recipe" that:

    In local development: creates both the binding CRD and the storage backend (e.g., MinIO).
    In production (Azure): only creates the binding CRD, with a parameter in the recipe to provide the storage account details.

This approach aligns with Radius' philosophy, is convenient for developers, and retains the core concept of bindings.

resources/2024-10-dapr-bindings.md
```diff
@@ -0,0 +1,326 @@
...
+#### User story 2
+
+As a Radius user, I want my application to send messages to external systems without having direct dependencies on them.
+
+This example uses an external SMTP server to send emails. The SMTP server could be centralized within a company's infrastructure or an external service like SendGrid.
+
+```bicep
+// Output binding to an SMTP server. 
+// https://docs.dapr.io/reference/components-reference/supported-bindings/smtp/
+resource mailing 'Applications.Dapr/bindings@2023-10-01-preview' = {
```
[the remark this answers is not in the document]
SoTrx: (Recap meeting)
Yes, Dapr is required for bindings to work. Behind the scenes, a binding instructs the Dapr sidecar to perform a specific type of action

resources/2024-10-dapr-bindings.md
```diff
@@ -0,0 +1,326 @@
...
+        Service_X
+    end
+
+    subgraph Pod
+        Application
+        Dapr_Sidecar
+    end
+
+    subgraph Kubernetes
+        Dapr_Binding
```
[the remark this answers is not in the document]
SoTrx: (Recap meeting)
It's the CRD !
I'll update the graph to represent this

resources/2024-10-dapr-bindings.md
```diff
@@ -0,0 +1,368 @@
...
+> **Issue Reference:** radius-project/radius#7960
+
+### Goals
+
+- Enable users to create, update, and delete Dapr Bindings Component (CRD) using Radius.
+- Create a list of recipes covering common use cases for Dapr Bindings. 
+
+### Non goals
+
+- (out-of-scope) Managing **external resources** used by Dapr Bindings. As we do not have management rights over the external resources, we will not be able to create, update, or delete them. The user will need to create the external resource and provide the necessary information to the Dapr Binding.
```
rynowak, line 39: I think it's also fair to say that this is *in-scope* when the user writes recipes. Writing or choosing the recipes will be user's responsibility.

resources/2024-10-dapr-bindings.md
```diff
@@ -0,0 +1,368 @@
...
+#### Common Usage Recipes
+
+Since a Binding is intended to interact with an external system outside of the operator control, a binding recipe would create only the CRD. This approach would require a two-step process to create a single CRD, which could otherwise be achieved in a single step using manual provisioning.
+
+To address this, rather than mapping [every possible binding](https://docs.dapr.io/reference/components-reference/supported-bindings/) to a recipe, the proposal is to create a few common usage recipes that cover the most frequently used cases.
+
+To maximize convenience for the user, these recipes will be designed with the following considerations:
+
+- **Local Development**: The recipe will create both the binding CRD and the "external resource".
+- **Production**: The recipe will create only the binding CRD, with the user providing the necessary information for connecting to the external resource.
```
rynowak, line 233: This is a valid usecase (pre-existing resource).

We'd usually think about the platform or infrastructure team as being the one to choose the recipes for production usage.

resources/2024-10-dapr-bindings.md
```diff
@@ -0,0 +1,368 @@
...
+### Goals
+
+- Enable users to create, update, and delete Dapr Bindings Component (CRD) using Radius.
+- Create a list of recipes covering common use cases for Dapr Bindings. 
+
+### Non goals
+
+- (out-of-scope) Managing **external resources** used by Dapr Bindings. As we do not have management rights over the external resources, we will not be able to create, update, or delete them. The user will need to create the external resource and provide the necessary information to the Dapr Binding.
+
+### User scenarios (optional)
```
willtsai, line 41: In both user scenarios, I recommend elaborating on the user ultimately leveraging Dapr Bindings to accomplish the abstraction, and then they are wanting to make use of these Dapr Bindings in Radius applications. e.g.
> As an operator using Radius, I want the application I manage to interface with other applications in my organization without exposing their implementation details to the developers by leveraging Dapr Bindings. Thus, I would like to define and use Dapr Bindings as a resource type within my Radius applications.

resources/2024-10-dapr-bindings.md
```diff
@@ -0,0 +1,368 @@
...
+This option is the most straightforward and efficient in terms of development time. Since the Dapr Building Block implementation is not yet complete, it would be better to have a simple implementation that can be improved later, even if it introduces some code duplication. Additionally, future implementations of User-Defined Types may lead to a refactor of the entire Applications.Dapr resource provider.
+
+### API design (if applicable)
+
+Aside from the new type, the API will remain identical to the other Dapr Building Blocks.
+
+**typespec/Applications.Dapr/bindings.tsp**
+```TypeSpec
+model DaprBindingResource
+  is TrackedResourceRequired<DaprBindingProperties, "DaprBindings"> {
```
rynowak, line 286: If the type name is `bindings` then this should be `TrackedResourceRequired<DaprBindingProperties, "bindings">`

resources/2024-10-dapr-bindings.md
```diff
@@ -0,0 +1,368 @@
...
+  - Production, AWS: Creates [CRD](https://docs.dapr.io/reference/components-reference/supported-bindings/s3/) only.
+
+- **Mailing Binding Recipe**
+  - Local Development: Creates both [CRD](https://docs.dapr.io/reference/components-reference/supported-bindings/smtp/) and a [MailDev](https://github.com/maildev/maildev) instance.
+  - Production: Creates [CRD](https://docs.dapr.io/reference/components-reference/supported-bindings/smtp/) only.
+
+### Detailed Design
+
+This design will require adding a new type to the Applications.Dapr resource provider. This includes:
+- Extending the TypeSpec schema definition to include the new Application.Dapr/Bindings resource
```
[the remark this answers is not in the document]
rynowak: Resource type names are `camelCase`.

resources/2024-10-dapr-bindings.md
```diff
@@ -0,0 +1,368 @@
...
+N/A
+
+## Alternatives considered
+
+N/A 
+
+## Design Review Notes
+
+Round 1:
+- In the first iteration, recipes were considered out of scope, as they would result in a two-step process to create a single CRD (the Dapr Binding component). This was reconsidered, as not supporting recipes for a resource contradicts the Radius design philosophy. Therefore, common usage recipes were added to the design.
```
rynowak, line 368: I don't quite understand this - apologies, I would out of the office when this was discussed. We have existing recipes for Dapr components that create both the infrastructure and the Dapr `Component` resource, so I'm not aware of any issues. LMK if I'm missing anything!

ytimocin (approved):
Other than a few nits, LGTM. Thanks!

rynowak (approved):
LGTM - thanks for the updates!

resources/2024-10-dapr-bindings.md
SoTrx: Sorry for the delay!

There is no issue with the creation process itself. You can create both the Binding CRD and the infrastructure just as you would with any other Dapr Building Block. Users can (and likely will) do this with their own configurations, and we certainly shouldn’t try to restrict that flexibility.

My point is only relevant to the initial binding recipes we plan to provide to users. Here’s some context: the Bindings Building Block is unique in Dapr. It predates most other building blocks and originally served as a way to accomplish tasks that did not yet have dedicated building blocks. This is why some bindings overlap with other blocks, such as Azure Storage Queue.

Bindings are generally basic, often supporting only CRUD-like operations. Additionally, Bindings is the only building block without a unified feature set across multiple backends—nearly every binding is different, and few are interchangeable.

Today, bindings are intended for interacting with external resources, as they are designed for basic interactions. For internal, app-scoped backing services, there is usually a better building block available.

While this approach doesn’t affect how Radius implements Bindings, it does impact the default recipes we include in radius/recipes. If we consider bindings best suited for external resources, then a binding recipe should only contain the Dapr Component CRD. This is why I initially opposed creating binding recipes, as it would involve a two-step process for what is effectively a single CRD creation.

In the first meeting, however, a very good point was raised: from Radius's perspective, it is difficult to justify not having any recipes for a resource.

So, my revised proposal is to create default recipes with this "external resource" perspective in mind. These recipes would set up the backing service in local-dev for convenience, but in production, they would have as input the credentials to an existing backing service. This approach isn’t meant to restrict users but rather as a way to nudge users in the right direction from the start. Of course, users can create custom recipes to override this behavior as they see fit.

I'm not entirely sure if I explained this clearly or if it make sense, so please don’t hesitate to challenge me or ask for clarification on any part of this! And if this direction makes sense, I’m happy to revise the design doc to reflect it.

rynowak: > So, my revised proposal is to create default recipes with this "external resource" perspective in mind. These recipes would set up the backing service in local-dev for convenience, but in production, they would have as input the credentials to an existing backing service. This approach isn’t meant to restrict users but rather as a way to nudge users in the right direction from the start. Of course, users can create custom recipes to override this behavior as they see fit.

Hey @SoTrx - I appreciate the response 👍 

I think the part where I'm stuck is why we need **recipes** for the "pre-existing" case. This is what `resourceProvisioning: 'manual'` is for.... when you want to configure the metadata directly.  

I'm happy to continue this discussion when we start looking at recipe examples since it's not really going to change the design of the `bindings` resource type.

rynowak:
Hi @SoTrx - can you take a look at the spellcheck errors? https://github.com/radius-project/design-notes/actions/runs/11671732590/job/32742154388?pr=68

For any false-positives reported by the spellcheck you can add them here: https://github.com/radius-project/design-notes/blob/main/.github/config/en-custom.txt

Or if they are technical terms I like to use backquotes to turn them into code. ``like this!``

After that I think this is ready to merge.

SoTrx:
> Hi @SoTrx - can you take a look at the spellcheck errors? https://github.com/radius-project/design-notes/actions/runs/11671732590/job/32742154388?pr=68
> 
> For any false-positives reported by the spellcheck you can add them here: https://github.com/radius-project/design-notes/blob/main/.github/config/en-custom.txt
> 
> Or if they are technical terms I like to use backquotes to turn them into code. `like this!`
> 
> After that I think this is ready to merge.
Hey Ryan, 
Sorry, I haven’t been able to work on this lately. Could you rerun the CI? I can’t view the latest spellcheck summary.
Thanks!

rynowak (approved):

rynowak:
Hi @SoTrx - there's a few more spelling errors here: https://github.com/radius-project/design-notes/actions/runs/12241896918/job/34148057351

Can you take a look?

rynowak:
Looks like the merge conflicts keep coming. Would be it be helpful for one of us maintainers to help finalize this? I know it's hard when things keep moving fast.

rynowak (approved):

rynowak:
Merging!

PR #70, opened 2024-10-18
create CODEOWNERS

willtsai:
Adding a CODEOWNERS file for administering repo membership and roles.

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -1,0 +1,2 @@
+# These owners are the maintainers and approvers of this repo
+*       @radius-project/maintainers-design-notes @radius-project/approvers-design-notes


sylvainsf (approved):
:shipit:

CODEOWNERS
```diff
@@ -0,0 +1,2 @@
+# These owners are the maintainers and approvers of this repo
+*       @radius-project/maintainers-design-notes @radius-project/approvers-design-notes
```
kachawla, line 2: @willtsai we should also update the CODEOWNERS file under .github directory: https://github.com/radius-project/design-notes/blob/main/.github/CODEOWNERS. Should do a check in all the repos to make sure the groups are up to date. Logged this issue for tracking https://github.com/radius-project/radius/issues/8074

PR #73, opened 2024-11-03
Threat Model for the UCP Component of Radius

ytimocin:
Threat Model for the UCP Component of Radius


architecture/2024-11-ucp-component-threat-model.md
```diff
@@ -0,0 +1,247 @@
...
+
+One of the most important goals of Radius is to be cloud agnostic and to enable portability of applications across platforms and environments. Therefore, the Radius UCP (Universal Control Plane) component helps the user apply ARM capabilities to non-Azure hosting models like AWS, GCP, and Kubernetes.
+
+In brief, UCP is a lightweight proxy that stands in front of the resource providers like Applications RP, AWS, and Azure. Based on the requests UCP gets, it forwards those requests to the appropriate service.
+
+## Terms and Definitions
+
+| Term                    | Definition                                                                                                                                  |
+| ----------------------- | ------------------------------------------------------------------------------------------------------------------------------------------- |
+| Radius Tracked Resource | Ex: `/planes/radius/local/resourceGroups/test-group/providers/System.Resources/resources/test-app-303153687ee5adbcf353bc6c2caa4373f31e04c6` |
```
rynowak, line 17: This isn't really sufficient to define what a tracked resource is. Is there a way we can talk about the security concerns without an extensive discussion of what the feature is?

architecture/2024-11-ucp-component-threat-model.md
```diff
@@ -0,0 +1,247 @@
...
+1. Radius user runs `rad deploy app.bicep`. `app.bicep` includes a Radius environment, a Radius application, a Redis Cache on AWS.
+2. ARM JSON Template is generated from the `app.bicep` file.
+3. CLI makes a PUT call to UCP: `/planes/radius/local/resourcegroups/{rg}/providers/Microsoft.Resources/deployments/{name}`
+4. UCP forwards that request to the Deployment Engine: `/planes/radius/local/resourcegroups/{rg}/providers/Microsoft.Resources/deployments/{name}`
+5. Deployment Engine makes 3 calls back to the UCP:
+   1. PUT for the Radius environment.
+   2. PUT for the Radius application.
+   3. PUT for the Redis Cache on AWS.
+6. UCP proxies these requests to the necessary resource providers and then starts polling the statuses of these operations.
+7. Deployment Engine also polls the operation statuses and results for these 3 resources.
```
rynowak, line 32: I think this is a good data flow that captures the majority of UCP's functionality and interactions with other systems.

architecture/2024-11-ucp-component-threat-model.md
```diff
@@ -0,0 +1,247 @@
...
+
```
nithyatsu, line 33: I think we poll only for the AWS resource, since app and env are both sync calls.

architecture/2024-11-ucp-component-threat-model.md
```diff
@@ -0,0 +1,247 @@
...
+- **Author**: ytimocin
+
+## Overview
+
+This document provides a threat model for the Radius UCP (Universal Control Plane) component. It identifies potential security threats to this critical part of Radius and suggests possible mitigations. The document includes an analysis of the system, its assets, identified threats, and recommended security measures to protect the system.
+
+One of the most important goals of Radius is to be cloud agnostic and to enable portability of applications across platforms and environments. Therefore, the Radius UCP (Universal Control Plane) component helps the user apply ARM capabilities to non-Azure hosting models like AWS, GCP, and Kubernetes.
+
+In brief, UCP is a lightweight proxy that stands in front of the resource providers like Applications RP, AWS, and Azure. Based on the requests UCP gets, it forwards those requests to the appropriate service.
+
```
rynowak, line 12: I'd expect to see a few things highlighted here:

- UCP may have credentials for Azure or AWS, and may use those credentials to manage resources on the user's behalf.
- UCP is responsible for routing HTTP traffic from clients to each destination (resource provider). In the future this will include RBAC and authentication. Right now there is no RBAC and traffic between UCP and RPs is unauthenticated.

architecture/2024-11-ucp-component-threat-model.md
```diff
@@ -0,0 +1,247 @@
...
+   2. PUT for the Radius application.
+   3. PUT for the Redis Cache on AWS.
+6. UCP proxies these requests to the necessary resource providers and then starts polling the statuses of these operations.
+7. Deployment Engine also polls the operation statuses and results for these 3 resources.
+
+### Architecture
+
+The UCP (Universal Control Plane) consists of several important pieces:
+
+- **Storage Provider**: Radius needs a data store to store all the information related to the resources of the installation. UCP keeps the information of resources by converting them to tracked resources. On the other hand, available implementations of the Radius Storage Provider:
```
brooke-hamilton, line 38: Typo - 
```suggestion
- **Storage Provider**: Radius needs a data store to store all the information related to the resources of the installation. UCP keeps the information of resources by converting them to tracked resources. Available implementations of the Radius Storage Provider:
```

architecture/2024-11-ucp-component-threat-model.md
ytimocin: Deployment Engine polls for all the resources that it requests from the resource providers via UCP.

architecture/2024-11-ucp-component-threat-model.md
```diff
@@ -0,0 +1,247 @@
...
+  1. Cosmos Database
+  2. etcd (in-memory or persistent)
+  3. In-Memory storage
+  4. PostgreSQL
+
+- **Secret Provider**: The UCP occasionally needs to create and store secrets. Available implementations of the UCP Secret Provider are:
+
+  1. etcd
+  2. Kubernetes Secrets
+
```
lakshmimsft, line 49: ooo..there's a new in-memory provider coming in

architecture/2024-11-ucp-component-threat-model.md
```diff
@@ -0,0 +1,247 @@
...
+1. **Creating or Updating a Cloud Provider Credential Resource**: [Azure](https://github.com/radius-project/radius/blob/95409fe179d7adca884a3fc1d82f326bc81c8da0/pkg/ucp/frontend/controller/credentials/azure/createorupdateazurecredential.go#L89), [AWS](https://github.com/radius-project/radius/blob/95409fe179d7adca884a3fc1d82f326bc81c8da0/pkg/ucp/frontend/controller/credentials/aws/createorupdateawscredential.go#L81). When a **Cloud Provider Credential** resource is deployed, Radius creates a new Kubernetes Secret or updates an existing one. The Kubernetes Secret is deleted when the **Cloud Provider Credential** is deleted.
+
+#### Data Serialization / Formats
+
+We use custom parsers to parse Radius-related resource IDs and do not use any other custom parsers and instead rely on Kubernetes built-in parsers. Therefore, we trust Kubernetes security measures to handle data serialization and formats securely. The custom parser that parses Radius resource IDs has its own security mechanisms that don't accept anything other than a Radius resource ID.
```
rynowak, line 101: Was this copied from the controller threat model?

I don't think we're using Kubernetes to parse JSON.

ytimocin: Yes, this was copied from the Controller Threat Model. I will update this part accordingly.

architecture/2024-11-ucp-component-threat-model.md
```diff
@@ -0,0 +1,247 @@
...
+In all instances where SHA-1 is utilized within the codebase, it serves for generating hashes to ensure uniqueness or detect changes efficiently. It is **not** used for cryptographic security purposes due to known vulnerabilities. For security-critical hashing, a more secure algorithm (e.g., SHA-256) should be employed.
+
+---
+
+#### Storage of Secrets
+
+Below you will find where and how Radius stores secrets. We create Kubernetes Secret objects and rely on Kubernetes security measures to protect these secrets.
+
+1. **Creating or Updating a Cloud Provider Credential Resource**: [Azure](https://github.com/radius-project/radius/blob/95409fe179d7adca884a3fc1d82f326bc81c8da0/pkg/ucp/frontend/controller/credentials/azure/createorupdateazurecredential.go#L89), [AWS](https://github.com/radius-project/radius/blob/95409fe179d7adca884a3fc1d82f326bc81c8da0/pkg/ucp/frontend/controller/credentials/aws/createorupdateawscredential.go#L81). When a **Cloud Provider Credential** resource is deployed, Radius creates a new Kubernetes Secret or updates an existing one. The Kubernetes Secret is deleted when the **Cloud Provider Credential** is deleted.
+
```
nithyatsu, line 98: Might be worth mentioning these credentials cannot be retrieved or updated through API calls.

architecture/2024-11-ucp-component-threat-model.md
```diff
@@ -0,0 +1,247 @@
...
+   1. **Purpose:** To hash input data to generate new ETags for resource versioning.
+   2. **Library:** Uses the Go standard `crypto/sha1` package: [crypto/sha1](https://pkg.go.dev/crypto/sha1)
+   3. **Type:** SHA-1  
+      _Note:_ SHA-1 usage here is for generating ETags and not for cryptographic security.
+
+---
+
+##### General Note on SHA-1 Usage
+
+In all instances where SHA-1 is utilized within the codebase, it serves for generating hashes to ensure uniqueness or detect changes efficiently. It is **not** used for cryptographic security purposes due to known vulnerabilities. For security-critical hashing, a more secure algorithm (e.g., SHA-256) should be employed.
```
brooke-hamilton, line 89: What is the reason for sticking with SHA-1?

architecture/2024-11-ucp-component-threat-model.md
```diff
@@ -0,0 +1,247 @@
...
+
+### Clients
+
+In this section, we will discuss the different clients of the Radius UCP (Universal Control Plane) component. Clients are systems that interact with the UCP component to trigger actions. Here are the clients of the UCP component:
+
+1. **Radius CLI**: Radius CLI interacts with UCP under the hood through the generated clients. For example, a `rad deploy` command needs to go to UCP and then through UCP to the Deployment Engine. UCP will do the forwarding of the requests to the Deployment Engine and the other components like Applications RP.
+
+2. **Direct API Querying**: Clients can interact with the UCP component directly by making API queries. For example, a client can query the status of a deployment or retrieve information about a specific resource by making a direct API call to the UCP server.
```
rynowak, line 109: **Every** component of Radius is a client of UCP 😆. I think we can simplify this section. There's a lot of text here, but it's not conveying a ton of information.

architecture/2024-11-ucp-component-threat-model.md
ytimocin: I agree. I will update the definition of the Tracked Resource. Not sure if that is what you mean, but we can also not talk about Tracked Resource, I guess.

architecture/2024-11-ucp-component-threat-model.md
rynowak: Here's a question that **is** worth answering. How do these clients authenticate?

architecture/2024-11-ucp-component-threat-model.md
```diff
@@ -0,0 +1,247 @@
...
+
+## Trust Boundaries
+
+We have a few different trust boundaries for the UCP component:
+
+- **Kubernetes Cluster**: The overall environment where the UCP component operates and receives requests from the clients.
+- **Namespaces within the Cluster**: Logical partitions within the cluster to separate and isolate resources and workloads.
+
+The UCP component lives inside the `radius-system` namespace in the Kubernetes cluster where it is installed. Other components of Radius like Controller, Deployment Engine, Dashboard, and Applications RP also live in the same namespace.
+
```
rynowak, line 156: What can we say about the trust model of clients?

What about the trust model of resource providers?

What about the trust model of external resource managers? (Azure or AWS)

architecture/2024-11-ucp-component-threat-model.md
ytimocin: Is Applications-RP a client of UCP? By client, I mean, is Applications-RP making a call to UCP? Or is UCP just forwarding the requests to Applications-RP?

architecture/2024-11-ucp-component-threat-model.md
ytimocin: Let's talk about the trust models of clients, resource providers, and external resource managers.

For resource providers, I think, the trust model is the `radius-system` namespace.

Some clients are in the same namespace, but some clients are not.

architecture/2024-11-ucp-component-threat-model.md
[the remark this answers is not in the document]
ytimocin: We also have the **etcd** implementation available and that is why I wanted to mention it.

architecture/2024-11-ucp-component-threat-model.md
ytimocin: Good idea.

architecture/2024-11-ucp-component-threat-model.md
rynowak: When I see a term in the definitions, I'd expect to see a clear definition. As a reader, I'd want to make sure I understand the term before reading the document.

I don't think we need every reader of the document to understand what a tracked resource is. I think the content in the implementation details section is enough, so you could remove this.

architecture/2024-11-ucp-component-threat-model.md
```diff
@@ -0,0 +1,247 @@
...
+### Architecture
+
+The UCP (Universal Control Plane) consists of several important pieces:
+
+- **Storage Provider**: Radius needs a data store to store all the information related to the resources of the installation. UCP keeps the information of resources by converting them to tracked resources. On the other hand, available implementations of the Radius Storage Provider:
+
+  1. Cosmos Database
+  2. etcd (in-memory or persistent)
+  3. In-Memory storage
+  4. PostgreSQL
```
rynowak, line 43: This list is missing `apiserver`, which is the implementation that's used for installed Radius.

ytimocin: I will add that one too.

architecture/2024-11-ucp-component-threat-model.md
nithyatsu, line 1: should we also capture that UCP receives payload on HTTP and validates it before consuming.

architecture/2024-11-ucp-component-threat-model.md
rynowak: >  Is Applications-RP a client of UCP? By client, I mean, is Applications-RP making a call to UCP? Or is UCP just forwarding the requests to Applications-RP?

Both. We should assume that every RP is both client and destination.

architecture/2024-11-ucp-component-threat-model.md
ytimocin: 1st: Resource ID parser
2nd: Go standard library
All data interchanges are done using JSON and we use GO Standard library..

architecture/2024-11-ucp-component-threat-model.md
ytimocin: Find out how auth works and how clients authenticate with UCP

ytimocin: What do we believe to be true when we send requests to App-RP or AWS or Azure?

architecture/2024-11-ucp-component-threat-model.md
ytimocin: What do we validate inside UCP in the incoming requests?

brooke-hamilton (approved):
This threat model is very clear and detailed. 🚀 I especially like the simple descriptions of the architecture and how UCP fits into the overall system.

architecture/2024-11-ucp-component-threat-model.md
```diff
@@ -0,0 +1,276 @@
...
+
+3. **Metrics Scrapers**: If metrics are enabled, Prometheus or other monitoring tools can scrape metrics from the Universal Control Plane.
+
+## Trust Boundaries
+
+### Trust Model of UCP Clients
+
+The clients of the UCP include other Radius components like Applications RP, CLI, Controller, Dashboard, and Deployment Engine. Most of these clients reside within the same Kubernetes cluster and namespace as the UCP, except for the Radius CLI, which operates outside of the Kubernetes cluster and, therefore, outside of the namespace.
+
+For clients that are in the same cluster and namespace as the UCP component, we rely on Kubernetes to implement the necessary security measures and establish the required trust boundary. For external clients--meaning clients that are outside of this trust boundary--we rely on the host machine's trust boundary and assume that it is not compromised. In such cases, proper authentication and authorization are essential to maintain security.
```
rynowak, line 154: We should really be talking about authentication here. Kubernetes namespaces apply RBAC to data stored in Kubernetes like Deployments, Secrets, etc.

Kubernetes namespaces **do not** limit traffic to our API. Please update this to talk about authentication.

architecture/2024-11-ucp-component-threat-model.md
```diff
@@ -0,0 +1,276 @@
...
+
+### Trust Model of Internal Resource providers
+
+Currently, our only resource provider is the Applications RP. As mentioned above, the Applications RP lives in the same Kubernetes cluster and namespace as the UCP. Therefore, the trust boundaries are defined by the Kubernetes cluster and the namespace within that cluster.
+
+### Trust Model of External Resource Manages (Azure and AWS)
+
+The UCP routes user requests to Azure and AWS whenever a resource from those providers is requested. In these cases, we rely on Azure and AWS to establish the trust boundary and operate under the assumption that they are secure and trustworthy.
```
rynowak, line 162: We should talk about how this is secured. I think we can make a strong assumption by discussing authentication.

architecture/2024-11-ucp-component-threat-model.md
```diff
@@ -0,0 +1,276 @@
...
+This threat model assumes that:
+
+1. The Radius installation is not tampered with.
+2. The Kubernetes cluster that Radius is installed on is not compromised.
+3. It is the responsibility of the Kubernetes cluster to authenticate users. Administrators and users with sufficient privileges can perform their required tasks. Radius cannot prevent actions taken by an administrator.
+4. The Data Store (can be one of these: etcd, Cosmos DB, PostgreSQL, in-memory, and API Server) that the UCP uses to keep important data is not compromised.
+5. The Secret Store (can be one of these: etcd, Kubernetes Secrets) that the UCP uses to keep important data is not compromised.
+6. The Queue that the UCP uses to write messages for the async operations is not compromised.
+7. External Resource Managers (Azure and AWS) are not compromised and are working as expected.
+
```
rynowak, line 175: These are good assumptions.

architecture/2024-11-ucp-component-threat-model.md
```diff
@@ -0,0 +1,276 @@
...
+1. **User runs `rad deploy app.bicep` using Radius CLI**: When a user runs `rad deploy app.bicep` using the Radius CLI, the Bicep file gets converted to a JSON Template file and is sent to the UCP.
+2. **UCP forwards the template to the Deployment Engine**
+3. **Deployment Engine sends the list of resources back to the UCP as requests**
+4. **UCP sends each request to the corresponding Resource Provider**
+5. **UCP updates its data store with the information of the resources**
+
+### Threats
+
+#### Threat: Spoofing Deployment Engine Could Cause Information Disclosure and Denial of Service
```
rynowak, line 190: This is true for any resource provider, not just the deployment engine.

architecture/2024-11-ucp-component-threat-model.md
```diff
@@ -0,0 +1,276 @@
...
+
+1. Secret data managed by the UCP is stored at rest in Kubernetes secrets or etcd. Our threat model assumes that the API server and mechanisms like Kubernetes authentication/RBAC are not compromised.
+2. Secrets managed by Radius are always placed in the same namespace as the object that "owns" them. This is a requirement of the Kubernetes RBAC model.
+3. Secrets managed by Radius are subject to the Kubernetes RBAC model for controlling access. Operators are expected to limit access for users using existing tools.
+
+**Status:** All mitigations listed are currently active. Operators are expected to secure their cluster and limit access for users.
+
+#### Threat: Escalation of Privilege by Using Radius to Circumvent Kubernetes RBAC Controls
+
+**Description:** A malicious actor could circumvent Kubernetes RBAC controls and create arbitrary resources in Kubernetes by using Universal Control Plane. UCP operates with a wide scope of permissions in Kubernetes and the cloud.
```
rynowak, line 219: > UCP operates with a wide scope of permissions in Kubernetes and the cloud.

This is not actually true for UCP. Can you please review the permissions that UCP has in Kubernetes and update this?

architecture/2024-11-ucp-component-threat-model.md
```diff
@@ -0,0 +1,276 @@
...
+
+**Mitigations:**
+
+1. **Implement RBAC within the UCP:** An authentication and authorization mechanism that verifies the identity of clients and enforces access policies must be developed.
+2. **Secure Communication Between UCP and Other Components:** mTLS should be enabled.
+3. **Network Policies and Firewall Rules:** Application of Kubernetes Network Policies to control traffic flow to and from the UCP.
+
+**Status:** Other than the second mitigation (secure communication) listed above, none of the mitigations is currently active.
+
+#### Threat: Use of Weak Cryptographic Hashing Algorithm (SHA-1)
```
rynowak, line 246: Can you be more specific about the threat? Complete the following sentence:

> A malicious actor could use X to do Y...

This reads like an action-item not a threat.

architecture/2024-11-ucp-component-threat-model.md
```diff
@@ -0,0 +1,276 @@
...
+1. **Replace SHA-1 with a Stronger Algorithm.**
+
+**Status:** The mitigation is not active as of now. An action item is created to update the crypto algorithm used in hashing resource IDs and generating ETags as well as in other components of Radius.
+
+## Open Questions
+
+## Action Items
+
+1. Use a hashing algorithm other than SHA-1 while computing the hash of resource IDs and generating ETags.
+2. Check if RBAC with Least Privilege is configured for every component to ensure that each component has only the permissions it needs to function. Make changes to the necessary components if required.
```
rynowak, line 263: > Check

I expect you to do that as part of writing this threat model. Please do that analysis and report back.

architecture/2024-11-ucp-component-threat-model.md
```diff
@@ -0,0 +1,276 @@
...
+3. Define and implement necessary Network Policies to ensure that communication is accepted only from expected and authorized components. Regularly review and update these policies to maintain security.
```
rynowak, line 264: Let's discuss this

architecture/2024-11-ucp-component-threat-model.md
```diff
@@ -0,0 +1,276 @@
...
+4. Containers should run as a non-root user wherever possible to minimize the risks. Check if we can run any of the Radius containers as non-root. Do the necessary updates.
```
rynowak, line 265: > Check if we can run any of the Radius containers as non-root. Do the necessary updates.

I expect you to do that as part of writing this threat model. Please do this analysis and report back.

architecture/2024-11-ucp-component-threat-model.md
ytimocin: Action Item: Secure communication between UCP and Resource Providers. Some form of Authentication and Network Policies and Documentation. mTLS is a form of authentication.

architecture/2024-11-ucp-component-threat-model.md
ytimocin: https://github.com/radius-project/design-notes/pull/73/files#diff-de92b57845f019bf07433840060635c493191b42971c4601f782cd98757d8c03R150

architecture/2024-11-ucp-component-threat-model.md
ytimocin: Added the details.

architecture/2024-11-ucp-component-threat-model.md
ytimocin: Should we talk about the credentials that the user provides us so that we can authenticate to these cloud providers?

architecture/2024-11-ucp-component-threat-model.md
```diff
@@ -0,0 +1,291 @@
...
+
+1. **Creating or Updating a Cloud Provider Credential Resource**: [Azure Implementation](https://github.com/radius-project/radius/blob/95409fe179d7adca884a3fc1d82f326bc81c8da0/pkg/ucp/frontend/controller/credentials/azure/createorupdateazurecredential.go#L89), [AWS Implementation](https://github.com/radius-project/radius/blob/95409fe179d7adca884a3fc1d82f326bc81c8da0/pkg/ucp/frontend/controller/credentials/aws/createorupdateawscredential.go#L81). When a **Cloud Provider Credential** resource is deployed, Radius creates a new Kubernetes Secret or updates an existing one. The Kubernetes Secret is deleted when the **Cloud Provider Credential** is deleted. We should also note that these credentials cannot be retrieved or updated through API calls.
+
+#### Data Serialization / Formats
+
+We use custom parsers exclusively for parsing Radius-related resource IDs. For all other parsing needs, we rely on the Go Standard Library's built-in parsers, trusting its security measures to handle data serialization and formats securely. Our custom parser for Radius resource IDs includes security mechanisms that ensure only valid Radius resource IDs are accepted.
```
rynowak, line 105: Can you reformat this paragraph? It jumps from topics in this order:

- We wrote a custom parser for resource IDs
- We use Go's parser for JSON
- Back to talking about our custom parser

rynowak: > Our custom parser for Radius resource IDs includes security mechanisms that ensure only valid Radius resource IDs are accepted.

Let's discuss and clarify this. I don't know what it means and I wrote the code.

architecture/2024-11-ucp-component-threat-model.md
```diff
@@ -0,0 +1,291 @@
...
+
+### Clients
+
+In this section, we will discuss the different clients of the Radius UCP (Universal Control Plane) component. Clients are systems that interact with the UCP component to trigger actions. Here are the clients of the UCP component:
+
+1. **All Components of Radius**: Every component other than UCP is a client of UCP. This list includes the Radius CLI, Deployment Engine, Controller, Deployment Engine, Dashboard, and Applications RP. An example of how Deployment Engine communicates with UCP can be found below:
+
+   ```text
```
rynowak, line 113: These are log messages from one of the services right? what's the value of including this to the reader?

architecture/2024-11-ucp-component-threat-model.md
```diff
@@ -0,0 +1,291 @@
...
+
+2. **Health Check Probes**: Kubernetes itself can act as a client by performing health and readiness checks on the Universal Control Plane.
+
+3. **Metrics Scrapers**: If metrics are enabled, Prometheus or other monitoring tools can scrape metrics from the Universal Control Plane.
+
+## Trust Boundaries
+
+### UCP Container runs as Non-Root
+
+As of November 24, 2024, the UCP container runs as a non-root user. This security measure helps to limit the potential impact of a container compromise by restricting the privileges available to the container. Running containers as non-root users is a best practice that enhances the overall security posture of the system. For more details, refer to the [Dockerfile](https://github.com/radius-project/radius/blob/96504063bbab8c0ee53b0955a9647cf00c7f5fae/deploy/images/ucpd/Dockerfile#L14).
```
rynowak, line 152: The fact that you did this analysis is good. I don't know that it's a trust boundary. Trust boundaries talk about the relationships and assumptions with external components and actors.

I'd suggest moving this to the implementation details.

architecture/2024-11-ucp-component-threat-model.md
```diff
@@ -0,0 +1,291 @@
...
+
+### Trust Model of UCP Clients
+
+UCP is a [Kubernetes API Service](https://kubernetes.io/docs/concepts/extend-kubernetes/api-extension/apiserver-aggregation/) that aims to extend Kubernetes with its APIs and capabilities. So anything that is sent to an API path that starts with `/apis/api.ucp.dev/v1alpha3` gets sent to the UCP through the Kubernetes API Server. This means that for external clients like Radius CLI, we rely on Kubernetes authentication and don't provide any additional authentication mechanism.
+
+![CLI - Kubernetes API Server - UPC](./2024-11-ucp-component-threat-model/cli-kubernetes-api-server-ucp.png)
```
rynowak, line 158: I'm glad you added this. I think if we frame this in the lens of "trust boundaries", we could say that requests coming into UCP from a client are originating from an **authorized user**.

architecture/2024-11-ucp-component-threat-model.md
```diff
@@ -0,0 +1,291 @@
...
+
+In terms of authentication between UCP and resource providers, we also don't have any authentication mechanisms right now. We use [AnonymousCredential](https://github.com/radius-project/radius/blob/main/pkg/azure/tokencredentials/anonymous.go#L50) while making calls to the resource providers from UCP. It basically means that there are no tokens in the request making the communication unauthenticated.
```
rynowak, line 160: This should be moved down a section.

architecture/2024-11-ucp-component-threat-model.md
```diff
@@ -0,0 +1,291 @@
...
+
+### Trust Model of Internal Resource providers
+
+Currently, our only resource provider is the Applications RP. As mentioned above, the Applications RP lives in the same Kubernetes cluster and namespace as the UCP. Therefore, the trust boundaries are defined by the Kubernetes cluster and the namespace within that cluster.
```
rynowak, line 164: > Therefore, the trust boundaries are defined by the Kubernetes cluster and the namespace within that cluster.

What does this mean? I feel like we keep discussing this topic. Kubernetes namespaces are not involved when we send traffic to a resource provider.

architecture/2024-11-ucp-component-threat-model.md
```diff
@@ -0,0 +1,291 @@
...
+
+### Trust Model of External Resource Managers (Azure and AWS)
+
+The UCP routes user requests to Azure and AWS whenever a resource from those providers is requested. In these cases, we rely on Azure and AWS to establish the trust boundary and operate under the assumption that they are secure and trustworthy. We provide necessary credentials that the user has provided to us and authenticate to these cloud providers. We, then, trigger the next action.
```
rynowak, line 168: I think most of the base ingredients are here. What's missing is DNS and TLS. 

You could say something like:

> We communicate with external resource managers through their public HTTPS APIs. We use standard authentication mechanisms, with credentials provided by the user. Therefore as assume that external resource managers are authentic, and our requests and subject to the authorization model provided by those resource managers.

architecture/2024-11-ucp-component-threat-model.md
```diff
@@ -0,0 +1,291 @@
...
+2. **UCP forwards the template to the Deployment Engine**
+3. **Deployment Engine sends the list of resources back to the UCP as requests**
+4. **UCP sends each request to the corresponding Resource Provider**
+5. **UCP updates its data store with the information of the resources**
+
+### Threats
+
+#### Threat: Spoofing a Resource Provider Could Cause Information Disclosure and Denial of Service
+
+**Description:** If a malicious actor can spoof the Deployment Engine by tampering with the configuration in the UCP, the UCP will start sending requests to the malicious server. The malicious server can capture the traffic, leading to information disclosure. This would effectively disable the UCP, causing a Denial of Service.
```
rynowak, line 198: Is this specific the deployment engine? or any resource provider?

rynowak: > This would effectively disable the UCP, causing a Denial of Service.

It would disable that resource provider's functionality.

architecture/2024-11-ucp-component-threat-model.md
```diff
@@ -0,0 +1,291 @@
...
+
+1. Secret data managed by the UCP is stored at rest in Kubernetes secrets or etcd. Our threat model assumes that the API server and mechanisms like Kubernetes authentication/RBAC are not compromised.
+2. Secrets managed by Radius are always placed in the same namespace as the object that "owns" them. This is a requirement of the Kubernetes RBAC model.
+3. Secrets managed by Radius are subject to the Kubernetes RBAC model for controlling access. Operators are expected to limit access for users using existing tools.
+
+**Status:** All mitigations listed are currently active. Operators are expected to secure their cluster and limit access for users.
+
+#### Threat: Escalation of Privilege by Using Radius to Circumvent Kubernetes RBAC Controls
+
+**Description:** A malicious actor could circumvent Kubernetes RBAC controls and create arbitrary resources in Kubernetes by exploiting the Universal Control Plane (UCP). The UCP has the following permissions as of November 24, 2024:
```
rynowak, line 225: Can you be more specific? What's the exploit?

architecture/2024-11-ucp-component-threat-model.md
```diff
@@ -0,0 +1,291 @@
...
+   2. Implementation: Review and update the RBAC policies regularly to ensure they adhere to the principle of least privilege.
+2. **Enhance Authorization Controls:**
+   1. Action: Develop and implement more granular authorization controls within the UCP to enforce fine-grained access policies.
+   2. Implementation: Introduce role-based access controls within the UCP itself, allowing for more specific permissions and roles tailored to different user needs.
+
+**Status:** These mitigations are partial and require configuration by the operator. We will revisit and improve this area in the future.
+
+#### Threat: Lack of Role-Based Access Control (RBAC) and Unauthorized Traffic
+
+**Description:** As mentioned above, as of November 15, 2024, the UCP does not implement RBAC, and communication between the UCP and resource providers is unauthenticated.
```
rynowak, line 245: This is making two points, and they appear unrelated:

- UCP does not implement RBAC
- communication between the UCP and resource providers is unauthenticated

rynowak: Can you clarify the threat? 

> A malicious actor could do X to accomplish Y

architecture/2024-11-ucp-component-threat-model.md
```diff
@@ -0,0 +1,291 @@
...
+3. **Network Policies and Firewall Rules:** Application of Kubernetes Network Policies to control traffic flow to and from the UCP.
+
+**Status:** Other than the second mitigation (secure communication) listed above, none of the mitigations is currently active.
+
+#### Threat: A Malicious Actor Could Exploit SHA-1 Weaknesses to Generate Hash Collisions
+
+**Description:** A malicious actor could exploit the known vulnerabilities of the SHA-1 hashing algorithm to generate hash collisions. The UCP currently uses SHA-1 for hashing resource IDs and generating ETags. This could potentially allow an attacker to create two different inputs that produce the same hash value, leading to unauthorized access or data manipulation.
+
+**Impact:** Although SHA-1 is used for non-security purposes, its vulnerabilities could be exploited to create hash collisions. This could result in unauthorized access to resources, data corruption, or other security weaknesses if the hashes are used in security-sensitive contexts. Using a stronger cryptographic hashing algorithm is essential to ensure the integrity and security of the system.
```
rynowak, line 261: > This could potentially allow an attacker to create two different inputs that produce the same hash value, leading to unauthorized access or data manipulation.

Can you be more specific? This document claims that we don't use SHA-1 in any security-sensitive context.

architecture/2024-11-ucp-component-threat-model.md
ytimocin: Action Item: We built a custom parser, and we need to do fuzz testing.

ytimocin: Don't say we did security. Red flag.

architecture/2024-11-ucp-component-threat-model.md
ytimocin: Remove this example.

architecture/2024-11-ucp-component-threat-model.md
ytimocin: Removed this part

architecture/2024-11-ucp-component-threat-model.md
ytimocin: Done

architecture/2024-11-ucp-component-threat-model.md
ytimocin: Updated this part

architecture/2024-11-ucp-component-threat-model.md
ytimocin: Updated this part:

```
The UCP routes user requests to Azure and AWS when resources from these providers are needed. We rely on Azure and AWS to establish the trust boundary and assume they are secure and trustworthy. We use the credentials provided by the user to authenticate with these cloud providers and then trigger the necessary actions.

Communication with external resource managers is done through their public HTTPS APIs using standard authentication mechanisms. Therefore, we assume that these external resource managers are authentic and that our requests are subject to their authorization models.
```

architecture/2024-11-ucp-component-threat-model.md
ytimocin: Simplified this part and updated according to the feedback.

architecture/2024-11-ucp-component-threat-model.md
ytimocin: Updated the threats with more information and updated the titles.

architecture/2024-11-ucp-component-threat-model.md
ytimocin: Added an action item for this.

architecture/2024-11-ucp-component-threat-model.md
ytimocin: Updated this part accordingly.

architecture/2024-11-ucp-component-threat-model.md
```diff
@@ -0,0 +1,259 @@
...
+**Mitigations:**
+
+1. **Secure Communication Between UCP and Resource Providers:** Implement a form of authentication (e.g., mTLS) to verify the authenticity of the communication between the UCP and resource providers.
+2. **Network Policies and Firewall Rules:** Apply Kubernetes Network Policies to control traffic flow to and from the UCP. Configure firewalls to allow only necessary traffic between trusted components.
+
+**Status:** None of the mitigations is currently active and we created action items to work on them.
+
+#### Threat: A Malicious Actor Could Exploit SHA-1 Weaknesses to Generate Hash Collisions
+
+**Description:** A malicious actor could exploit the known vulnerabilities of the SHA-1 hashing algorithm to generate hash collisions. The UCP currently uses SHA-1 for hashing resource IDs and generating ETags. Although SHA-1 is not used in security-sensitive contexts, its vulnerabilities could still be exploited to create hash collisions, potentially leading to data integrity issues.
```
rynowak, line 227: I don't understand why this is a security threat. It's too vague.

architecture/2024-11-ucp-component-threat-model.md
```diff
@@ -0,0 +1,259 @@
...
+
+1. Secret data managed by the UCP is stored at rest in Kubernetes secrets or etcd. Our threat model assumes that the API server and mechanisms like Kubernetes authentication/RBAC are not compromised.
+2. Secrets managed by Radius are always placed in the same namespace as the object that "owns" them. This is a requirement of the Kubernetes RBAC model.
+3. Secrets managed by Radius are subject to the Kubernetes RBAC model for controlling access. Operators are expected to limit access for users using existing tools.
+
+**Status:** All mitigations listed are currently active. Operators are expected to secure their cluster and limit access for users.
+
+#### Threat: A Malicious Actor Could Exploit Lack of RBAC to Gain Unauthorized Access and Escalate Privileges
+
+**Description:** A malicious actor could exploit the lack of Role-Based Access Control (RBAC) within the Universal Control Plane (UCP) to gain unauthorized access and escalate privileges. By circumventing Kubernetes RBAC controls, the attacker could use the UCP's extensive permissions to create, delete, or modify resources within the Kubernetes cluster. The UCP has the following permissions as of November 24, 2024:
```
rynowak, line 194: Can you be more specific? How would a malicious actor gain unauthorized access?

architecture/2024-11-ucp-component-threat-model.md
rynowak: Please be more specific. A malicious user could do X to achieve Y. Y should be grounded in one of these: https://en.wikipedia.org/wiki/STRIDE_model

architecture/2024-11-ucp-component-threat-model.md
```diff
@@ -0,0 +1,259 @@
...
+   1. **Action:** Identify all instances where SHA-1 is used in the codebase.
+   2. **Implementation:** Replace SHA-1 with SHA-256 or another secure hashing algorithm for hashing resource IDs and generating ETags. Ensure that the new algorithm is consistently used across all components. Test to verify that the change does not impact system functionality.
+
+**Status:** The mitigation is not active as of now. An action item has been created to update the cryptographic algorithm used in hashing resource IDs and generating ETags, as well as in other components of Radius.
+
+## Open Questions
+
+## Action Items
+
+1. **Use a Stronger Hashing Algorithm:**
```
rynowak, line 243: I don't understand this as an action item based on the document. The document says both that SHA-1 is used for non-cryptographic purposes, and also that it's a security threat. 

It feels like you're taking a compromise position that's going to result in a lot of work for nebulous benefit. 

Here's what I'd suggest:

- Let's review [this threat](https://github.com/radius-project/design-notes/pull/73/files#diff-de92b57845f019bf07433840060635c493191b42971c4601f782cd98757d8c03R227) for veracity.
- Let's get feedback from some security experts if we're unsure. 

We should really believe in the things we add as action items.

architecture/2024-11-ucp-component-threat-model.md
ytimocin: Removed this one since I couldn't define the threat. A malicious actor needs to have access to UCP to be able to make it do things. But if UCP is tampered with, then we miss one of our assumptions.

architecture/2024-11-ucp-component-threat-model.md
ytimocin: Removed this one also since it is not really a security issue. Also removed the action item. We can revisit if we get another feedback.

rynowak (approved):

PR #74, opened 2024-11-04
Feature Spec- Authoring a User-Defined-Type schema

Reshrahim:
This is a child design document of UDT covers the authoring experience of UDT schema

### DIFF
--- a/architecture/2024-10-user-defined-types-authoring-experience-feature-spec.md
+++ b/architecture/2024-10-user-defined-types-authoring-experience-feature-spec.md
@@ -19,7 +19,7 @@ Below goals are out of scope for the first iteration but might be considered in 
 
 - Advanced scenarios defining the capabilities of the resource type in the schema definition. This includes defining capabilities like connections to other resource, output resources or infrastructure resources produced by the resource type etc.
 - Defining child resources or nested resources in the schema definition
-- Adding a new resource methods or operations to the resource type
+- Adding new resource methods or operations to the resource type
 - Providing a full-fledged TypeSpec tooling experience for authoring the schema definition with autocompletion, error validation at development time.
 - Providing solutions to handle breaking changes on API versions of the resource type
 
@@ -68,6 +68,7 @@ As called out in the [feature spec](/architecture/2024-06-resource-extensibility
     | Resource type description | The description of the resource type | No |
     | Properties | The data model of the resource type | Yes |
     | API Version | The version of the resource type | Yes |
+    | Capabilities | The capabilities of the resource type | No |
 
     Deb creates a plaid.yaml file with the following schema definition
 
@@ -93,7 +94,8 @@ As called out in the [feature spec](/architecture/2024-06-resource-extensibility
                             connectionString:
                                 type: 'string'
                                 description: 'Connection string to the messaging service' 
-                                required: true             
+                                required: true 
+                    capabilities: ["Recipe"]           
     ....
     ```
 
@@ -158,7 +160,8 @@ resource namespace: 'Mycompany.Messaging'
                         logAnalytics:
                             type: 'string'
                             description: 'Usage type of the messaging service'  
-                            default: 'off'        
+                            default: 'off' 
+                capabilities: ["Recipe"]        
 ```
 
 Deb creates the updated resource type schema via the CLI
@@ -204,7 +207,8 @@ resource namespace: 'Mycompany.Messaging'
                             required: true             
                         messageForwarding:
                             type: 'string'
-                            description: 'Usage type of the messaging service'          
+                            description: 'Usage type of the messaging service'
+                capabilities: ["Recipe"]           
 ```
 
 Deb creates the updated resource type schema via the CLI
@@ -214,7 +218,7 @@ rad resource-type create -f plaid.yaml
 Error: Unable to create the resource-type `plaid` for version `2024-10-01`. The schema definitions has a breaking change with the required property `instance_type`. Please delete older versions of the resource type and try again.
 ```
 
-FAQs:
+**FAQs:**
 
 1. How does Radius handle breaking changes in the schema definition ?
     For MVP, we flag breaking changes to the users and provide guidance or documentation to handle the breaking changes. In future iterations, we can provide tooling support to handle breaking changes in the schema definition.
--- a/architecture/2024-10-user-defined-types-authoring-experience-feature-spec.md
+++ b/architecture/2024-10-user-defined-types-authoring-experience-feature-spec.md
@@ -4,7 +4,7 @@
 
 ## Summary
 <!-- A paragraph or two to summarize the topic area. Just define it in summary form so we all know what it is. -->
-User-Defined Types enable Platform engineers to define and deploy their organization's custom services in Radius. More details around the feature, user scenarios and experience can be found [here](/architecture/2024-06-resource-extensibility-feature-spec.md). This document will focus on the authoring experience for defining the user-defined resource type schema.
+User-Defined Types enable Platform engineers to define and deploy their organization's custom services in Radius. More details around the feature, user scenarios and end-end-experience can be found [here](/architecture/2024-06-resource-extensibility-feature-spec.md). This feature specification document details on the requirements and the user experience for authoring and updating the schema for the User-Defined-Type.
 
 ### Top level goals
 <!-- At the most basic level, what are we trying to accomplish? -->
@@ -21,7 +21,7 @@ These goals are out of scope for the first iteration but might be considered in 
 - Advanced scenarios defining the capabilities of the resource type in the schema definition. This includes defining capabilities like connections to other resource, output resources or infrastructure resources produced by the resource type etc.
 - Defining child resources or nested resources in the schema definition
 - Extensibility on CRUDL operations for the resource type
-- Providing solutions to handle breaking changes on API versions of the resource type
+- Providing solutions to handle breaking changes on API versions of the resource type (?)
 
 ## User profile and challenges
 <!-- Define the primary user and the key problem / pain point we intend to address for that user. If there are multiple users or primary and secondary users, call them out.   -->
@@ -155,7 +155,7 @@ Resource type `plaid` created successfully for version `2024-11-01`
 Question 
 
 1. We leave it up to the user to decide whether to use the new optional property in their application definitions and Recipes. Should we provide any guidance?
-2. If Deb tries to add an optional property within the same version or without a default value, what should be the behavior ? Should we consider any change warrants a api version upgrade like Azure/ ARM does?
+2. If Deb tries to add an optional property within the same version or without a default value, what should be the behavior ? Should we consider that any change warrants a api version upgrade like Azure/ ARM does?
 
 #### Sub-scenario 2: Deb adds a required property and updates the resource type Plaid.
 Deb wants to add a required property `messageForwarding`. This is considered to be a breaking change as it would require the Infrastructure operator teams and developers to update their Recipes and application definitions to use the new property in the resource type Plaid. Deb should be provided with an error message when a breaking change is detected in the schema definition.
--- a/architecture/2024-10-user-defined-types-authoring-experience-feature-spec.md
+++ b/architecture/2024-10-user-defined-types-authoring-experience-feature-spec.md
@@ -176,7 +176,6 @@ Any change to the resource-type schema warrants the API version to be upgraded. 
 1. We leave it up to the user to decide whether to use the new optional property in their application definitions and Recipes. Should we provide any guidance/feature from GitOps perspective?
 <>
 
-
 #### Sub-scenario 2: Deb adds a required property and updates the resource type Plaid.
 Deb wants to add a required property `messageForwarding`. This is considered to be a breaking change as it would require the Infrastructure operator teams and developers to update their Recipes and application definitions to use the new property in the resource type Plaid. Deb should be provided with an error message when a breaking change is detected in the schema definition.
 
--- a/architecture/2024-10-user-defined-types-authoring-experience-feature-spec.md
+++ b/architecture/2024-10-user-defined-types-authoring-experience-feature-spec.md
@@ -120,7 +120,7 @@ Updates to schema definitions can be classified into two categories: non-breakin
 There are two key sub-scenarios to consider when Deb updates the schema for the internal messaging service Plaid
 
 #### Sub-scenario 1: Deb adds a optional property and updates the resource type Plaid.
-Deb wants to add an optional property `logAnalytics` to know whether the resource type should have log analytics enabled. This is considered to be a non-breaking change as it is an optional property with default value. Users should be able to use the new property without any impact across versions on their existing applications.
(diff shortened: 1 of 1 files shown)


architecture/2024-10-user-defined-types-authoring-experience-feature-spec.md
```diff
@@ -0,0 +1,230 @@
...
+
+| Change type | Description | Example |
+|-------------|-------------|---------|
+| Non-breaking changes | Changes that can be made to the schema definition without impacting existing applications that use the resource type. | Eg: Adding an optional property to the schema definition |
+| Breaking changes | Changes that are not backward/forward compatible. These changes are disruptive and often requires the users to update their applications to use the new schema definition. | Eg: Adding a required property to the schema definition |
+
+There are two key sub-scenarios to consider when Deb updates the schema for the internal messaging service Plaid
+
+#### Sub-scenario 1: Deb adds a optional property and updates the resource type Plaid.
+Deb wants to add an optional property `logAnalytics` to know whether the resource type should have log analytics enabled. This is considered to be a non-breaking change as it is an optional property with default value. Users should be able to use the new property without any impact across versions on their existing applications.
```
Reshrahim, line 123:
```suggestion
Deb wants to add an optional property `logAnalytics` to have log analytics enabled for the resource-type. This is considered to be a non-breaking change as it is an optional property with default value. Users should be able to use the new property without any impact across versions on their existing applications.
```

architecture/2024-10-user-defined-types-authoring-experience-feature-spec.md
```diff
@@ -0,0 +1,230 @@
...
+    | API Version | The version of the resource type | Yes |
+
+    Deb creates a plaid.yaml file with the following schema definition
+
+    ```yaml
+    namespace: 'Mycompany.Messaging'
+    resourceTypes:
+        'Mycompany.Messaging/plaidResource':
+            name: 'Name of the resource'
+            version: '2024-10-01'
```
rynowak, line 75: There are three *big* concepts for us to work with, and they are all nested.

- Resource provider
  - Resource type
    -  API version  

---

The simplest file structure is to define one resource provider per-file. Then you'd end up with a structure like:

```yaml
# Resource provider 
name: 'MyCompany.Messaging`
... # other resource provider settings go here

resourceTypes:
  # Resource type
  'plaidResource': 
     ... # other resource type settings go here

    apiVersions:
    # API version
      '2024-01-01': 
        schema: ...
        capabilities: ...
        ... # other api version settings go here
```

When you nest these concepts in YAML, they can either be map-like or array-like. I tend to prefer map-like because there's less ambiguity. (Your example is both, and it's redundant 😆).

Example map-like:

```yaml
# Resource provider 
resourceTypes:
  'plaidResource': 
    apiVersions:
      '2024-01-01': 
        schema: ...
        capabilities: ...
```

Example array-like:


```yaml
# Resource provider 
resourceTypes:
  - name: 'plaidResource'
    apiVersions:
      - name: '2024-01-01'
        schema: ...
        capabilities: ...
```

architecture/2024-10-user-defined-types-authoring-experience-feature-spec.md
```diff
@@ -0,0 +1,230 @@
...
+* **Author**: Reshma Abdul Rahim (@reshrahim)
+
+## Summary
+<!-- A paragraph or two to summarize the topic area. Just define it in summary form so we all know what it is. -->
+User-Defined Types enable Platform engineers to define and deploy their organization's custom services in Radius. More details around the feature, user scenarios and end-end-experience can be found [here](/architecture/2024-06-resource-extensibility-feature-spec.md). This feature specification document details on the requirements and the user experience for authoring and updating the schema for the User-Defined-Type.
+
+### Top level goals
+<!-- At the most basic level, what are we trying to accomplish? -->
+
+- Enable a simple and intuitive experience for platform engineers to define the schema for their User-Defined-Type with minimal lines of code
```
rynowak, line 12: Concept count is better than lines of code. There's all kinds of wierd games to play if LOC is the thing we're optimizing.

architecture/2024-10-user-defined-types-authoring-experience-feature-spec.md
```diff
@@ -0,0 +1,230 @@
...
+- Ensure that there is low concept count for the user to understand and get started with defining the schema for their User-Defined-Type
+- Ensure that users have clear visibility into the validation process of the schema definition to guarantee its correctness.
```
rynowak, line 14: Can we uplevel this? I don't think this is about users understanding how validation is implemented.

architecture/2024-10-user-defined-types-authoring-experience-feature-spec.md
```diff
@@ -0,0 +1,230 @@
...
+
+### Non-goals (out of scope)
+<!-- What are we explicitly not trying to accomplish? -->
+These goals are out of scope for the first iteration but might be considered in future iterations based on user feedback.
+
+- Providing a full-fledged TypeSpec tooling experience for authoring the schema definition with autocompletion, error validation at development time.
+- Advanced scenarios defining the capabilities of the resource type in the schema definition. This includes defining capabilities like connections to other resource, output resources or infrastructure resources produced by the resource type etc.
+- Defining child resources or nested resources in the schema definition
+- Extensibility on CRUDL operations for the resource type
+- Providing solutions to handle breaking changes on API versions of the resource type (?)
```
rynowak, line 24: Can we highlight which of these are out of scope forever vs out of scope for the first cut? I think this could send the message that some of these ideas aren't important, and they obviously are.

architecture/2024-10-user-defined-types-authoring-experience-feature-spec.md
```diff
@@ -0,0 +1,230 @@
...
+                        description: 'Connection string to the messaging service'
+    ....
+    ```
+ Questions
+1. Is resource namespace needed or can we parse this from the resource type ?
+
+ Now Deb can register this resource type via the CLI
+
+    ```bash
+    rad resource-type create -f plaid.yaml
```
rynowak, line 98: Is the file meant to container a single resource type or all types for a resource provider? It looks like it's doing both.

architecture/2024-10-user-defined-types-authoring-experience-feature-spec.md
```diff
@@ -0,0 +1,230 @@
...
+    ```
+The boilerplate code for CRUDL operations for the resource type is generated automatically and the resource type is registered in Radius.
+
+Now on creation, the schema should be validated for correctness and completeness. Deb should be provided with error messages if the schema is incorrect or incomplete.
```
rynowak, line 102: Do we want to specify the validations we're going to perform right now?

architecture/2024-10-user-defined-types-authoring-experience-feature-spec.md
```diff
@@ -0,0 +1,230 @@
...
+
+```bash
+rad resource-type create -f plaid.yaml
+Resource type `plaid` created successfully for version `2024-11-01`
+```
+
+Question 
+
+1. We leave it up to the user to decide whether to use the new optional property in their application definitions and Recipes. Should we provide any guidance?
+2. If Deb tries to add an optional property within the same version or without a default value, what should be the behavior ? Should we consider that any change warrants a api version upgrade like Azure/ ARM does?
```
rynowak, line 158:
```suggestion
2. If Deb tries to add an optional property within the same version or without a default value, what should be the behavior ? Should we consider that any change warrants a api version upgrade like Azure/ ARM does?
```

yes

architecture/2024-10-user-defined-types-authoring-experience-feature-spec.md
```diff
@@ -0,0 +1,230 @@
...
+Deb creates the updated resource type schema via the CLI
+
+```bash
+rad resource-type create -f plaid.yaml
+Resource type `plaid` created successfully for version `2024-11-01`
+```
+
+Question 
+
+1. We leave it up to the user to decide whether to use the new optional property in their application definitions and Recipes. Should we provide any guidance?
```
rynowak, line 157: I don't understand the question. Optional is optional.

architecture/2024-10-user-defined-types-authoring-experience-feature-spec.md
```diff
@@ -0,0 +1,230 @@
...
+                messageForwarding:
+                    type: 'string'
+                    description: 'Usage type of the messaging service'          
+```
+
+Deb creates the updated resource type schema via the CLI
+
+```bash
+rad resource-type create -f plaid.yaml
+Error: Unable to create the resource-type `plaid` for version `2024-10-01`. The schema definitions has a breaking change with the required property `instance_type`. Please delete older versions of the resource type and try again.
```
rynowak, line 192: What's the source of truth for this? What are we comparing?

architecture/2024-10-user-defined-types-authoring-experience-feature-spec.md
```diff
@@ -0,0 +1,230 @@
...
+```
+
+Questions
+1. For MVP, can we just flag breaking changes and provide documentation on how to handle breaking changes ?
```
rynowak, line 196: Either breaking changes are explicitly supported, or explicitly blocked. If they are supported, we have to define what the behavior is. 

I'd lean towards explicitly blocked to start, and we can add ways to make them explicitly supported in the future.

architecture/2024-10-user-defined-types-authoring-experience-feature-spec.md
```diff
@@ -0,0 +1,230 @@
...
+Error: Unable to create the resource-type `plaid` for version `2024-10-01`. The schema definitions has a breaking change with the required property `instance_type`. Please delete older versions of the resource type and try again.
+```
+
+Questions
+1. For MVP, can we just flag breaking changes and provide documentation on how to handle breaking changes ? 
+2. What are the breaking changes that we need to consider ? 
+    - Adding a required property
+    - Removing a required property
+    - Changing the type of a property
+    - Changing the name of a property
```
rynowak, line 201: Trying to precisely define each type of breaking change is a task for writing the documentation. We should define it in-terms of assignability.

- Valid data from an older API version can be *assigned* to a new version and still be valid. 
- Valid data from an new API version can be *assigned* to an old version and still be valid. 

Fields may be dropped during the conversion process while moving from new -> old.

architecture/2024-10-user-defined-types-authoring-experience-feature-spec.md
Reshrahim: Thanks for catching that. I have updated the format to be more map-like.

architecture/2024-10-user-defined-types-authoring-experience-feature-spec.md
Reshrahim: I have separated extensibility on CRUDL operations as a goal out of scope forever.

architecture/2024-10-user-defined-types-authoring-experience-feature-spec.md
Reshrahim: I intended to do it RP level. updated the file structure and creation

architecture/2024-10-user-defined-types-authoring-experience-feature-spec.md
Reshrahim: Sorry the question is incomplete. I was thinking about guidance/feature from GitOps perspective. Is that something we should provide or think about ? How Gitops tools understand changes to the Radius resource-type schemas ?

architecture/2024-10-user-defined-types-authoring-experience-feature-spec.md
Reshrahim: I assumed we have the schema with the version stored internally for a resource type. Can't we compare the new schema with that?

architecture/2024-10-user-defined-types-authoring-experience-feature-spec.md
Reshrahim: I added a basic list of overall error validation themes we need to do. Anything else we want to add?

architecture/2024-10-user-defined-types-authoring-experience-feature-spec.md
rynowak: > version stored internally

I think we should probably do two things:

- Compare **locally** where possible. eg: all of the API versions defined in a single file can be checked. We should do this because its fast to implement and provide the fastest possible feedback to the user.
- Compare **remotely** when data is missing to prevent data loss. eg: if you deleted an API version from your local file, but it still exists on the server. We should do this to prevent data loss.
- Compare **remotely** when an API version is being updated in place. Same reason as above.

architecture/2024-10-user-defined-types-authoring-experience-feature-spec.md
```diff
@@ -0,0 +1,259 @@
...
+|-------------|-------------|---------|
+| Non-breaking changes | Changes that can be made to the schema definition without impacting existing applications that use the resource type. | Eg: Adding an optional property to the schema definition |
+| Breaking changes | Changes that are not backward/forward compatible. These changes are disruptive and often requires the users to update their applications to use the new schema definition. | Eg: Adding a required property to the schema definition |
+
+There are two key sub-scenarios to consider when Deb updates the schema for the internal messaging service Plaid
+
+#### Sub-scenario 1: Deb adds a optional property and updates the resource type Plaid.
+Deb wants to add an optional property `logAnalytics` to have log analytics enabled for the resource-type. This is considered to be a non-breaking change as it is an optional property with default value. Users should be able to use the new property without any impact across versions on their existing applications.
+
+```yaml
```
rynowak, line 137: A few comments on this, maybe we should huddle and close down the following.

`resource namespace` isn't a great property name, because it requires quotes. I'd suggest either `namespace` or `name`.

---

See `resource namespace` again. A property can't both have a string value and children. We need to nest the types under another property. Open to suggestions on the naming, I like being concise and saying `types`.

This needs to look like:

```yaml
namespace: 'MyCompany.Messaging`
types:
  'plaidResources':
    apiVersions: ...
```

---

`required: true` isn't how required properties work in OpenAPI. It's more complicated 😆 

Here's the spec: https://json-schema.org/draft/2020-12/json-schema-validation#name-required (have fun)

Here's an example:

```yaml
schema:
  type: 'object'
  properties:
    a:
      type: string 
    b:
      type: string 
    c:
      type: string 
  required: [ 'a', 'b' ]
```

Note that `required` is an array of property names, and is a sibling of `properties`.

There's a HUGE amount of complexity in OpenAPI/JSON Schema. We'll support a **restricted** subset of what's possible.  Do you want me to take a stab at **that** document?

Reshrahim: I will update the eg with names and types as the properties. It would be helpful if you can help with the restricted subset of whats possible? It could be something small we start with and expand as we identify requirements based on usage/adoption

rynowak: I can help with that .... after Kubecon 😆

PR #76, opened 2024-11-25
Authorization feature spec

zachcasper:


architecture/2024-11-authz-feature-spec.md
```diff
@@ -0,0 +1,392 @@
...
+
+Radius is inherently a multi-tenant system with several different personas. Separation of concerns between developers and platform engineers is a core concept. Developers model their applications using resource types abstracted away from the infrastructure while platform engineers configure how those resource types are deployed. Radius is also designed to support multiple different applications, developers, and groups of developers. 
+
+The building blocks of this multi-tenancy is implemented today with resource groups. Every resource, whether it is an application, an environment, or a recipe, resides in a resource group. However, today, access to Radius is binary—you ether have access to the API or you do not. It is not possible to control permissions to these resources and resource groups. This feature specification describes the requirements and user experience for authorizing users to perform a granular set of actions on a granular set of resources and resource groups.
+
+### Goals
+
+This feature specification describes:
+
+* Granting and revoking permission to users to perform Radius administrative actions at the control-plane level and tenant level
```
Reshrahim, line 15: Permission to install, delete and update Radius installation?

architecture/2024-11-authz-feature-spec.md
```diff
@@ -0,0 +1,392 @@
...
+* Granting and revoking permission to users to perform create, read, update, delete actions on resources and resource groups including applications, user-defined resource types, environments, and recipe registrations
+* Granting and revoking permission to users to associate resources between resource groups
+
+### Non-goals (out of scope)
+
+This feature specification only addresses authorization. A separate feature specification is under development for authenticating users.
```
willtsai, line 21: Would multi-tenancy support for Radius overall be a part of the scope for this?

architecture/2024-11-authz-feature-spec.md
```diff
@@ -0,0 +1,392 @@
...
+
+## Summary
+
+Radius is inherently a multi-tenant system with several different personas. Separation of concerns between developers and platform engineers is a core concept. Developers model their applications using resource types abstracted away from the infrastructure while platform engineers configure how those resource types are deployed. Radius is also designed to support multiple different applications, developers, and groups of developers. 
+
+The building blocks of this multi-tenancy is implemented today with resource groups. Every resource, whether it is an application, an environment, or a recipe, resides in a resource group. However, today, access to Radius is binary—you ether have access to the API or you do not. It is not possible to control permissions to these resources and resource groups. This feature specification describes the requirements and user experience for authorizing users to perform a granular set of actions on a granular set of resources and resource groups.
+
+### Goals
+
+This feature specification describes:
```
Reshrahim, line 13: Shd we have also have a deploy to permission to deploy an application to an environment ?

architecture/2024-11-authz-feature-spec.md
```diff
@@ -0,0 +1,392 @@
...
+
+* Granting and revoking permission to users to perform Radius administrative actions at the control-plane level and tenant level
+* Granting and revoking permission to users to perform create, read, update, delete actions on resources and resource groups including applications, user-defined resource types, environments, and recipe registrations
```
willtsai, line 16: Is it in scope for "intra role" read/write granularity? e.g. a user may be able to read the contents of an environment but not have write access to modify or delete it

architecture/2024-11-authz-feature-spec.md
```diff
@@ -0,0 +1,392 @@
...
+### Scenario 1 – Adding Radius administrators
+
+As a Radius administrator, I just installed Radius on a Kubernetes cluster and now I need to add an additional user as a Radius administrator.
+
+**User Experience**
+
+```bash
+# Radius administrative access is granted via membership in the radius-admin role in the radius-system namespace
+# Add user-1 to the radius-admin role
+kubectl create rolebinding user-1-admin --role radius-admin --user user-1 -n radius-system
```
Reshrahim, line 52: Given our control plane might be runtime agnostic in future, this might be a command we support in Radius?

architecture/2024-11-authz-feature-spec.md
```diff
@@ -0,0 +1,392 @@
...
+**Resource type administrators** – A resource type administrator needs to maintain the resource types which are available in the Radius system and manage who can use those resource types within their applications.
+
+**Recipe administrators** – A recipe administrator needs to manage the set of recipes registered to an environment. They must author, test, and maintain infrastructure as code modules such as Terraform or Bicep to implement the recipe. In many organizations, a recipe administrator and an environment administrator will be the same person. However, in other organizations, administration of recipes may be delegated to specialty teams for certain resource types; e.g., DBAs may maintain recipes for database resource types.
+
+## Scenarios
+
+The following scenarios will describe a final setup similar to the image below.
+
+![](2024-11-authz-feature-spec/authz.png)
+
```
Reshrahim, line 42: A summary of the user, roles, the permissions we are thinking and what it allows the user to do would be helpful

architecture/2024-11-authz-feature-spec.md
willtsai, line 16: I'm assuming the read permissions would apply to all read scenarios? e.g. `show`, `list`, `app graph`, dashboard visualizations, etc.

architecture/2024-11-authz-feature-spec.md
willtsai, line 52: what about for non-kubernetes environments?

willtsai (commented):
Overall this is a great start, many details that nicely clarify things. I suggest adding a table to illustrate the matrix of roles and permissions you propose to put in place to help summarize things a bit.

architecture/2024-11-authz-feature-spec.md
```diff
@@ -0,0 +1,392 @@
...
+As a Radius administrator, I need to create a resource group for applications and application resources to be created in as well as set permissions which define who can create applications in that group.
+
+**User Experience**
+
+```bash
+# Create application resource group 
+rad group create rg-app-1
+
+# Create developer role
+rad role definition create -f developer-role-definition.yaml
```
Reshrahim, line 70: Shouls this just be `rad role create` and you provide the definition either via file or flags

architecture/2024-11-authz-feature-spec.md
zachcasper: No, these are external to Radius actions. In these cases, you would need access to the Kubernetes cluster.

architecture/2024-11-authz-feature-spec.md
zachcasper: Only in so far as authorization is foundational functionality for multi-tenancy. I'm sure there are several other multi-tenancy things we need to think about. But if we had AuthZ and the ability to deploy to external Kubernetes clusters, we're very very close to being multi-tenant.

architecture/2024-11-authz-feature-spec.md
zachcasper: When Radius runs on a different platform, there will be a different way of managing the admin user. Admin or root is always platform dependent. The key point is that it's only for admin not for other roles.

zachcasper: See previous comment.

architecture/2024-11-authz-feature-spec.md
zachcasper: It's modeled after [az role definition create](https://learn.microsoft.com/en-us/cli/azure/role/definition) and [az role assignment create](https://learn.microsoft.com/en-us/cli/azure/role/assignment)

architecture/2024-11-authz-feature-spec.md
```diff
@@ -0,0 +1,390 @@
...
+As a platform engineer, I just installed Radius using my Kubernetes cluster credentials. I want to access Radius via my typical user credentials from my IdP.
+
+**User Experience**
+
+```bash
+# Assign the Radius Administrator role to a user from the IdP
+rad role assignment create \
+  --assignee me@my-company.net \
+  --role radius-admin \
+  --scope /planes/radius/*
```
brooke-hamilton, line 121: Can scopes be other things besides resource groups? If a scope is an individual resource, then does a role still make sense? For example, would we want to assign `radius-admin` to a single resource or resource group?

architecture/2024-11-authz-feature-spec.md
```diff
@@ -0,0 +1,390 @@
...
+rad role assignment create \
+  --assignee developer@my-company.net \
+  --role deployer \
+  --scope /planes/radius/MyCompany/resourceGroup/default-environment
+```
+
+**Result**
+
+1. `developer@my-company.net` is added to the out-of-the-box role `developer` for the `default-application` resource group 
+2. `developer@my-company.net` is added to the out-of-the-box role `deplorer` for the `default-environment` resource group
```
lakshmimsft, line 167: nit: typo 'developer`

architecture/2024-11-authz-feature-spec.md
```diff
@@ -0,0 +1,390 @@
...
+
+**Exceptions**
+
+The operation fails and informs the user interactively if:
+
+* The current user is not a member of the `radius-admin` or `group-admin` roles or the `cluster-admin` RoleBinding in the Kubernetes cluster
+* The resource group already exists
+* The role assignment already exists
+
+#### User story 4 – Create resource group for an application
```
brooke-hamilton, line 212: Can you give more detail on how we would grant or revoke the ability to use a resource type namespace?

architecture/2024-11-authz-feature-spec.md
zachcasper: Ha they deplore environments!

architecture/2024-11-authz-feature-spec.md
```diff
@@ -0,0 +1,390 @@
...
+ **User Experience**
+
+When Radius is installed, two resource groups are created:
+
+1. `default-application` – When a user deploys an application without specifying a resource group, it is deployed to this resource group.
+2. `default-environment` – When a user creates an environment without specifying a resource group, it is deployed to this resource group.
+
+> [!NOTE]
+>
+> The default resource group is split to encourage best practices for actual Radius use. Splitting resource groups introduces the concept of resource groups and Radius' authorization model.
```
lakshmimsft, line 148: Not sure I understand fully why the split at the resourcegroup.

architecture/2024-11-authz-feature-spec.md
```diff
@@ -0,0 +1,390 @@
...
+
+1. An empty resource group is created called `non-prod-env`
+2. The `env-admin` role is granted to the cloud engineering group to manage environments within the `non-prod-env` resource group
+
+**Exceptions**
+
+The operation fails and informs the user interactively if:
+
+* The current user is not a member of the `radius-admin` or `group-admin` roles or the `cluster-admin` RoleBinding in the Kubernetes cluster
+* The resource group already exists
```
rynowak, line 209: Should this say `The resource group does not exist`?

architecture/2024-11-authz-feature-spec.md
```diff
@@ -0,0 +1,390 @@
...
+* The role assignment already exists
```
rynowak, line 210: In general our CLI `create` commands always mean `create or update` and print a message indicating which one it was. 

Likewise `delete` means "make it gone".

The reason is for scriptability. It's really hard to script script that that's state dependent.

architecture/2024-11-authz-feature-spec.md
```diff
@@ -0,0 +1,390 @@
...
+# Update the Developer role
+rad role definition update -f developer-role-definition.yaml
+```
+
+The contents of `developer-role-definition.yaml` is similar to:
+
+```yaml
+---
+name: developer
+description: Role to manage applications and application resources within a resource group
```
rynowak, line 230: This should look like a UCP resource. 

```yaml
name: developer
properties: 
  description: Role to manage applications and application resources within a resource group
Comment
  actions:
  # Grant permissions to CRUDL applications in the scoped resource group
    - Applications.Core/applications/*
  # Restrict using any resource type except those in the MyCompany.App namespace
    - MyCompany.App/*
```

architecture/2024-11-authz-feature-spec.md
```diff
@@ -0,0 +1,390 @@
...
+actions:
+  # Grant permissions to CRUDL applications in the scoped resource group
+  - Applications.Core/applications/*
+  # Restrict using any resource type except those in the MyCompany.App namespace
+  - MyCompany.App/*
```
rynowak, line 235: Just thinking aloud, but it would be really fun to demo this with a `rad role-definition edit` (like `kubectl edit`).

I think that creates a really strong narrative about customization.

architecture/2024-11-authz-feature-spec.md
```diff
@@ -0,0 +1,390 @@
...
+## Feature Summary
+
+| Priority | Size | Feature                                                      |
+| -------- | ---- | ------------------------------------------------------------ |
+| p0       | S    | Out-of-the-box role definitions                              |
+| p0       | L    | UCP validates membership in either a role definition or the Kubernetes cluster-admin RoleBinding on each API call |
+| p0       | S    | Create or delete a role assignment for an individual user from the configured IdP system for a resource group (`rad role assignment create --assignee user@mycompany.net`) |
+| p1       | S    | Create or delete a role assignment for a group from the configured IdP system for a resource group (`rad role assignment create --assignee my-group@mycompany.net`) |
+| p1       | M    | Ability to create, update, and delete a role definition      |
+| p2       | M    | Replace the `default` resource group with `default-application` and `default-environment` and modify `rad environment`, `rad deploy`, and `rad run` to use these new groups |
```
rynowak, line 382: I really want to dig into this more.

architecture/2024-11-authz-feature-spec.md
```diff
@@ -0,0 +1,390 @@
+# Radius Authorization Feature Specification
+
+* **Author**: Zach Casper (@zachcasper)
+
+## Summary
+
+Radius is inherently a multi-tenant system with several different personas. Separation of concerns between developers and platform engineers is a core concept. Developers model their applications using resource types abstracted away from the infrastructure while platform engineers configure how those resource types are deployed. Radius is also designed to support multiple different applications, developers, and groups of developers. 
+
+The building blocks of this multi-tenancy is implemented today with resource groups. Every resource, whether it is an application, an environment, or a recipe, resides in a resource group. However, today, access to Radius is binary—you ether have access to the API or you do not. It is not possible to control permissions to these resources and resource groups. This feature specification describes the requirements and user experience for authorizing users to perform a granular set of actions on a granular set of resources and resource groups.
```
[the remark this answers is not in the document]
zachcasper: >I also think about the term multi-tenancy as an organizational separation rather than a team/role separation.
It's an overloaded term that refers to both in different contexts. I've clarified that in the text.

architecture/2024-11-authz-feature-spec.md
```diff
@@ -0,0 +1,390 @@
...
+
+**Identity and Access Management** – An overloaded term which can refer to the overall topic of authentication and authorization, modules within a system which implement both authentication and authorization, or product names such as AWS IAM
+
+**Role** – A collection of permitted actions; typically named for the user persona
+
+**Action** – An act performed upon a resource such as create, read (or show), update, delete, list, or a resource-specific action such as restart; typically is mapped one to one with an API call or a CLI command
+
+**Scope** – The set of resources which actions are allowed to be performed; within Radius, the scope is typically a resource group
+
+**Role Definition** – A commonly used pattern in systems which includes the role name and a collection of allowed actions; specific implementations include Kubernetes Roles, AWS IAM roles, Azure role definitions, or Google Cloud custom roles
```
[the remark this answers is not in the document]
zachcasper: ~~**Role** – A collection of permitted actions; typically named for the user persona~~
**Role** – An abstract term which describes the permitted actions of a user persona

architecture/2024-11-authz-feature-spec.md
```diff
@@ -0,0 +1,390 @@
...
+
+**Role Assignment** – A commonly used pattern in systems which maps a role definition to a user or set of users; specific implementations include Kubernetes RoleBinding, Azure role assignments, or Google Cloud role binding
+
+## User profiles and challenges
+
+**Platform engineers** – Platform engineers configure and maintain Radius. These users are primarily concerned with configuring Radius to support the various development and infrastructure teams within their organization. They need a flexible authorization scheme built into Radius which is easy to setup in the beginning, but flexible enough to grow as Radius is used by more teams in more ways. This flexibility will enable platform engineers to grant and restrict actions at a high level on a wide range of resources, or on a fine-grained level.
```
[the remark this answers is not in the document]
zachcasper: The operator persona is simply missing here. Added
>**Operators** – Operators, sometimes referred to as Site Reliability Engineers (SREs) ensure the availability, performance, and security applications running in production. In some cases, the operator persona is combined with the developer and referred to as DevOps. In other cases, the security responsibility is delegated to a dedicated security operator (SecOps).

architecture/2024-11-authz-feature-spec.md
```diff
@@ -0,0 +1,390 @@
...
+
+**Deployer** – The Deployer role is for the same persona as the Developer role. The role exists to control which environments the developer can deploy applications to. The Developer role and Deployer role are typically assigned to the same user at the same time, just with different scopes.
+
+**Environment Administrator** – The Environment Administrator role grants permission to manage environments and connections to cloud platforms. 
+
+**Resource Type Administrator** – The Resource Type Administrator role grants permission to manage the resource types which are available in the Radius tenant. The Radius Administrator or Resource Group Administrator can further control who can use those resource types within their applications.
+
+**Recipe Administrator** – The Recipe Administrator role grants permission to register and unregistered recipes in an environment. 
+
```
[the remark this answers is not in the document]
zachcasper: They are not hierarchical. This is just the out of the box configuration. Any role definition could be deleted without affecting the others (except radius-admin).

architecture/2024-11-authz-feature-spec.md
zachcasper: Yes, scopes can be other things besides a resource group. For example, the resource type admin are alllowed the `System.Resources/resourceproviders/MyCompany.App/*` actions on `/planes/radius/MyCompany/ResourceTypes`.

architecture/2024-11-authz-feature-spec.md
```diff
@@ -0,0 +1,390 @@
...
+rad role assignment create \
+  --assignee me@my-company.net \
+  --role radius-admin \
+  --scope /planes/radius/*
+```
+
+**Result**
+
+1. `me@my-company.net` is added to the out-of-the-box role `radius-admin` 
+2. The user can now authenticate to Radius using the IdP and continue administering Radius with the users standard credentials
```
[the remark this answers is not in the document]
zachcasper: I was not imagining a switch from Kubernetes auth to IdP for now. When we host Radius outside of Kubernetes obviously the behavior would change. 

The way I think about it is that a Kubernetes cluster admin would always have Radius admin privileges. Cluster admin would be analogous to having root on a server. If an IdP were to be unavailable, the cluster admin would always have access. 

I do not have strong opinions on how this should work. I view the Kubernetes authZ as a stepping stone to IdP.

architecture/2024-11-authz-feature-spec.md
```diff
@@ -0,0 +1,390 @@
...
+
+#### User story 2 – Assigning the Developer role to the default application resource group
+
+As a platform engineer, I need to assign the Developer role to a developer for the first time. There are no applications created in Radius yet and I am not familiar with resource groups yet.
+
+ **User Experience**
+
+When Radius is installed, two resource groups are created:
+
+1. `default-application` – When a user deploys an application without specifying a resource group, it is deployed to this resource group.
```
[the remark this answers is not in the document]
zachcasper: My objective was describing as-is functionality with today's `default` group. I'm assuming a more precise restatement of what you're saying is that **_All applications created within a resource group must have a unique name_**.

architecture/2024-11-authz-feature-spec.md
zachcasper: The pattern we are encouraging is applications are held in one set of resource groups while environments are held in a separate set of resource groups.

architecture/2024-11-authz-feature-spec.md
```diff
@@ -0,0 +1,390 @@
...
+**Result**
+
+1. `developer@my-company.net` is added to the out-of-the-box role `developer` for the `default-application` resource group 
+2. `developer@my-company.net` is added to the out-of-the-box role `deplorer` for the `default-environment` resource group
+
+**Exceptions**
+
+The operation fails and informs the user interactively if:
+
+* The current user is not a member of the `radius-admin` or `group-admin` roles or the `cluster-admin` RoleBinding in the Kubernetes cluster
```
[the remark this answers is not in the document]
zachcasper: >~~The current user is not a member of the `radius-admin` or `group-admin` roles or the `cluster-admin` RoleBinding in the Kubernetes cluster~~

The current user is not a member of the `radius-admin` or `group-admin` roles

Also added:

> [!NOTE]
>
> Radius always treats a user with a `cluster-admin` Kubernetes RoleBinding on the cluster as `radius-admin`. In a new Radius installation, there is no identity provider (IdP) configured. Therefore the Radius administrator must configure an IdP before any user can access Radius without having `cluster-admin` permissions.

architecture/2024-11-authz-feature-spec.md
zachcasper: Fixed

architecture/2024-11-authz-feature-spec.md
zachcasper: Corrected

architecture/2024-11-authz-feature-spec.md
zachcasper: A user is granted the ability to perform operations (actions in this functional spec) on a set of resource types. For example, to create a container a developer would need Applications.Core/containers/* added to their role definition.  Then they would need that role definition assigned to them within a specific scope (i.e., resource group). 

If an admin wanted to restrict usage of Applications.Core/containers, they could remove that from their role definition and replace it with something like MyCompany.App/* which would be the entire resource catalog for MyCompany.App.

architecture/2024-11-authz-feature-spec.md
```diff
@@ -0,0 +1,390 @@
...
+
+### Scenario 2 – Simple internal developer platform
+
+A simple internal developer platform is a configuration similar to a first-time setup by a platform engineer for their organization's developers. In this scenario, Radius will be used to manage a handful of applications each with their own development team. A single platform engineering team manages all environments, resource types, and recipes. 
+
+For this scenario, the platform engineer will use the authorization features of Radius to control permissions. However, since Radius is new to the organization, the platform engineer expects to use prescriptive guidance from Radius using pre-defined roles and best practices.
+
+Radius will ship with several roles pre-defined. These roles grant permission to perform actions on certain resource types. These roles will not be assigned to any user out of the box except the Radius administrator role. The documentation will describe how to assign a role to a user or group of users and describe how to define the scope of those permissions.
+
```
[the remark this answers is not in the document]
zachcasper: >~~These roles will not be assigned to any user out of the box except the Radius administrator role.~~

> These roles will not be assigned to any user out of the box except the Radius administrator role which is inherently assigned to users with the `cluster-admin` Kubernetes RoleBinding on the cluster. 

This behavior can not be disabled for now.

architecture/2024-11-authz-feature-spec.md
```diff
@@ -0,0 +1,390 @@
...
+#### Out-of-the-box roles
+
+**Radius Administrator** – The Radius Administrator has superuser access to the Radius system.
```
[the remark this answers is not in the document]
zachcasper: | Role Definition                  | Short Name    | Authorizer      | Resources                                            | Allowed Action |
| -------------------------------- | ------------- | --------------- | ---------------------------------------------------- | -------------- |
| Kubernetes Cluster Administrator | cluster-admin | K8s RoleBinding | *                                                    | *              |
| Radius Administrator             | radius-admin  | Configured IdP  | /planes/radius/MyCompany                             | *              |

architecture/2024-11-authz-feature-spec.md
```diff
@@ -0,0 +1,390 @@
...
+
+For this scenario, the platform engineer will use the authorization features of Radius to control permissions. However, since Radius is new to the organization, the platform engineer expects to use prescriptive guidance from Radius using pre-defined roles and best practices.
+
+Radius will ship with several roles pre-defined. These roles grant permission to perform actions on certain resource types. These roles will not be assigned to any user out of the box except the Radius administrator role. The documentation will describe how to assign a role to a user or group of users and describe how to define the scope of those permissions.
+
+#### Out-of-the-box roles
+
+**Radius Administrator** – The Radius Administrator has superuser access to the Radius system. 
+
+**Resource Group Administrator** – The Resource Group Administrator role grants permission to manage resource groups, role definitions, and role assignments.
```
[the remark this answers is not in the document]
zachcasper: I agree. I debated this in my mind as well. It could legitimately be separated. Indeed, a Radius admin could modify these groups to be that way. I ended up defaulting to the simpler option which is to group resource groups and roles together.

architecture/2024-11-authz-feature-spec.md
```diff
@@ -0,0 +1,390 @@
...
+
+**Developer** – The Developer role grants permission to create applications and application resources. It does not allow them to deploy resources to an environment without the associated Deployer role. Out of the box, developers create, update, delete, list, get, and deploy resources from `Application.Core` (except Environment and Extenders), `Applications.Datastores`, `Applications.Messaging`, and `Applications.Dapr`. Environments is reserved for the platform engineering team. Extenders are not included by default to encourage the use of pre-defined resource types. If the platform engineer creates additional resource types, they can add the resource type namespace (aka resource provider name) to this role.
+
+**Deployer** – The Deployer role is for the same persona as the Developer role. The role exists to control which environments the developer can deploy applications to. The Developer role and Deployer role are typically assigned to the same user at the same time, just with different scopes.
```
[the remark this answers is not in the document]
sylvainsf: I believe the story here is Developer A can deploy to Beta, but not Prod. Developer B (some prod SRE person) can deploy to Prod.

architecture/2024-11-authz-feature-spec.md
```diff
@@ -0,0 +1,390 @@
...
+
+**Environment Administrator** – The Environment Administrator role grants permission to manage environments and connections to cloud platforms. 
+
+**Resource Type Administrator** – The Resource Type Administrator role grants permission to manage the resource types which are available in the Radius tenant. The Radius Administrator or Resource Group Administrator can further control who can use those resource types within their applications.
+
+**Recipe Administrator** – The Recipe Administrator role grants permission to register and unregistered recipes in an environment.
```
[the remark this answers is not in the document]
zachcasper: I marked this as a p3 feature. Let's not spend time on it right now until we validate the use cases exists. The idea was that a subset of recipes would be delegated to another team. For example, DBAs would only manage recipes and not environments.

architecture/2024-11-authz-feature-spec.md
```diff
@@ -0,0 +1,390 @@
...
+| Resource Type Administrator  | type-admin   | System.Resources/resourceproviders                   | *              |
+| Recipe Administrator         | recipe-admin | Only the recipe property of environments             | *              |
+
+#### User story 1 – Assigning the Radius Administrator role
+
+> [!NOTE]
+>
+> In a new Radius installation, there is no identity provider (IdP) configured. Radius does not have a built-in identity system, therefore the Radius administrator must configure an IdP before Radius will recognize and authorize the user. Roles can be assigned prior to IdP configuration, but without an IdP there is no effect. 
+
+As a platform engineer, I just installed Radius using my Kubernetes cluster credentials. I want to access Radius via my typical user credentials from my IdP.
```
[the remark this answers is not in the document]
zachcasper: No. I reframed this to be if you're authenticating via Kubernetes cluster-admin, then you are the cluster-admin and have * access to all tenants. When you create a radius-admin, it's for a specific tenant.

architecture/2024-11-authz-feature-spec.md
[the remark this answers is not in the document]
sylvainsf: I get the idea, but I don't know if we need to solve a company's change control problems. The thinking here I'm reading is that some group might be responsible for a given recipe's development and to lock down who can deploy new versions of it. I do think this is a common requirement but not necessarily one we need to enforce in code for V1.

architecture/2024-11-authz-feature-spec.md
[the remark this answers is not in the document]
zachcasper: I listed `cluster-admin` as a distinct user with admin across all tenants in the role definitions. So they are distinct, but have two different authenticators (K8s versus IdP).

architecture/2024-11-authz-feature-spec.md
```diff
@@ -0,0 +1,390 @@
...
+
+#### Out-of-the-box roles
+
+**Radius Administrator** – The Radius Administrator has superuser access to the Radius system. 
+
+**Resource Group Administrator** – The Resource Group Administrator role grants permission to manage resource groups, role definitions, and role assignments.
+
+**Developer** – The Developer role grants permission to create applications and application resources. It does not allow them to deploy resources to an environment without the associated Deployer role. Out of the box, developers create, update, delete, list, get, and deploy resources from `Application.Core` (except Environment and Extenders), `Applications.Datastores`, `Applications.Messaging`, and `Applications.Dapr`. Environments is reserved for the platform engineering team. Extenders are not included by default to encourage the use of pre-defined resource types. If the platform engineer creates additional resource types, they can add the resource type namespace (aka resource provider name) to this role.
```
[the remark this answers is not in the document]
zachcasper: My view is extenders and being able to specify a recipe on a resource breaks the separation of duties between developers and platform engineers.

architecture/2024-11-authz-feature-spec.md
```diff
@@ -0,0 +1,524 @@
...
+  * The default resource group is renamed `env-default`
+* `rad run` and `rad deploy` error out if a group is not specified in the command line or config.yaml
+* `rad run` and `rad deploy` now create the specified group if it does not exist
+* `rad run` and `rad deploy` respect the role definitions and assignments described in this document
+* The `Creating application...` output from `rad run` is modified to be more precise
+* New `rad workspace update` command
+
+> [!NOTE]
+>
+> The naming convention for resource groups is `app-group-name` for groups container applications and application resources and `env-group-name` for environments and their recipes. "Resource group" is the long name and should be used in the UI and documentation. "Group" is the short name used in CLI commands or other technical contexts.
```
rynowak, line 82: For the naming convention, any resource group contain an application needs to also contain the environment name as well. 

If you imagine an app named `shoestore` that's deployed to `prod-east` and `prod-west`, the problem is obvious. These two deployments need to be separate groups, or they collide.

eg:

- `prod-east-shoestore`
- `prod-west-shoestore`

architecture/2024-11-authz-feature-spec.md
```diff
@@ -0,0 +1,524 @@
...
+
+> [!NOTE]
+>
+> The UI should refer to resources with the fully qualified name unless the resource group context is obvious. In the examples below, The application is referred to as "getting-started" since the "app-getting-started" resource group was referenced. But the environment is referred to as "env-default/my-kube-context" since it was not previous referenced.
```
rynowak, line 86: +1

This is an obvious improvement we can make to a bunch of experiences.

architecture/2024-11-authz-feature-spec.md
```diff
@@ -0,0 +1,524 @@
...
+
+The operation fails and informs the user interactively if:
+
+* There is no group specified via the command line or in the workspace config
+* The user's role assignment does not have permission to create a resource group; the user should contact the Radius administrator (not applicable in this user story since Radius was just installed, but true otherwise)
+* The user's role assignment does not have permission to deploy applications to the environment; the user should specify an alternative environment via `--environment` or by updating the workspace, or contacting the Radius administrator  (not applicable in this user story since Radius was just installed, but true otherwise)
+
+##### User Experience 2
+
+The user can also add the group to the workspace via a new `rad workspace update` command:
```
rynowak, line 125: we already have this as `rad group switch`

architecture/2024-11-authz-feature-spec.md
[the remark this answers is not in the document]
rynowak: I think this is worth more consideration since you added automatic resource group creation to `rad run` and `rad deploy`.

architecture/2024-11-authz-feature-spec.md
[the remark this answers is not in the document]
rynowak: I'm aligned on this. It feels scary but when you do the analysis it becomes clear that it's not an escalation of privilege. The cluster admin could do any number of things to tamper and grant themselves root privilege on Radius.

architecture/2024-11-authz-feature-spec.md
rynowak: @zachcasper can you update this since we decided not to have a default application resource group?

zachcasper: Updated

architecture/2024-11-authz-feature-spec.md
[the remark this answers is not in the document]
zachcasper: @rynowak Ideally, for scenario 2 (simple enterprise use case), the admin could assign a developer role to a user and that developer would be granted the ability to create resource groups. The developer would have access to that new resource group but not other non-admin users. That does not fit well within our existing model.

I'm assuming this could be implemented with each resource would have an owner property (e.g., developer deploys an app, resource group gets created and the developer assigned as owner), or some other fancy ACL on the resource. That seems complex and not the direction we want to go. 

Thoughts?

rynowak (approved):

PR #77, opened 2025-01-23
Serverless Container Runtimes

willtsai:
add feature spec for Serverless Container Runtimes

### DIFF
--- a/architecture/2025-01-serverless-feature-spec.md
+++ b/architecture/2025-01-serverless-feature-spec.md
@@ -1,0 +1,121 @@
+# Topic: Serverless Container Runtimes
+
+* **Author**: Will Tsai (@willtsai)
+
+## Topic Summary
+<!-- A paragraph or two to summarize the topic area. Just define it in summary form so we all know what it is. -->
+Given the importance of serverless infrastructure in the modern application landscape, it is a priority for Radius to expand beyond Kubernetes and support serverless compute platforms. The initial expansion will focus on support for an unopinionated serverless container runtime (e.g. Azure Container Instances, AWS Elastic Container Service) before exploring integrations with other more opinionated serverless platforms (e.g. AWS Fargate, Azure Container Apps).
+
+This document describes the high-level overview for expanding the Radius platform to enable management of serverless container runtimes. The goal is to provide a seamless experience for developers to deploy and manage serverless containers in a way that is consistent with the existing Radius model, including Environments, Applications, and Resources.
+
+### Top level goals
+<!-- At the most basic level, what are we trying to accomplish? -->
+1. Genericize the Radius platform to support deployment of applications on serverless (and other) compute platforms beyond Kubernetes.
+1. Enable developers to deploy and manage serverless containers in a way that is consistent with the existing Radius environment and application model.
+1. Make all Radius features (Recipes, Connections, App Graph, UDT, etc.) available to engineers building applications on Radius-enabled supported serverless platforms.
+1. Radius support for unopinionated or process-driven serverless container runtimes (e.g. AWS Elastic Container Service, Azure Container Instances, Azure Container Apps, AWS Fargate).
+
+### Non-goals (out of scope)
+<!-- What are we explicitly not trying to accomplish? -->
+1. Hosting the Radius control plane on serverless (and other) compute platforms outside of Kubernetes.
+1. Radius support for more opinionated or event-driven serverless platforms (e.g. Azure Functions, AWS Lambda).
+
+## User profile and challenges
+<!-- Define the primary user and the key problem / pain point we intend to address for that user. If there are multiple users or primary and secondary users, call them out.   -->
+Primarily, our users are application developers, operators, and platform engineers who wish to leverage Radius in building and managing applications on serverless compute platforms. Secondarily, our platform engineering users may have their own internal stakeholders (e.g. developers and IT teams) for whom they are building a developer platform.
+
+### User persona(s)
+<!-- Who is the target user? Include size/org-structure/decision makers where applicable. -->
+**IT Operator**: An IT operator is responsible for helping developers deploy and manage applications on serverless compute platforms. They are familiar with containerized applications (e.g. Docker) and may have experience with Kubernetes and other cloud-native technologies. They are responsible for ensuring that applications are deployed and running correctly, and that the platform is secure, scalable, reliable, and follows best practices. They use Radius to preconfigure Environments and Recipes for use by the application developers that they support.
+
+**Application Developer**: A developer building applications that need to run and be managed on serverless compute platforms. They are familiar with containerized applications (e.g. Docker), but may or may not have experience with Kubernetes and other cloud-native technologies. With Radius, they will primarily deal with application definition files (e.g. `app.bicep`) and will leverage Radius features like Connections and Recipes. In many cases, their environments will be provided to them and preconfigured with the necessary settings and Recipes.
+
+**Platform Engineer**: A platform engineer responsible for building and maintaining the developer platform that developers and operators use to build and deploy applications. They are likely familiar with Kubernetes and other cloud-native technologies, but are definitely familiar with containerized applications (e.g. Docker). They are responsible for ensuring that the platform is secure, scalable, reliable, easy to use, and enforces best practices.
+
+**Site Reliability Engineer**: Applies updates or patches to compute infrastructure and workloads as needed to ensure the application scales appropriately to continue running without issue.
+
+### Challenge(s) faced by the user
+<!-- What challenges do the user face? Why are they experiencing pain and why do current offerings not meet their need? -->
+With the additional complexity and specialized knowledge required to manage and operate Kubernetes clusters, many users are looking to serverless compute platforms as a way to simplify their infrastructure and reduce operational overhead. Services like ACI and ECS provide a way to run containerized workloads without the need to manage the underlying infrastructure, while still providing the benefits of containers like portability and scalability. Azure Container Apps and AWS Fargate take this a step further by providing a fully managed serverless container platform that abstracts away the need to manage containers altogether.
+
+Users who build applications exclusively on serverless compute platforms are not able to adopt Radius. These users face the same challenges as those who have chosen to adopt Radius for Kubernetes: making sense of complex architectures, suboptimal troubleshooting experiences, pain in dealing with a plethora of cross-platform tools, difficulties in enforcing best practices, and hindered team collaboration due to unclear separation of concerns.
+
+Even if the user uses a mix of Kubernetes and serverless and they have adopted Radius, the same challenges persist and are even exacerbated when they can only use Radius for only their Kubernetes applications.
+
+### Positive user outcome
+<!-- What is the positive outcome for the user if we deliver this, i.e. what is the value proposition for the user? Remember, this is user-centric. -->
+The value to the user is that they can now use Radius to build and manage their applications on serverless compute platforms and thus taking advantage of the same benefits that Radius provides for Kubernetes applications.
+
+## Key scenarios
+<!-- List ~3-7 high level scenarios to clarify the value and point to how we will decompose this big area into component capabilities. We may ultimately have more than one level of scenario. -->
+
+### Scenario 1: Model Radius Environment, Application, and Container resources for serverless
+<!-- One or two sentence summary -->
+Enable the ability to define and run serverless compute with necessary extensions in a Radius [environment](https://docs.radapp.io/reference/resource-schema/core-schema/environment-schema/). 
+
+Users can specify Radius [application](https://docs.radapp.io/reference/resource-schema/core-schema/application-schema/) definitions for applications running on serverless compute platforms. 
+> Must be sure to include serverless platform specific customizations via a `runtimes` property, similar to how Kubernetes patching was implemented for [containers](https://docs.radapp.io/reference/resource-schema/core-schema/container-schema/#runtimes).
+
+Allow for Radius abstraction of a [container](https://docs.radapp.io/reference/resource-schema/core-schema/container-schema/) resource that can be deployed to serverless compute platforms, leveraging the [`runtimes` property](https://docs.radapp.io/reference/resource-schema/core-schema/container-schema/#runtimes). Must be sure to include serverless platform specific configurations via [connections](https://docs.radapp.io/reference/resource-schema/core-schema/container-schema/#connections), [extensions](https://docs.radapp.io/reference/resource-schema/core-schema/container-schema/#extensions), and routing to other resources via [gateways](https://docs.radapp.io/reference/resource-schema/core-schema/gateway/).
+> One idea to explore is whether we can we build extensibility via Recipes - i.e. allow Recipes for Containers, which themselves can be serverless containers.
+
+### Scenario 2: User interfaces for serverless--Radius API, CLI, Dashboard
+<!-- One or two sentence summary -->
+Enable deployment and management of serverless compute resources via the existing Radius API and CLI commands.
+
+Serverless resources that are modeled in Radius should be available in the App Graph and Dashboard for visualization and management.
+
+### Scenario 3: Punch-through to platform-specific features and incremental adoption of Radius into existing serverless applications
+<!-- One or two sentence summary -->
+Allow for platform-specific features to be used in Radius applications via abstraction "punch-through" mechanisms, similar to how Kubernetes-specific features are supported in Radius via [base YAML or PodSpec patching](https://docs.radapp.io/guides/author-apps/containers/overview/#kubernetes) functionalities. 
+> Stretch goal: Ability to add Radius to existing serverless applications without requiring a full rewrite of the application, similar to how Radius can be added to existing Kubernetes applications via [Kubernetes manifest](https://docs.radapp.io/tutorials/add-radius/) or [Helm chart](https://docs.radapp.io/tutorials/helm/) annotations. Note: The Kubernetes/Helm support works because Kubernetes itself is extensible. Other systems like ACA are not extensible in the same way but we should explore if there are options to make this work.
+
+## Key dependencies and risks
+<!-- What dependencies must we take in order to enable this scenario? -->
+<!-- What other risks are you aware of that need to be mitigated. If you have a mitigation in mind, summarize here. -->
+<!-- **Dependency Name** – summary of dependency.  Issues/concerns/risks with this dependency -->
+<!-- **Risk Name** – summary of risk.  Mitigation plan if known. If it is not yet known, no problem. -->
+**Dependency: orchestrator and API for the underlying serverless platform** - For each additional serverless platform that Radius supports, we have a dependency on the underlying orchestrator for that platform. For example, for Azure Container Instances, we depend on the Azure API. For AWS Elastic Container Service, we depend on the AWS API. There is a risk that the underlying orchestrator may not provide the necessary APIs to support the Radius model, the APIs may not be stable or reliable, or the APIs may not be publicly available. This will need to be investigated as a part of the implementation.
+
+**Risk: platform specific features that cannot be implemented in Radius** - There might be platform-specific features that cannot (or should not) be implemented in Radius. For example, Kubernetes has features for taints and tolerations that are not common across compute platforms and thus should not be implemented in Radius. This risk can be mitigated by providing mechanisms to punch-through the Radius abstraction and use platform-specific features directly, like the [Kubernetes customization options](https://docs.radapp.io/guides/author-apps/containers/overview/#kubernetes) currently available in Radius.
+
+**Risk: Differing platforms between Radius control plane and application runtime** - There is a need to deploy and maintain a Kubernetes cluster to host the Radius control plane is needed even if the application is completely serverless and this might not work for some customers (e.g. those who absolutely do not want to use Kubernetes). For now we will accept this risk and consider alternative hosting platforms for the Radius control plane to be out of scope.
+
+**Risk: Significant refactoring work anticipated** - There will be a significant amount of work to refactor the Radius code as a part of implementation, especially since the current codebase is heavily Kubernetes-centric. While this is an unavoidable risk, we will ensure that the refactoring work is done with future extensibility in mind so that it becomes easier to add support for additional compute platforms in the future.
+
+## Key assumptions to test and questions to answer
+<!-- If you are making assumptions that, if incorrect, would cause us to significantly alter our approach to this scenario, make them explicit here.  Also call out how / when you plan to validate key assumptions. -->
+<!-- What big questions must we answer in order to clarify our plan for this scenario.  When and how do you plan to answer those questions (prototype feature x, user research, competitive research, etc.) -->
+**Assumption**: The Radius model can be extended to support serverless compute platforms without significant changes to the existing model. Serverless platforms have orchestrators that expose APIs that can be used to model the resources in Radius. We have validated these assumptions by building a prototype for Azure Container Instances.
+
+**Assumption**: As a result of the implementation, the refactoring done to the core Radius codebase to support Serverless container runtimes should also allow for easier extensibility of Radius to support other new platforms going forward. In other words, the work done for Serverless should be genericized within the Radius code and not resemble custom implementations for Kubernetes and select Serverless platforms.
+
+**Question**: How do we handle the case where the underlying serverless platform (especially opinionated ones like Azure Container Apps or AWS Fargate) does not provide the necessary APIs to support the Radius model? We will answer this question by building prototypes for additional serverless platforms and evaluating the feasibility of supporting them in Radius.
+
+## Current state
+<!-- If we already have some ongoing investment in this area, summarize the current state and point to any relevant documents. -->
+There has been a proof-of-concept for ACI support in Radius built on top of the ACI nGroups features. Here is a demo of the prototype that has been created so far: https://youtu.be/NMNZE22nSQI?si=Dq7Q5WVKgHularsO&t=1201
+
+## Details of user problem
+<!-- <Write this in first person. You basically want to summarize what “I” as a user am trying to accomplish, why the current experience is a problem and the impact it has on me, my team, my work and or biz, etc…. i.e. “When I try to do x aspect of cloud native app development, I have the following challenges / issues….<details>. Those issues result in <negative impact those challenges / issues have on my work and or business.> -->
+
+## Desired user experience outcome
+<!-- <Write this as an “I statement” that expresses the new capability from user perspective … i.e. After this scenario is implemented “I can do, x, y, z, steps in cloud native app developer and seamlessly etc... As a result <summarize positive impact on my work / business>  -->
+
+### Detailed user experience
+ <!-- <List of steps the user goes through from the start to the end of the scenario to provide more detailed view of exactly what the user is able to do given the new capabilities>  -->
+<!-- Step 1
+Step 2
+… -->
+
+## Key investments
+<!-- List the features required to enable this scenario(s). -->
+
+### Feature 1
+<!-- One or two sentence summary -->
+
+### Feature 2
+<!-- One or two sentence summary -->
+
+### Feature 3
+<!-- One or two sentence summary -->


architecture/2025-01-serverless-feature-spec.md
```diff
@@ -0,0 +1,279 @@
...
+### Top level goals
+<!-- At the most basic level, what are we trying to accomplish? -->
+1. Genericize the Radius platform to support deployment of applications on serverless (and other) compute platforms beyond Kubernetes.
+1. Enable developers to deploy and manage serverless containers in a way that is consistent with the existing Radius environment and application model.
+1. Make all Radius features (Recipes, Connections, App Graph, UDT, etc.) available to engineers building applications on Radius-enabled supported serverless platforms.
+1. Radius support for unopinionated or process-driven serverless container runtimes (e.g. AWS Elastic Container Service, Azure Container Instances, Azure Container Apps, AWS Fargate).
+
+### Non-goals (out of scope)
+<!-- What are we explicitly not trying to accomplish? -->
+1. Hosting the Radius control plane on serverless (and other) compute platforms outside of Kubernetes.
```
ytimocin, line 31: Does this mean that Radius will still be installed on a Kubernetes cluster but will deploy resources to a Serverless platform if requested?

ytimocin: If so, wouldn't that be a blocker if the user doesn't have a Kubernetes cluster?

architecture/2025-01-serverless-feature-spec.md
```diff
@@ -0,0 +1,279 @@
...
+
+## Key scenarios
+<!-- List ~3-7 high level scenarios to clarify the value and point to how we will decompose this big area into component capabilities. We may ultimately have more than one level of scenario. -->
+
+### Scenario 1: Model Radius Environment, Application, and Container resources for serverless
+<!-- One or two sentence summary -->
+Enable the ability to define and run serverless compute with necessary extensions in a Radius [environment](https://docs.radapp.io/reference/resource-schema/core-schema/environment-schema/). 
+
+Users can specify Radius [application](https://docs.radapp.io/reference/resource-schema/core-schema/application-schema/) definitions for applications running on serverless compute platforms. 
+> Must be sure to include serverless platform specific customizations via a `runtimes` property, similar to how Kubernetes patching was implemented for [containers](https://docs.radapp.io/reference/resource-schema/core-schema/container-schema/#runtimes).
```
sylvainsf, line 68: One of the earliest punch-throughs we supported for Kubernetes itself is annotations, I don't think V1 needs to address portability of k8s annotations but as long as this schema is flexible we may need to look at how that would work. ECS has container metadata similar to annotations for instance.

architecture/2025-01-serverless-feature-spec.md
```diff
@@ -0,0 +1,279 @@
...
+
+Allow for Radius abstraction of a [container](https://docs.radapp.io/reference/resource-schema/core-schema/container-schema/) resource that can be deployed to serverless compute platforms. Must be sure to include serverless platform specific configurations via [connections](https://docs.radapp.io/reference/resource-schema/core-schema/container-schema/#connections), [extensions](https://docs.radapp.io/reference/resource-schema/core-schema/container-schema/#extensions), and routing to other resources via [gateways](https://docs.radapp.io/reference/resource-schema/core-schema/gateway/).
+> One future idea to explore is whether we can we build extensibility via Recipes - i.e. allow Recipes for Containers, which themselves can be serverless containers.
```
ytimocin, line 71: Do you mean that we will and should have Recipes that will let us deploy Serverless Containers? Not fully understanding this sentence here.

architecture/2025-01-serverless-feature-spec.md
```diff
@@ -0,0 +1,279 @@
...
+Allow for platform-specific features to be used in Radius applications via abstraction "punch-through" mechanisms, similar to how Kubernetes-specific features are supported in Radius via [base YAML or PodSpec patching](https://docs.radapp.io/guides/author-apps/containers/overview/#kubernetes) functionalities that is achieved through the [`runtimes`](https://docs.radapp.io/reference/resource-schema/core-schema/container-schema/#runtimes) property in the container definition. For example, ACI has a confidential containers offering that may not be common across serverless platforms, but Radius should allow for this feature to be used in ACI containers.
+> Stretch goal: Ability to add Radius to existing serverless applications without requiring a full rewrite of the application, similar to how Radius can be added to existing Kubernetes applications via [Kubernetes manifest](https://docs.radapp.io/tutorials/add-radius/) or [Helm chart](https://docs.radapp.io/tutorials/helm/) annotations. Note: The Kubernetes/Helm support works because Kubernetes itself is extensible. Other systems like ACA are not extensible in the same way but we should explore if there are options to make this work.
```
kachawla, line 76: Could you share an example for this?

architecture/2025-01-serverless-feature-spec.md
brooke-hamilton, line 31: This is really interesting - users will be required to have Kubernetes even when they do not want to deploy to Kubernetes. Is the reason for this to reduce the scope of the effort? Or is there another reason for keeping the control plane on Kubernetes only?

architecture/2025-01-serverless-feature-spec.md
```diff
@@ -0,0 +1,279 @@
...
+
+**Risk: platform specific features that cannot be implemented in Radius** - There might be platform-specific features that cannot (or should not) be implemented in Radius. For example, Kubernetes has features for taints and tolerations that are not common across compute platforms and thus should not be implemented in Radius. This risk can be mitigated by providing mechanisms to punch-through the Radius abstraction and use platform-specific features directly, like the [Kubernetes customization options](https://docs.radapp.io/guides/author-apps/containers/overview/#kubernetes) currently available in Radius.
+
+**Risk: Differing platforms between Radius control plane and application runtime** - There is a need to deploy and maintain a Kubernetes cluster to host the Radius control plane is needed even if the application is completely serverless and this might not work for some customers (e.g. those who absolutely do not want to use Kubernetes). For now we will accept this risk and consider alternative hosting platforms for the Radius control plane to be out of scope.
```
sylvainsf, line 91: I feel like this will also touch on Radius multi-cluster. Inherently if Radius is on k8s but managing ACI the model will be extremely similar to multiple k8s clusters.

architecture/2025-01-serverless-feature-spec.md
```diff
@@ -0,0 +1,279 @@
...
+> Stretch goal: Ability to add Radius to existing serverless applications without requiring a full rewrite of the application, similar to how Radius can be added to existing Kubernetes applications via [Kubernetes manifest](https://docs.radapp.io/tutorials/add-radius/) or [Helm chart](https://docs.radapp.io/tutorials/helm/) annotations. Note: The Kubernetes/Helm support works because Kubernetes itself is extensible. Other systems like ACA are not extensible in the same way but we should explore if there are options to make this work.
+
+### Scenario 3: User interfaces for serverless--Radius API, CLI, Dashboard
+<!-- One or two sentence summary -->
+Enable deployment and management of serverless compute resources via the existing Radius API and CLI commands. Serverless resources that are modeled in Radius should be available in the App Graph and Dashboard for visualization and management.
```
ytimocin, line 80: Are we going to have a scenario where a user can have applications running on Kubernetes clusters and Serverless platforms? I am talking about an app.bicep file with containers with different runtimes.

ytimocin: Combination of both worlds.

architecture/2025-01-serverless-feature-spec.md
```diff
@@ -0,0 +1,279 @@
...
+
+## Key dependencies and risks
+<!-- What dependencies must we take in order to enable this scenario? -->
+<!-- What other risks are you aware of that need to be mitigated. If you have a mitigation in mind, summarize here. -->
+<!-- **Dependency Name** – summary of dependency.  Issues/concerns/risks with this dependency -->
+<!-- **Risk Name** – summary of risk.  Mitigation plan if known. If it is not yet known, no problem. -->
+**Dependency: orchestrator and API for the underlying serverless platform** - For each additional serverless platform that Radius supports, we have a dependency on the underlying orchestrator for that platform. For example, for Azure Container Instances, we depend on the Azure API. For AWS Elastic Container Service, we depend on the AWS API. There is a risk that the underlying orchestrator may not provide the necessary APIs to support the Radius model, the APIs may not be stable or reliable, or the APIs may not be publicly available. This will need to be investigated as a part of the implementation.
```
ytimocin, line 87: Any local Serverless Platform like a local Kubernetes Cluster?

architecture/2025-01-serverless-feature-spec.md
brooke-hamilton, line 31: Would a side effect of running the control plane on Kubernetes for serverless deployments be that we could run the Radius control plane separately from the deployed cluster?

architecture/2025-01-serverless-feature-spec.md
ytimocin, line 91: Why can't we run Radius on just Docker? I believe Dapr does that with `dapr init` vs `dapr init -k`.

architecture/2025-01-serverless-feature-spec.md
willtsai: > Is the reason for this to reduce the scope of the effort?
Yes, this is precisely the reason. Hosting Radius itself on serverless or other compute platforms will be a future feature that follows this.

architecture/2025-01-serverless-feature-spec.md
```diff
@@ -0,0 +1,279 @@
...
+### Step 1: Define and deploy a Radius Environment for serverless
+
+The user defines a new Radius Environment for serverless compute by creating a new Environment definition file (e.g. `env.bicep`) and specifying the necessary settings for the serverless platform before deploying the environment using `rad deploy env.bicep`.
+
+```diff
+resource environment 'Applications.Core/environments@2023-10-01-preview' = {
+  name: 'myenv'
+  properties: {
+    compute: {
++      kind: 'aci'   // Required. The kind of container runtime to use, e.g. 'aci' for Azure Container Instances and 'ecs' for AWS Elastic Container Service
```
ytimocin, line 123: Can we have a kind that would support multiple kinds? Can I have two apps (one on Kubernetes cluster and one on ECS) in a single environment? I see that as a possible use case for a lot of users.

architecture/2025-01-serverless-feature-spec.md
```diff
@@ -0,0 +1,279 @@
...
+    ]
+    runtimes: {
++      aci: {
++        // Add ACI-specific properties here to punch-through the Radius abstraction, e.g. sku, osType, etc.
++        sku: 'Confidential' // 'Standard', 'Dedicated', etc.
++        }
++      ecs: {
++        // Add AWS ECS-specific properties here to punch-through the Radius abstraction
++        }
++      }
```
kachawla, line 203: Is the expectation here that only one runtime will be allowed at once and it has to match what's configured as the compute for the environment?

architecture/2025-01-serverless-feature-spec.md
kachawla, line 123: What's the motivation to strictly map an environment to only one type of compute?

architecture/2025-01-serverless-feature-spec.md
ytimocin: It requires a lot of code changes but would that be a good investment? 🧐

architecture/2025-01-serverless-feature-spec.md
```diff
@@ -0,0 +1,279 @@
...
+- ACI [Confidential Containers](https://learn.microsoft.com/en-us/azure/container-instances/container-instances-confidential-overview): ACI feature that provides a secure enclave for your containerized applications to run in a confidential computing environment.
+- ACI [Spot Instances](https://learn.microsoft.com/en-us/azure/container-instances/container-instances-spot-containers-overview): ACI feature that allows you to run interruptible workloads at a reduced cost compared to the standard price by taking advantage of unused capacity in Azure datacenters.
+
+### Top level goals
+<!-- At the most basic level, what are we trying to accomplish? -->
+1. Genericize the Radius platform to support deployment of applications on serverless (and other) compute platforms beyond Kubernetes.
```
lakshmimsft, line 24: Is there a subset of the listed serverless platforms we aim to target?

architecture/2025-01-serverless-feature-spec.md
zachcasper: I disagree this is a valid use case. If there was an app that the platform team wanted to run on ECS and another on EKS, they would configure two different environments. Semantically, an environment is a single place, so we should not have multiple places (i.e. different container platforms) within a single environment.

architecture/2025-01-serverless-feature-spec.md
```diff
@@ -0,0 +1,279 @@
+# Topic: Serverless Container Runtimes
```
zachcasper, line 1:
```suggestion
# Topic: Serverless Container Platforms
```

zachcasper: A container runtime is a different thing: containerd, CRI-O, Docker

zachcasper: You likely needs a glossary. 

Serverless Container Platform (ACI, ACA, ECS Fargate, Cloud Run)
Event-driven Serverless Platform (Lambda, Azure Functions, Google Cloud Functions)
Container Runtime (containerd, Docker, CRI-O)
Service
Function
Job
ACI Ngroups
...

architecture/2025-01-serverless-feature-spec.md
```diff
@@ -0,0 +1,279 @@
...
+
+### Step 2: Define a Radius Application for serverless
+
+The user defines a new Radius Application for serverless compute by creating a new Application definition file (e.g. `app.bicep`) and specifying the necessary settings for the container runtime platform. Note that the application definitions are container runtime platform-agnostic, thus this same application definition can be deployed to both Kubernetes and serverless compute platforms.
+
+```diff
+resource app 'Applications.Core/applications@2023-10-01-preview' = {
+  name: 'myapp'
+  properties: {
+    environment: environment
```
zachcasper, line 153: The developer should not have to specify the environment in their application definition. The environment should be specified at deployment time; e.g., `rad deploy`. We also do not want to allow the developer to specify environment extensions. Those should be reserved for the platform engineer.

architecture/2025-01-serverless-feature-spec.md
zachcasper, line 68: That's unfortunate that this was encoded as `runtimes`. A more precise word is `orchestrator`. Consider changing. There are use cases where a developer or platform engineer needs to specify the container runtime (containerd is the default, but some applications have dependencies on Docker).

architecture/2025-01-serverless-feature-spec.md
```diff
@@ -0,0 +1,279 @@
...
+resource db 'Applications.Datastores/redisCaches@2023-10-01-preview' = {
+  name: 'db'
+  properties: {
+    application: application
+    environment: environment
+  }
+}
+```
+
+### Step 5: Deploy the application to the serverless platform
```
zachcasper, line 248: We need to talk about how ingress works. We need L4 ingress at a minimum and ideally L7. I actually do not see how to do L4 ingress today. The gateway resource looks like it's only L7.

architecture/2025-01-serverless-feature-spec.md
zachcasper: We should add a prioritized list. That would be:
1. ACI
2. ECS
3. ACA
4. Cloud Run

Then maybe:

6. Knative
7. Lambda
8. Azure Functions
9. Cloud Functions

zachcasper: Then one day AWS Batch, Azure Batch, and Google Cloud Batch. With this portfolio, we could easily model long-running services, functions, and jobs abstractly which is pretty much everything.

architecture/2025-01-serverless-feature-spec.md
zachcasper: This is a separate roadmap item. https://github.com/radius-project/roadmap/issues/39

zachcasper: This is a separate roadmap item. Sometimes referred to as multicluster. I refer to it as deploy to external Kubernetes cluster. Or the title of this https://github.com/radius-project/roadmap/issues/42

architecture/2025-01-serverless-feature-spec.md
zachcasper: You can already do this with a UDT and a recipe. The extensibility would come via UDTs not recipes.

architecture/2025-01-serverless-feature-spec.md
zachcasper: Knative would be nice eventually.

zachcasper (changes requested):

architecture/2025-01-serverless-feature-spec.md
```diff
@@ -0,0 +1,279 @@
...
++        scope: '/subscriptions/.../resourceGroups/myrg'
++      }
++    }
+    extensions: [
+      {
+        kind: 'kubernetesMetadata'
+        labels: {
+          'team.contact.name': 'frontend'
+        }
++        // Add other serverless platform-specific extensions here
```
sk593, line 138: Is the expectation that we'd have one entry per extension or one per platform?

architecture/2025-01-serverless-feature-spec.md
```diff
@@ -0,0 +1,279 @@
...
+```diff
+resource environment 'Applications.Core/environments@2023-10-01-preview' = {
+  name: 'myenv'
+  properties: {
+    compute: {
++      kind: 'aci'   // Required. The kind of container runtime to use, e.g. 'aci' for Azure Container Instances and 'ecs' for AWS Elastic Container Service
++      namespace: 'default' // This is currently required, but may be made optional if it doesn't apply to all compute platforms
++      resourceGroup: '/subscriptions/.../resourceGroups/myrg' // This was introduced as a part of the ACI prototype, but may be changed depending on the implementation, need to figure out if namespace and resourceGroup should be combined into a single more generic property
+    }
++    providers: { // This was introduced as a part of the ACI prototype, but may be changed depending on the implementation
```
sk593, line 127: What's the advantage of adding a provider to the Bicep template vs having a user register a provider via `rad init`? The provider would still be tracked by Radius in the case of `rad init`

architecture/2025-01-serverless-feature-spec.md
```diff
@@ -0,0 +1,279 @@
+# Topic: Serverless Container Runtimes
+
+* **Author**: Will Tsai (@willtsai)
```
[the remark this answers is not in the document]
willtsai: Perhaps, but let's figure that out in a separate PR to clean up the locations of feature spec docs. Right now I'm just following the same pattern as previous feature specs being stored in the relevant area's folder, e.g. [UDTfeature spec](https://github.com/radius-project/design-notes/blob/main/architecture/2024-06-resource-extensibility-feature-spec.md) and [AuthZ feature spec](https://github.com/radius-project/design-notes/blob/main/architecture/2024-11-authz-feature-spec.md)

architecture/2025-01-serverless-feature-spec.md
```diff
@@ -0,0 +1,279 @@
...
+- ACI [Spot Instances](https://learn.microsoft.com/en-us/azure/container-instances/container-instances-spot-containers-overview): ACI feature that allows you to run interruptible workloads at a reduced cost compared to the standard price by taking advantage of unused capacity in Azure datacenters.
+
+### Top level goals
+<!-- At the most basic level, what are we trying to accomplish? -->
+1. Genericize the Radius platform to support deployment of applications on serverless (and other) compute platforms beyond Kubernetes.
+1. Enable developers to deploy and manage serverless containers in a way that is consistent with the existing Radius environment and application model.
+1. Make all Radius features (Recipes, Connections, App Graph, UDT, etc.) available to engineers building applications on Radius-enabled supported serverless platforms.
+1. Radius support for unopinionated or process-driven serverless container runtimes (e.g. AWS Elastic Container Service, Azure Container Instances, Azure Container Apps, AWS Fargate).
```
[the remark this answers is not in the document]
willtsai: "unopinionated" are platforms providing more generic compute primitives while "process-driven" platforms provide compute as a fully managed service that's based on running specific processes.

architecture/2025-01-serverless-feature-spec.md
willtsai: which platforms we target and when is more of a roadmap/prioritization decision that a feature spec detail, so i'll keep this prioritized list out of this doc for now.

architecture/2025-01-serverless-feature-spec.md
```diff
@@ -0,0 +1,279 @@
...
+With the additional complexity and specialized knowledge required to manage and operate Kubernetes clusters, many users are looking to serverless compute platforms as a way to simplify their infrastructure and reduce operational overhead. Services like ACI and ECS provide a way to run containerized workloads without the need to manage the underlying infrastructure, while still providing the benefits of containers like portability and scalability. Azure Container Apps and AWS Fargate take this a step further by providing a fully managed serverless container platform that abstracts away the need to manage containers altogether.
+
+Users who build applications exclusively on serverless compute platforms are not able to adopt Radius. These users face the same challenges as those who have chosen to adopt Radius for Kubernetes: making sense of complex architectures, suboptimal troubleshooting experiences, pain in dealing with a plethora of cross-platform tools, difficulties in enforcing best practices, and hindered team collaboration due to unclear separation of concerns.
```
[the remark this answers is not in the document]
willtsai: Good points, I'll incorporate this into the spec, especially:

> "forced choose a container platform prior to actually building the application"

> "It is very difficult to move between platforms once an application has been built—even if it is containerized."

architecture/2025-01-serverless-feature-spec.md
willtsai: Noted.

willtsai: Agreed, the abstraction "punch-throughs" are elaborated on in the later detailed user experience sections.

architecture/2025-01-serverless-feature-spec.md
willtsai: I'll clarify here, but I meant deploying a container resource using a recipe such that all the user has to do when declaring a container is specifying the bare minimum (e.g. name, application) and let the predefined recipe handle the rest. I suppose this will require UDT (e.g. defining specific types of containers as individual resource types), I will add this detail.

architecture/2025-01-serverless-feature-spec.md
willtsai: Do you mean an example of the feature on a serverless platform that is equivalent to Kubernetes annotations?

architecture/2025-01-serverless-feature-spec.md
willtsai: This will not be a functionality for serverless, at least not in the initial implementation. We'll require a single compute platform per environment, which means all containers within the same application must run in the same compute platform for a single deployment. I'll note this in the spec. This has the benefit of simplifying this serverless feature significantly but we also don't expect users to combine compute platforms for a single application as things like cross-platform DNS, ingress, etc. are challenging.

architecture/2025-01-serverless-feature-spec.md
willtsai: Do you mean running Radius on Docker within an ACI compute cluster?

willtsai: Very good point, I'll add a risk specifically mentioning that Serverless may require us to implement (or at least consider) the multi-cluster feature.

architecture/2025-01-serverless-feature-spec.md
willtsai: Agree with @zachcasper on this.

willtsai: @zachcasper has a good explanation above for this:
> If there was an app that the platform team wanted to run on ECS and another on EKS, they would configure two different environments. Semantically, an environment is a single place, so we should not have multiple places (i.e. different container platforms) within a single environment.

architecture/2025-01-serverless-feature-spec.md
willtsai: I think there can be a case where the provider for `rad init` might have a different scope than the `providers.scope` supplied in the `env.bicep`. For example, the scope of the cloud provider hosting the compute for a given environment might be `scope: '/subscriptions/${subscriptionId}/resourceGroups/${resourceGroup}'`. The provider credentials saved from a `rad init --full` command might have scope beyond this specific resource group and thus the same credentials can be used to deploy many environments with different scopes within the same provider subscription or account.

I will make a note of this nuance.

architecture/2025-01-serverless-feature-spec.md
willtsai: `environment` is a required property today (see the [reference doc](https://docs.radapp.io/reference/resource-schema/core-schema/application-schema/#properties)). The way we expect developers to use this today is to declare the `environment` as a parameter and let the `rad deploy` set the environment. I'll add this nuance into the example, good catch.

> `@description('The environment ID of your Radius Application. Set automatically by the rad CLI.')`
> `param environment string`

Today developers have the ability to specify application scoped extensions and that's what these are, they are not environment scoped extensions.

architecture/2025-01-serverless-feature-spec.md
willtsai: To make the app definition portable, we must allow several `runtimes` to be declared per container and these should all be optional for cases where the user wants to punch through the Radius abstraction. If they declare a `runtimes` property that doesn't match the targeted deployment environment's compute, we should simply ignore that property for that deployment. If they haven't declared any `runtimes` property that match the compute of the targeted deployment environment, then we should deploy their container assuming that no `runtimes` property was provided and thus no "punch-through" behavior will be applied.

I'll add this as a note to the spec, good question.

architecture/2025-01-serverless-feature-spec.md
```diff
@@ -0,0 +1,279 @@
...
+
+## Detailed user experience
+ <!-- <List of steps the user goes through from the start to the end of the scenario to provide more detailed view of exactly what the user is able to do given the new capabilities>  -->
+<!-- Step 1
+Step 2
+… -->
+
+### Step 1: Define and deploy a Radius Environment for serverless
+
+The user defines a new Radius Environment for serverless compute by creating a new Environment definition file (e.g. `env.bicep`) and specifying the necessary settings for the serverless platform before deploying the environment using `rad deploy env.bicep`.
```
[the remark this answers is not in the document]
willtsai: Today we have a `providers` environment property, perhaps the `kube-context` could be provided here?

> providers: {
>     // if the compute platform for this environment is ACI, then the provider must be Azure
>     azure: {
>       scope: '/subscriptions/.../resourceGroups/myrg'
>     }
>     // if the compute platform for this environment is ECS, then the provider must be AWS
>     aws: {
>       scope: '/planes/aws/aws/accounts/${account}/regions/${region}'
>     }
>   }

architecture/2025-01-serverless-feature-spec.md
```diff
@@ -0,0 +1,391 @@
...
++       // Add AWS ECS-specific properties here to punch-through the Radius abstraction
++     }
++      }
+  }
+}
+```
+
+> Note: We assume that the `sku: 'Confidential'` property for ACI (and other comparable properties across platforms like ECS) is scoped to the individual container and not the entire application or environment. We will need to validate this assumption as a part of the implementation.
+
+> Note: To make the app definition portable, we must allow several `runtimes` to be declared per container and these should all be optional for cases where the user wants to punch through the Radius abstraction. If they declare a `runtimes` property that doesn't match the targeted deployment environment's compute, we should simply ignore that property for that deployment. If they haven't declared any `runtimes` property that match the compute of the targeted deployment environment, then we should deploy their container assuming that no `runtimes` property was provided and thus no "punch-through" behavior will be applied.
```
brooke-hamilton, line 245: Does this also mean that an application could specify a punch through that is not used when deployed to a different environment. For example, could a user specify a K8S punch through and still deploy the app to ACI, in which case the K8S punch through would be ignored? If yes, then I agree. 👍

architecture/2025-01-serverless-feature-spec.md
```diff
@@ -0,0 +1,391 @@
...
+  properties: {
+    application: application
+    container: {
+      image: 'ghcr.io/radius-project/samples/demo:latest'
+      ports: {
+        web: {
+          containerPort: 3000
+        }
+      }
++     // all the other container properties should also be implemented (e.g. `env`, `readinessProbe`, `livenessProbe`, etc.)
```
sylvainsf, line 222: As container properties are considered portable we will need to evaluate and map them to equivalent properties in the target compute platform. We should explicitly call out the behavior when a developer defines a property not available in all supported compute platforms (deploy anyway and ignore unused properties?

architecture/2025-01-serverless-feature-spec.md
```diff
@@ -0,0 +1,391 @@
...
+
+### Step 6: Deploy the application to the serverless platform
+
+The user deploys the application to the serverless platform by running the `rad run` or `rad deploy` command with the application definition file targeting the serverless environment (e.g. `app.bicep`).
+
+### Step 7: View the serverless containers in the CLI and Radius Dashboard
+
+After successful deployment, the user can view the serverless containers in the CLI using the `rad app graph` command or via the Application Graph in the Radius Dashboard.
+
+## Key investments
```
brooke-hamilton, line 360: Should we add a key investment for Radius managing external environments? This could be a separately deployable feature that has value on its own, but it is definitely required for ACI support.

brooke-hamilton (commented):
🚀 Nice document 💯

architecture/2025-01-serverless-feature-spec.md
brooke-hamilton: Maybe a good future feature would be the ability to add additional platform-specific properties in downstream environments, similar to what a GitOps workflow would allow. For example, the dev specifies a K8S property and the operations person overrides that in production with a different value.

architecture/2025-01-serverless-feature-spec.md
willtsai: I think there are a certain set of core properties that would be common across all container runtimes (e.g. `image`, `ports`, `env`) and these should be included as top-level properties within the container definition and should not be ignored if provided, assuming that these core properties are limited to ones that are universal across container platforms. The properties that may not be universal across platforms (e.g. `sku: 'Confidential'`, `osType`) should not be top-level properties but should rather be encapsulated in the `runtimes` property. These properties within the runtimes can be ignored if they don't apply to the target deployment environment (see the note immediately following this section). I'll clarify this in the doc.

architecture/2025-01-serverless-feature-spec.md
willtsai: > For example, could a user specify a K8S punch through and still deploy the app to ACI, in which case the K8S punch through would be ignored?

Yes, this is exactly what I'm proposing, glad you agree :)

> For example, the dev specifies a K8S property and the operations person overrides that in production with a different value.

For this one, are you saying that we should add the ability to apply universal container configurations at the environment level that would override the container configurations defined within individual applications? If so, I can see a use case for this but I wonder if that's something RBAC (when that feature is available) can take care of? For example, the `sku` of the container compute is specifiable in the environment and then RBAC rules prevent developers from overriding `sku` configurations from within their application definitions.

architecture/2025-01-serverless-feature-spec.md
willtsai: I added a note in the risks section (see **Risk: Deploying to multiple clusters** ) where I call out that we might have to partially implement (or at least take into consideration) the [Inter-cluster app deployment and management](https://github.com/radius-project/roadmap/issues/42) feature. I hesitate to add this to the explicit set of features we plan to deliver as that itself is a separate (and possibly large) project that we probably don't want to fully implement as a subproject of serverless. In other words, we should only build enough functionality for Radius managing external environments that will satisfy our Serverless features. Let me know if you disagree and we can discuss further.

architecture/2025-01-serverless-feature-spec.md
```diff
@@ -0,0 +1,534 @@
+# Topic: Serverless Container Platforms
+
+* **Author**: Will Tsai (@willtsai)
+
+## Topic Summary
+<!-- A paragraph or two to summarize the topic area. Just define it in summary form so we all know what it is. -->
+Given the importance of serverless infrastructure in the modern application landscape, it is a priority for Radius to expand beyond Kubernetes and support additional container platforms with lower operational overhead. The initial expansion will focus on support for Azure Container Instances, then AWS Elastic Container Service including AWS Fargate. This will be followed by more feature rich platforms including Azure Container Apps and eventually Google CloudRun.
```
kachawla, line 7: > The initial expansion will focus on support for Azure Container Instances, then AWS Elastic Container Service including AWS Fargate. This will be followed by more feature rich platforms including Azure Container Apps and eventually Google CloudRun.

Just curious - how did we decide the priority of different platforms?

architecture/2025-01-serverless-feature-spec.md
```diff
@@ -0,0 +1,534 @@
...
+- [User Defined Types](https://github.com/radius-project/design-notes/blob/main/architecture/2024-06-resource-extensibility-feature-spec.md) (UDT): A Radius feature that allows you to define custom resource types that can be used in your application definitions. As of this writing, the feature is a work in progress with its design document available [here](https://github.com/radius-project/design-notes/blob/main/architecture/2024-07-user-defined-types.md).
+
+### Top level goals
+<!-- At the most basic level, what are we trying to accomplish? -->
+1. Genericize the Radius platform to support deployment of applications on serverless (and other) compute platforms beyond Kubernetes.
+1. Enable developers to deploy and manage serverless containers in a way that is consistent with the existing Radius environment and application model.
+1. Make all Radius features (Recipes, Connections, App Graph, UDT, etc.) available to engineers building applications on Radius-enabled supported serverless platforms.
```
brooke-hamilton, line 26: Are we saying that 100% of Radius features that are available on Kubernetes will also be available on serverless? Should we add a caveat to say that all serverless platforms may not support all Radius features?

architecture/2025-01-serverless-feature-spec.md
```diff
@@ -0,0 +1,534 @@
...
+1. Radius support for unopinionated (e.g. AWS Elastic Container Service, Azure Container Instances) or process-driven (Azure Container Apps, AWS Fargate, Google CloudRun) serverless container platforms.
+
+### Non-goals (out of scope)
+<!-- What are we explicitly not trying to accomplish? -->
+1. Hosting the Radius control plane on serverless (and other) compute platforms outside of Kubernetes. This is a separate project tracked in the roadmap: https://github.com/radius-project/roadmap/issues/39.
+1. The ability to run the Radius control plane separately from the compute cluster to which it is deploying applications. This is a separate project tracked in the roadmap: https://github.com/radius-project/roadmap/issues/42. However, given that the initial requirement is for a Kubernetes-hosted Radius control plane to deploy applications to a serverless compute platform, we might need to partially implement or at least take into consideration the separation of the Radius control plane cluster from the target deployment cluster as a part of serverless implementation.
```
ytimocin, line 32: We are kind of achieving this with this feature actually, right? Deploying to Serverless from a Radius that is running on Kubernetes.

architecture/2025-01-serverless-feature-spec.md
```diff
@@ -0,0 +1,534 @@
...
+
+## Detailed user experience
+ <!-- <List of steps the user goes through from the start to the end of the scenario to provide more detailed view of exactly what the user is able to do given the new capabilities>  -->
+<!-- Step 1
+Step 2
+… -->
+
+### Step 1: Define and deploy a Radius Environment for serverless
+
+The user defines a new Radius Environment for serverless compute by creating a new Environment definition file (e.g. `env.bicep`) and specifying the necessary settings for the serverless platform before deploying the environment using `rad deploy env.bicep`.
```
nithyatsu, line 133: Today, rad cli can speak only to a radius in kubernetes. 
We have to also capture the cli side changes that would be needed to speak with serverless. We might have to add on config.yaml we have today to capture the kind of radius instance. "kind" and "context" should support the new feature.
```
workspaces:
    default: dev
    items:
        default:
            connection:
                context: kind-kind
                kind: kubernetes
            environment: /planes/radius/local/resourceGroups/default/providers/Applications.Core/environments/default
            scope: /planes/radius/local/resourceGroups/default
```

architecture/2025-01-serverless-feature-spec.md
```diff
@@ -0,0 +1,534 @@
...
+
+**Assumption**: As a result of the implementation, the refactoring done to the core Radius codebase to support Serverless container runtimes should also allow for easier extensibility of Radius to support other new platforms going forward. In other words, the work done for Serverless should be genericized within the Radius code and not resemble custom implementations for Kubernetes and select Serverless platforms.
+
+**Assumption**: If there was an app that the platform team wanted to run on serverless (e.g. ECS, ACI) and another on Kubernetes (e.g. EKS, AKS) they would configure two different environments. Semantically, an environment is a single place, so we should not have multiple places (i.e. different container platforms) within a single environment. We think that this assumption is reasonable given multiple compute targets would require a cross container compute service discovery which users are unlikely to have the desire in setting up. We'll validate this assumption with feedback from potential users.
```
brooke-hamilton, line 115: Do we think that it would be difficult for a Radius instance to manage multiple environments at the same time? If the deployments specify which environment to deploy to, would the internal support of multiple environments be difficult?

architecture/2025-01-serverless-feature-spec.md
ytimocin: I think Docker uses **containerd** behind the scenes as the container runtime.

architecture/2025-01-serverless-feature-spec.md
```diff
@@ -0,0 +1,534 @@
...
+
+Allow for Radius abstraction of a [container](https://docs.radapp.io/reference/resource-schema/core-schema/container-schema/) resource that can be deployed to serverless compute platforms. Must be sure to include serverless platform specific configurations via [connections](https://docs.radapp.io/reference/resource-schema/core-schema/container-schema/#connections), [extensions](https://docs.radapp.io/reference/resource-schema/core-schema/container-schema/#extensions), and routing to other resources via [gateways](https://docs.radapp.io/reference/resource-schema/core-schema/gateway/).
+> One future idea to explore is whether we can we build extensibility via UDT and Recipes - i.e. allow Recipes to deploy predefined Container types, which themselves can be serverless containers. In other words, developers can deploy a UDT container resource using a Recipe such that all the user has to do when declaring a container is specifying the bare minimum (e.g. name, application) and let the predefined Recipe handle the rest. This will require the implementation of UDTs (e.g. defining specific types of containers as individual resource types).
```
kachawla, line 70: Have we considered directly building in recipe support natively into Radius Container resource rather than through UDT?

ytimocin, line 70: This would require us to add something like UDE (User-Defined Engine) since we will need the necessary engine (or driver) for that runtime.

ytimocin: And then the UDT can choose the driver that it needs to use.

architecture/2025-01-serverless-feature-spec.md
```diff
@@ -0,0 +1,534 @@
...
+
+### Scenario 2: Define and create core resources (i.e. Gateway, Secret Store, Extender) for use in serverless containers
+<!-- One or two sentence summary -->
+Enable the ability to define and create core Radius resources (i.e. Gateway, Secret Store, Extender) for use in serverless containers. Radius would leverage the solution available on the hosting platform to create and manage these resources. For example, if the serverless compute platform has a built-in secret store, Radius would use that instead of creating its own. This would allow for a consistent experience across different serverless compute platforms while still leveraging the unique features of each platform.
```
kachawla, line 74: `Extender` is a very open ended resource type, so I don't think it fits the category this scenario captures.

architecture/2025-01-serverless-feature-spec.md
```diff
@@ -0,0 +1,534 @@
...
+      image: 'ghcr.io/radius-project/samples/demo:latest'
+      ports: {
+        web: {
+          containerPort: 3000
+        }
+      }
++     // all the other container properties should also be implemented (e.g. `env`, `readinessProbe`, `livenessProbe`, etc.)
+    }
+    extensions: [
++     {
```
brooke-hamilton, line 229: Would we need to specify a runtime under `extensions`, or are we planning to match element names? If we are matching element names, we could run into naming conflicts where the same name means different things on different platforms.

architecture/2025-01-serverless-feature-spec.md
ytimocin: I mean I think we don't need to put in extra effort for this because Radius will already be deployed in a Kubernetes cluster and in the app.bicep file we should be able to define resources in both worlds.

architecture/2025-01-serverless-feature-spec.md
ytimocin: I mean running Radius on containers (https://docs.dapr.io/getting-started/install-dapr-selfhost/) so that the user wouldn't need a Kubernetes cluster to do a deployment of Serverless resources.

architecture/2025-01-serverless-feature-spec.md
```diff
@@ -0,0 +1,534 @@
...
+
+**Risk: Deploying to multiple clusters** - Given that the initial implementation for Serverless support will assume that the Radius control plane is hosted on Kubernetes and can deploy applications to serverless platforms, we might have to partially implement (or at least take into consideration) the [Inter-cluster app deployment and management](https://github.com/radius-project/roadmap/issues/42) feature that is currently in our roadmap. In other words, Radius should be able to manage applications across environments that are not on the same Kubernetes cluster as the control plane. This will be something to be explored as part of the tech design and implementation.
+
+**Risk: Significant refactoring work anticipated** - There will be a significant amount of work to refactor the Radius code as a part of implementation, especially since the current codebase is heavily Kubernetes-centric. While this is an unavoidable risk, we will ensure that the refactoring work is done with future extensibility in mind so that it becomes easier to add support for additional compute platforms in the future.
+
+## Key assumptions to test and questions to answer
```
brooke-hamilton, line 108: Should we add an assumption that Radius support of GitOps will not be affected by whether the platform is serverless or Kubernetes?

architecture/2025-01-serverless-feature-spec.md
```diff
@@ -0,0 +1,534 @@
...
+        path: '/'
++       destination: 'http://${demo.name}:3000'
+      }
+    ]
+  }
+}
+```
+
+> Note: For the Kubernetes implementation, Radius installs and configures Contour as the default ingress controller to enable Gateway features. For serverless platforms, we will need to evaluate the default ingress/gateway for each platform and implement it in a way that allows for flexibility in the future to support other ingress/gateway options.
+
```
nithyatsu, line 325: would we consider recipes for gateways so that based on environment, a gateway thats compatible could be deployed?

architecture/2025-01-serverless-feature-spec.md
willtsai: The GitOps integrations that Radius has (Flux) or will (ArgoCD) implmented is kunernetes-specific -- I will call this out

architecture/2025-01-serverless-feature-spec.md
```diff
@@ -0,0 +1,534 @@
...
+          'team.contact.name': 'frontend'
+        }
++       // Add other serverless platform-specific extensions here
+      }
+    ]
+  }
+}
+```
+
+> Note: The `providers` property in the environment is necessary to specify the scope of the environment and is different from the provider credentials that were registered using `rad init`. The provider credentials are used to authenticate with the underlying compute provider, while the `providers` environment property is used to specify the scope of the environment. Thus, we might consider renaming the Environment `providers` property to avoid confusion.
```
kachawla, line 175: Are we planning to restrict credentials to map 1-1 with the compute platform as well?

architecture/2025-01-serverless-feature-spec.md
willtsai: We wanted the first expansion beyond Kubernetes to be into an unopinionated compute platform that provides raw compute primitives that we can build upon. This would allow us to better learn how to model the extensibility of Radius core that paves the way for expansion into more opinionated or feature rich compute offerings.

architecture/2025-01-serverless-feature-spec.md
willtsai: I don't think we'd add that restriction since the rad credentials can apply to things beyond the compute.

architecture/2025-01-serverless-feature-spec.md
willtsai: I think as a part of the tech design we should address this and change the schema if it makes more sense because I agree with your point.

architecture/2025-01-serverless-feature-spec.md
willtsai: Yes, I think so, but this would be a separate feature that's outside the scope of what we're proposing for initial Serverless support. Related issue: https://github.com/radius-project/radius/issues/6521

architecture/2025-01-serverless-feature-spec.md
```diff
@@ -0,0 +1,534 @@
...
+
+### Step 6: Define a Radius Secret Store for the serverless container
+The user defines a `Applications.Core/secretStores` resource in the `app.bicep` and then deploys it into the serverless environment. Radius leverages the serverless hosting platform's secrets management solution to create and store the secret. The user can then reference the secret store in the application definition file (e.g. `app.bicep`) and use it to store secrets for the serverless container.
+
+```diff
+resource secrets 'Applications.Core/secretStores@2023-10-01-preview' = {
+  name: 'secrets'
+  properties:{
+    application: application
+    type: 'generic'
```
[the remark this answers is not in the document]
willtsai: Potentially - though I think this is a decision or change to be made at implementation

architecture/2025-01-serverless-feature-spec.md
willtsai: Good point, moved it out to its own scenario

architecture/2025-01-serverless-feature-spec.md
willtsai: ah gotcha, this is a different feature that I think should be tackled when we design and implement https://github.com/orgs/radius-project/projects/8/views/1?pane=issue&itemId=55074612&issue=radius-project%7Croadmap%7C42

architecture/2025-01-serverless-feature-spec.md
nithyatsu: yes. We can resolve this.

brooke-hamilton (approved):

sylvainsf (approved):
Approved! Thanks for all the hard work getting this just right.

PR #81, opened 2025-02-10
End-to-end UDT feature spec

zachcasper:
[User-defined resource type feature request](https://github.com/radius-project/radius/issues/6688)


architecture/2025-02-user-defined-resource-type-feature-spec.md
```diff
@@ -0,0 +1,1303 @@
...
+>
+> Since YAML has already been implemented, the initial release of user-defined resource types may continue use YAML if that eases the implementation. However, the format should change to Bicep very quickly and support for YAML removed. Radius will only support one file format. This change is reflected in the feature summary at the bottom of this document. 
+
+--—
+
+**User Experience** 
+
+```bash
+# Create the resource type
+rad resource-type create --from-file postgreSQL-resource-type.bicep
```
superbeeny, line 122: Do we envision other sources of configuration for resource types?

architecture/2025-02-user-defined-resource-type-feature-spec.md
```diff
@@ -0,0 +1,1303 @@
...
+
+```yaml
+extension radius
+
+resource MyCompany.Data/postgreSQL 'System.Resources/resourceTypes@2023-10-01-preview' = {
+  name: 'MyCompany.Data/postgreSQL'
+  api: {
+    // The API versioning scheme is left to the user to determine. It is just a string to Radius.
+    // The Kubernetes scheme is used here to demonstrate that a variety of schemes may be used.
+    version: 'v1alpha1'
```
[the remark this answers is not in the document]
zachcasper: Yes, the API version is line 137. Yes to the resourceTypes system resource.

architecture/2025-02-user-defined-resource-type-feature-spec.md
zachcasper: I can see other sources. Like you have stored the entire catalog in an S3 bucket for example. But I do not see other formats other than Bicep.

architecture/2025-02-user-defined-resource-type-feature-spec.md
```diff
@@ -0,0 +1,1303 @@
...
+
+resource MyCompany.Data/postgreSQL 'System.Resources/resourceTypes@2023-10-01-preview' = {
+  name: 'MyCompany.Data/postgreSQL'
+    version: 'v1alpha1'
+    schema: {
+      properties: {
++       // Required input property
+        size: {
+          type: 'string'
+          enum: ['S', 'M', 'L', 'XL']
```
willdavsmith, line 189: I'm assuming a `required` property is missing here

architecture/2025-02-user-defined-resource-type-feature-spec.md
```diff
@@ -0,0 +1,1303 @@
...
++              description: 'Password for the database user'
+            }
+          }
+        }
+      // Specifying required properties
+      required: ['size']
+    }
+  }
+}
+```
```
superbeeny, line 416: 👍

architecture/2025-02-user-defined-resource-type-feature-spec.md
```diff
@@ -0,0 +1,1303 @@
...
+rad resource-type list
+NAMESPACE                RESOURCE TYPE
+MyCompany.App            service
+MyCompany.Data           postgreSQL
+MyCompany.Net            gateway
+```
+
+```bash
+# List all resource types filtering by namespace
+rad resource-type list --namespace MyCompany.Data
```
superbeeny, line 436: any concerns about confusing this with kubernetes namespaces?

architecture/2025-02-user-defined-resource-type-feature-spec.md
```diff
@@ -0,0 +1,1303 @@
...
+
+1. The postgreSQL resource type is created
+
+**Exceptions**
+
+The operation fails and informs the user interactively if:
+
+1. The resource type already exists
+2. The resource type definition does not properly compile into a `System.Resources/resourceTypes` resource
+3. The API does is in conformance with the OpenAPI v3 specification
```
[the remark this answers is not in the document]
nithyatsu: We will have schema validation on both cli and server side.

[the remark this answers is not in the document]
Reshrahim: this must be the schema defn

architecture/2025-02-user-defined-resource-type-feature-spec.md
```diff
@@ -0,0 +1,1303 @@
...
+- Cloud-provider specific services which use recipes to provision 
+
+## Goals
+
+- Enable platform engineers to define user-defined resource types to meet their organization's unique requirements 
+- Enable Radius community developers to extend Radius resource types independently of the core project 
+- Enable platform engineers and Radius community developers to share their user-defined resource types and associated recipes 
+- Enable developers to use and integrate user-defined resource types into their applications seamlessly
+- Enable specifying user-defined resource types with minimal configuration overhead 
+- Support all core Radius functionality with user-defined resources
```
lakshmimsft, line 37: Can we discuss, is the understanding that this is intended to be a long term goal.

architecture/2025-02-user-defined-resource-type-feature-spec.md
zachcasper: I'm interested in feedback here. But the way I think about it is there are two sides of this conversation. The application side and the infrastructure side. Resource types lie on the application side so we really shouldn't be talking about Kubernetes namespaces that much.

architecture/2025-02-user-defined-resource-type-feature-spec.md
```diff
@@ -0,0 +1,1303 @@
...
+
+Each user story is described with detailed examples. The examples are used to convey the user experience and technical detail. While attention has been paid to technical accuracy such as syntax and structure, expressiveness is prioritized over accuracy. The syntax and structure is expected to change during implementation.
+
+At the conclusion of the user stories, the Feature Summary section lists specific features which are required to fulfill the described user story. These features are an initial decomposition of work derived from the user stories. The features are stack ranked in order of priority resulting in a roadmap of features for user-defined resource types. Therefore, the reader is encouraged to reference the Feature Summary when questioning the priority of a user story—low priority user stories are described side by side with higher priority ones. 
+
+## Definition of Terms
+
+**Resource** – An abstraction representing an application, application component, or cloud resource
```
ytimocin, line 56: Is the environment and application component?

architecture/2025-02-user-defined-resource-type-feature-spec.md
```diff
@@ -0,0 +1,1303 @@
...
+        // Optional input property
+        logging-verbosity: {
+          type: string
+          enum: ['TERSE', 'DEFAULT', 'VERBOSE']
+        }
+        // Output property set by the recipe
+        connection-string: {
+          type: 'string'
+          readOnly: true
++          env-variable: POSTGRESQL_CONNECTION_STRING
```
willdavsmith, line 301: cool!

architecture/2025-02-user-defined-resource-type-feature-spec.md
```diff
@@ -0,0 +1,1303 @@
...
++      // Set environment variables in the container by referencing properties via Bicep
++      env:{
++        ORDERS_DB_CONNECTION_STRING: {
++          value: ordersDB.connectionString
++        }
++        ORDERS_DB_USERNAME: {
++          value: ordersDB.credentias.username
++        }
++        ORDERS_DB_PASSWORD: {
++          value: ordersDB.credentias.password
```
brooke-hamilton, line 261: Typo:
```suggestion
+          value: ordersDB.credentials.username
+        }
+        ORDERS_DB_PASSWORD: {
+          value: ordersDB.credentials.password
```

architecture/2025-02-user-defined-resource-type-feature-spec.md
```diff
@@ -0,0 +1,1303 @@
...
+**Result** 
+
+1. The postgreSQL resource type is created
+
+**Exceptions**
+
+The operation fails and informs the user interactively if:
+
+1. The resource type already exists
+2. The resource type definition does not properly compile into a `System.Resources/resourceTypes` resource
```
willtsai, line 160: what would cause it to no properly compile? syntax issues, etc.?

architecture/2025-02-user-defined-resource-type-feature-spec.md
```diff
@@ -0,0 +1,1303 @@
...
+          }
+        }
+      // Specifying required properties
+      required: ['size']
+    }
+  }
+}
+```
+
+The description property on the resource type should be at least 2 KB, or approximately one page. Properties can be smaller.
```
willdavsmith, line 418: why do we need to specify this? is this something we're including in our docs?

architecture/2025-02-user-defined-resource-type-feature-spec.md
```diff
@@ -0,0 +1,1303 @@
...
+
+**User-defined resource type** – All resource types which are not core resource types
+
+**Sample resource types** – Previously referred to as portable resource types; resource types which ship with Radius as examples of user-defined resource types; includes SQL Server, MongoDB, Redis, RabbitMQ, and Dapr and their respective recipes
+
+**Resource type catalog** – The entire collection of user-defined resource types in a Radius tenant; may include multiple namespaces and resource types
+
+**Resource type namespace** – A logical grouping of resource types; resource type names are unique within a namespace
+
+**Resource provider** – A technical term used only in the implementation of Radius; adopted from Azure resource manager which orchestrates resource lifecycle events with services which actually provide resources (e.g. Azure VMs) whereas Radius user-defined resource types do not have a resource provider; existing references of resource providers in the UI are most likely referring to the resource type namespace
```
ytimocin, line 72: What is the difference between the two? Is `Applications.Core` the namespace (and also the provider) of `environments` resource type?

architecture/2025-02-user-defined-resource-type-feature-spec.md
```diff
@@ -0,0 +1,1303 @@
...
+
+There was early concern about requiring platform engineers to use Bicep given their familiarity is centered on traditional infrastructure as code solutions, namely Terraform. However, feedback from users was that they would prefer to use Bicep because of the additional capabilities it provides, the limitations of using YAML for complex resource types, and the overall ease of use of Bicep.
+
+> [!NOTE]
+>
+> Since YAML has already been implemented, the initial release of user-defined resource types may continue use YAML if that eases the implementation. However, the format should change to Bicep very quickly and support for YAML removed. Radius will only support one file format. This change is reflected in the feature summary at the bottom of this document.
```
brooke-hamilton, line 114: What is the cost of having YAML and Bicep? Is it significant to maintain both if the format is OpenAPI?

architecture/2025-02-user-defined-resource-type-feature-spec.md
```diff
@@ -0,0 +1,1303 @@
...
+
+**Resource type catalog** – The entire collection of user-defined resource types in a Radius tenant; may include multiple namespaces and resource types
+
+**Resource type namespace** – A logical grouping of resource types; resource type names are unique within a namespace
+
+**Resource provider** – A technical term used only in the implementation of Radius; adopted from Azure resource manager which orchestrates resource lifecycle events with services which actually provide resources (e.g. Azure VMs) whereas Radius user-defined resource types do not have a resource provider; existing references of resource providers in the UI are most likely referring to the resource type namespace
+
+> [!WARNING]
+>
+> The term resource provider in the context of Radius is semantically confusing to users. A user-defined resource type inherently will never have a resource provider since the user-defined resource type is built on top of other resource providers (such as Azure VMs). Resource providers will remain in the implementation of Radius but not in the UI.
```
kachawla, line 76: Azure uses it a bit more loosely than just the service that provides resources, the naming convention is used for extensibility through custom resource providers https://learn.microsoft.com/en-us/azure/azure-resource-manager/custom-providers/overview

architecture/2025-02-user-defined-resource-type-feature-spec.md
willtsai: +1 - seems like having the required property within the definition of each property might provide more upfront clarity, albeit at the cost of verbosity.

architecture/2025-02-user-defined-resource-type-feature-spec.md
zachcasper: Applications.Core is the resource type namespace in the UI. Provider is an implementation term.

architecture/2025-02-user-defined-resource-type-feature-spec.md
Reshrahim: The required property is defined following the OpenAPI spec -https://json-schema.org/draft/2020-12/json-schema-validation#name-required

architecture/2025-02-user-defined-resource-type-feature-spec.md
```diff
@@ -0,0 +1,1303 @@
...
+
+**Core resource types** – Primitive resource types which other resource types can build on; includes application, environment, container, gateway, secrets, and volumes
+
+**User-defined resource type** – All resource types which are not core resource types
+
+**Sample resource types** – Previously referred to as portable resource types; resource types which ship with Radius as examples of user-defined resource types; includes SQL Server, MongoDB, Redis, RabbitMQ, and Dapr and their respective recipes
+
+**Resource type catalog** – The entire collection of user-defined resource types in a Radius tenant; may include multiple namespaces and resource types
+
+**Resource type namespace** – A logical grouping of resource types; resource type names are unique within a namespace
```
lakshmimsft, line 70: my understanding is that  resource types are unique within a resource provider. Need to discuss how namespace fits in here

brooke-hamilton (commented):
🚀

architecture/2025-02-user-defined-resource-type-feature-spec.md
```diff
@@ -0,0 +1,1303 @@
...
+# Create the resource type
+rad resource-type create --from-file postgreSQL-resource-type.bicep
+Creating resource type MyCompany.data/postgreSQL
+The resource type MyCompany.Data/postgreSQL has been created
+```
+
+**`postgreSQL-resource-type.bicep`**:
+
+```yaml
+extension radius
```
sk593, line 130: Why do we need the radius extension enabled if we're registering new types?

architecture/2025-02-user-defined-resource-type-feature-spec.md
zachcasper: Oh boy a lot looks familiar here. I didn't know about this, thanks.

architecture/2025-02-user-defined-resource-type-feature-spec.md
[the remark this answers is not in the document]
nicolejms: looks like a typo - assuming you meant AOPI does not conform..

architecture/2025-02-user-defined-resource-type-feature-spec.md
Reshrahim: By core functionality of Radius we mean Recipes, radCLI integration, Application graph and Dashboard etc should work seamlessly with the User-defined-types

architecture/2025-02-user-defined-resource-type-feature-spec.md
```diff
@@ -0,0 +1,1303 @@
...
+      // Specifying required properties
+      required: ['size']
+    }
+  }
+}
+```
+
+**Result**
+
+When a developer creates a postgreSQL resource and a connection to that resource from a container, the environment variables are automatically set in the container.
```
willtsai, line 327: might be worthwhile to think about how to name that `env-variable` property to clearly denote that it's an env var that will be injected into the container that connects to said resource.

architecture/2025-02-user-defined-resource-type-feature-spec.md
```diff
@@ -0,0 +1,1303 @@
...
+
+> [!WARNING]
+>
+> The term resource provider in the context of Radius is semantically confusing to users. A user-defined resource type inherently will never have a resource provider since the user-defined resource type is built on top of other resource providers (such as Azure VMs). Resource providers will remain in the implementation of Radius but not in the UI.
+
+**Recipe** – A Terraform configuration or Bicep template which deploys a resource type within an environment
+
+**Recipe manifest** –  A file containing a collection of resource type to recipe mappings; the mapping includes the location (but not the contents) of the Terraform configuration or Bicep template
+
+**Create** – A Radius API action where a resource gets created in the Radius tenant; i.e. the contents of the request payload is copied into the control plane; e.g., resource types are created
```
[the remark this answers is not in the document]
zachcasper: Probably. This doc essentially ignores updates for a follow-up doc on Day 2 scenarios.

architecture/2025-02-user-defined-resource-type-feature-spec.md
```diff
@@ -0,0 +1,1303 @@
...
+
+**Result** 
+
+1. The postgreSQL resource type is created
+
+**Exceptions**
+
+The operation fails and informs the user interactively if:
+
+1. The resource type already exists
```
ytimocin, line 159: So, we don't allow an update on the same resource type. The user will have to delete an existing one and then create again if there is an error in the definition.

architecture/2025-02-user-defined-resource-type-feature-spec.md
zachcasper: I was thinking about it from a user perspective. If they start with YAML because it's more familiar, they will quickly hit a dead end in terms of functionality. Plus we would have to explain when to use YAML and when to use Bicep. I'd prefer to avoid all of that. It's just not worth the extra effort two support two.

architecture/2025-02-user-defined-resource-type-feature-spec.md
```diff
@@ -0,0 +1,1303 @@
...
+```
+
+ **Result**
+
+1. Radius confirms the Terraform configuration file is accessible
+1. Recipe is registered in the environment
+
+**Exceptions**
+
+The operation fails and informs the user interactively if:
```
willtsai, line 592: would it also fail if the Recipe does not include all the properties required by the UDT?

architecture/2025-02-user-defined-resource-type-feature-spec.md
ytimocin: I believe that reading from a path like a GitHub repository could be helpful. I can imagine an organization having a repo just for the resource types.

architecture/2025-02-user-defined-resource-type-feature-spec.md
```diff
@@ -0,0 +1,1303 @@
...
+}
+```
+
+**Result**
+
+The environment variables set specified by the developer are set in the container.
+
+**User Experience 2 – Automatically Injected by platform engineer** 
+
+The platform engineer can specify default environment variables which will automatically be injected into a container when a connection is established. This is similar to existing functionality for portable resource types in Radius today (see the [Redis type](https://docs.radapp.io/reference/resource-schema/cache/redis/#environment-variables-for-connections) for example). Note that user experience 1 and 2 can be used side by side.
```
kachawla, line 274: > This is similar to existing functionality for portable resource types in Radius today

Just to be explicit, the key difference will be that the env variable keys will be user defined (example ` POSTGRESQL_CONNECTION_STRING` below) rather than a standard convention decided by Radius.

architecture/2025-02-user-defined-resource-type-feature-spec.md
```diff
@@ -0,0 +1,1303 @@
...
+  --resource-type MyCompany.Data/postgreSQL \
+  --template-kind terraform \
+  --template-path postgreSQL.tf
+Registering recipe for MyCompany.Data/postgreSQL
+The recipe for MyCompany.Data/postgreSQL is registered in the my-env environment
+```
+
+> [!IMPORTANT]
+>
+> See Other Changes for changes related to the `rad recipe register` command.
```
kachawla, line 519: Could you please call them out? I'm unable to see any difference :/

architecture/2025-02-user-defined-resource-type-feature-spec.md
willdavsmith: `System.Resources/resourceTypes` is still a Radius type right?

architecture/2025-02-user-defined-resource-type-feature-spec/image-20250130153426974.png
ytimocin, line 1: nit: typo: ingress

architecture/2025-02-user-defined-resource-type-feature-spec/image-20250205155136117.png
ytimocin, line 1: very cool

architecture/2025-02-user-defined-resource-type-feature-spec.md
willdavsmith: oh I see this is a note for the developer not to trim the response or something.

architecture/2025-02-user-defined-resource-type-feature-spec.md
Reshrahim: Yes

architecture/2025-02-user-defined-resource-type-feature-spec.md
Reshrahim: Update story as part of Day 2 experience needs to be flushed more. What we allow depends on if it is breaking or non-breaking change. The user experience should be in a way that they should version the changes so that there is less disruption for the downstream dependencies like recipes and a running resource from the type.

architecture/2025-02-user-defined-resource-type-feature-spec.md
Reshrahim: yeah I think it should fail if the Recipes doesn't output the required properties as outputs

architecture/2025-02-user-defined-resource-type-feature-spec.md
kachawla: I was assuming this to be Kubernetes namespace until I saw this comment. Sounds obvious after reading this.. but not necessarily at first

architecture/2025-02-user-defined-resource-type-feature-spec.md
zachcasper: Changed to connected-container-env-var

architecture/2025-02-user-defined-resource-type-feature-spec.md
```diff
@@ -0,0 +1,1361 @@
...
+            username: {
+              type: 'string'
+            }
+            password: {
+              type: 'string'
+            }
+          }
+        }
++     // Specifying required properties
++      required: ['size']
```
lakshmimsft, line 211: can this be included within the property declaration:
```
size: {
          type: 'string'
          enum: ['S', 'M', 'L', 'XL']
          required: true
          ...
          }
```

architecture/2025-02-user-defined-resource-type-feature-spec.md
```diff
@@ -0,0 +1,1361 @@
...
+
+Almost all applications connect to other systems which are managed independently. These systems could be other applications within the organization or software as a service applications managed by other vendors. These dependencies present challenges in a complex, interconnected organization. The Radius application graph is designed to help managed these dependencies by documenting connections between application components and visualizing them via the Radius dashboard and API. However, today, the application graph only supports connections between Radius-managed components. With user-defined resource types, organizations can represent application dependencies which are external to the application, environment, or organization.
+
+Radius already restricts the ability to create resources which do not have an associated recipe in the target environment. If a resource is created in an environment without a recipe for that resource type, it will fail with an error stating that no recipe was found for the corresponding resource type. Not having a recipe for a resource type in an environment is how the platform engineer controls which resources can be deployed in which environment.
```
ytimocin, line 605: Not sure if I understand this correctly. We have two types of resource provisioning: **recipe** and **manual**. I don't think we need to have a recipe for a specific resource type to have an instance of that resource type in a given environment.

architecture/2025-02-user-defined-resource-type-feature-spec.md
zachcasper: That would be better in my mind. I believe @Reshrahim said this was an OpenAPI thing. If we can do what you propose, I prefer that.

architecture/2025-02-user-defined-resource-type-feature-spec.md
```diff
@@ -0,0 +1,1361 @@
...
+          properties: {
+            username: {
+              type: 'string'
+              description: 'Username for the external service'
+              connected-resource-environment-variable: EXTERNAL_SERVICE_USERNAME
+            }
+            password: {
+              type: 'string'
+              description: 'Password for the external service user'
+              connected-resource-environment-variable: EXTERNAL_SERVICE_PASSSWORD
```
ytimocin, line 652: nit:
```suggestion
              connected-resource-environment-variable: EXTERNAL_SERVICE_PASSWORD
```

architecture/2025-02-user-defined-resource-type-feature-spec.md
```diff
@@ -0,0 +1,1361 @@
...
+
+resource twilio 'MyCompany.App/externalService@v1alpha1' = {
+  name: 'twilio'
+  properties: {
+    environment: production
+    connection-string 'https://api.twilio.com/2010-04-01/Accounts/${twilio-account-sid}'
+    credentials:
+      username: ${twilio-username}
+      password: ${twilio-password}
+}
```
ytimocin, line 688: By default, the resourceProvisioning property gets set to `recipe`. But you can also set it to `manual`: https://github.com/radius-project/radius/blob/main/test/testrecipes/test-bicep-recipes/resource-creation.bicep#L14. This is just an example.

architecture/2025-02-user-defined-resource-type-feature-spec.md
```diff
@@ -0,0 +1,1361 @@
...
+        ...
+      }
+    }
+  }
+}
+```
+
+When a developer creates a `MyCompany.App/service` resource, they will not need to have any awareness that an Envoy proxy is also created. This abstraction enables the platform engineer to easily inject sidecar containers. Envoy is shown in the example here, but other common sidecars are container security scanning tools such as Aqua Security and observability tools such as OpenTelemetry or Datadog. 
+
+The developer will, however, see that Envoy is deployed when they inspect the application graph since the child resources are standard Radius resources and will appear in the application graph.
```
ytimocin, line 782: Does Radius support child resources as of now? Is there an example Radius Bicep to that?

architecture/2025-02-user-defined-resource-type-feature-spec.md
```diff
@@ -0,0 +1,1361 @@
...
+
+var twilio-account-sid = 'ACXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX'
+var twilio-username = 'twilio-prod-user'
+var twilio-password = 'o84nouvTiHWiw97sbq6B'
+
+resource twilio 'MyCompany.App/externalService@v1alpha1' = {
+  name: 'twilio'
+  properties: {
+    environment: production
+    connection-string 'https://api.twilio.com/2010-04-01/Accounts/${twilio-account-sid}'
```
nithyatsu, line 684: these would have the connected-resource-environment-variable attribute too correct?

architecture/2025-02-user-defined-resource-type-feature-spec.md
zachcasper: That's true for portable resource types only right? Are you suggesting we reuse the `resourceProvisioning: 'manual'` property for UDT? If so, that seems like a valid approach to have consistency between the two.

architecture/2025-02-user-defined-resource-type-feature-spec.md
zachcasper: Yes, these could have those env var specified here as well.

architecture/2025-02-user-defined-resource-type-feature-spec.md
lakshmimsft: this (explicit `manual` declaration) is not needed since UDTs that support recipes will Always be deployed with recipes. Pls confirm

architecture/2025-02-user-defined-resource-type-feature-spec.md
```diff
@@ -0,0 +1,1361 @@
...
+```yaml
+resource MyCompany.App/externalService 'System.Resources/resourceTypes@2023-10-01-preview' = {
+  name: 'MyCompany.App/externalService'
+  description: '''
+    The external service resource type represents a resource which
+    is not managed by Radius but appears in the application graph.
+  '''
+  // If true (default) allow recipes to be registered to deploy resources
+  // If false, resource can be created in any environment
+  allowRecipes: false
```
nithyatsu, line 631: downside here, is that a user cannot switch from production to qa and expect to deploy the same bicep and automatically connect to a staging twilio  url. Am I understanding this correctly?

architecture/2025-02-user-defined-resource-type-feature-spec.md
```diff
@@ -0,0 +1,1361 @@
...
+
+Radius will continue to ship with core resource types including containers, auto-scalers, gateways, volumes, and secrets. It will have built-in logic for deploying these resources to various container platforms including Kubernetes, ECS, ACA, ACI, and CloudRun. In order to allow the creation of more abstract, application-oriented resource types, user-defined resource types will support the concept of child resources.
+
+Child resources are resources which get deployed when its parent resource is deployed. The lifecycle of the child resource is tied to that of the parent. These child resources can be a mix of core resource types and user-defined resource types. 
+
```
lakshmimsft, line 729: We will support one layer of child resources going ahead. do we anticipate further nesting? children of children??

zachcasper: I would expect one layer of children within a single resource type definition. But that child could have another child within its definition. So syntactically, one level of children, but logically, many. But the that multi-layer nesting would be in the future. Single layer is all that's needed in the beginning.

architecture/2025-02-user-defined-resource-type-feature-spec.md
ytimocin: I mean, I think UDTs that support recipes should also allow manual provisioning too. `SupportsRecipes` should be an extra capability on top of the default capabilities.

**_Note: This is just my opinion :)_**

architecture/2025-02-user-defined-resource-type-feature-spec.md
lakshmimsft: okay, I'll defer to @nithyatsu , @Reshrahim who've gone thru OpenAPI spec.

architecture/2025-02-user-defined-resource-type-feature-spec.md
```diff
@@ -0,0 +1,1204 @@
...
+- Enable platform engineers to define user-defined resource types to meet their organization's unique requirements 
+- Enable Radius community developers to extend Radius resource types independently of the core project 
+- Enable platform engineers and Radius community developers to share their user-defined resource types and associated recipes 
+- Enable developers to use and integrate user-defined resource types into their applications seamlessly
+- Enable specifying user-defined resource types with minimal configuration overhead 
+- Support all core Radius functionality with user-defined resources 
+
+## Non-Goals (out of scope)
+
+* TODO Configuring and keeping up to date Radius and Bicep tooling on developer workstations
```
Reshrahim, line 55: I think we want to optimize the experience on the Bicep extensions for user-defined resource types given the friction with users. Should this be a goal?

architecture/2025-02-user-defined-resource-type-feature-spec.md
```diff
@@ -0,0 +1,1204 @@
...
+* Modifying, versioning, and deprecation of resource types
+* Deployment of updates to resources
+* Radius-maintained and tested sample repository
```
Reshrahim, line 58: if not a fully automated repository, I think we would still require a basic repository for users to share and consume user-defined resource types and Recipes?

architecture/2025-02-user-defined-resource-type-feature-spec.md
```diff
@@ -0,0 +1,1204 @@
...
+
+As a platform engineer, I need to create a resource type in Radius. I want to define typed properties for my developers to use in their application definitions and include documentation and examples.
+
+**Summary**
+
+The platform engineer will create a resource type in their Radius tenant using the CLI. The resource type will have a namespace and an OpenAPI schema. 
+
+> [!NOTE]
+>
+> A previous version of this document proposed defining resource types using Bicep. After the initial launch of user-defined resource types, this proposal has been withdrawn for the following reasons:
```
Reshrahim, line 118: I think we also had feedback from our users on being able to do validation and having conditional functions on the resource type definitions which took us through the Bicep route. How are we handling that?

Reshrahim: I do see the solution for conditional properties under user story 2. May be a note here would be helpful.

architecture/2025-02-user-defined-resource-type-feature-spec.md
```diff
@@ -0,0 +1,1204 @@
...
+      - 'L': 2.0 vCPU, 8 GiB memory, 60 GiB storage
+      - 'XL': 4.0 vCPU, 16 GiB memory, 100 GiB storage
+
+READ-ONLY PROPERTIES
+  - host (string) Fully qualified host name of the database server
+  - port (integer) The port number of the database server
+  - credentials.username (string) The username for the database
+  - credentials.password (string) The password for the database user
+
+CONNECTED CONTAINER ENVIRONMENT VARIABLES
```
Reshrahim, line 454: nit: This makes sense in the context of this example but I think the name could be generic as functions can also connect to user-defined resource types. May be "Environment Variables for Connection"

architecture/2025-02-user-defined-resource-type-feature-spec.md
```diff
@@ -0,0 +1,1204 @@
...
+
+1. Compile a Bicep extension (this is performed first to catch compilation errors early)
+2. Create or update the resource type in Radius
+3. Store the Bicep extension within the Radius control plane which has the appropriate enhancements to support Bicep extensions
+
+When the resource type is deleted, the associated Bicep extension is deleted as well.
+
+When the developer executes rad CLI commands Radius will:
+
+* `rad workspace create` – Creating a new workspace will create a bicepconfig.json in the working directory which has a reference to each Radius-hosted Bicep extension (if bicepconfig.json exists in the working directory, update the file to ensure it has each of the Radius-hosted Bicep extensions)
```
Reshrahim, line 506: `rad workspace` would pull the Bicep extensions from the k8s context?

architecture/2025-02-user-defined-resource-type-feature-spec.md
```diff
@@ -0,0 +1,1204 @@
...
+            required:
+            - environment
+            - database
+```
+
+**Result**
+
+When a developer creates a postgreSQL resource and a connection to that resource from a container, the environment variables are automatically set in the container.
+
+### User Story 10 – UDT→UDT
```
Reshrahim, line 861:
```suggestion
### User Story 9 – UDT→UDT
```

architecture/2025-02-user-defined-resource-type-feature-spec.md
zachcasper: I added some user feedback which explains why we originally thought Bicep.

architecture/2025-02-user-defined-resource-type-feature-spec.md
zachcasper: An environment variable only makes sense in the context of a container. We explicitly mean containers, not other types like functions. This is containers in general, not Applications.Core/containers.

architecture/2025-02-user-defined-resource-type-feature-spec.md
zachcasper: It would pull it from the Radius control plane, which is referenced via a Kubernetes context.

architecture/2025-02-user-defined-resource-type-feature-spec.md
```diff
@@ -0,0 +1,1212 @@
...
+> However, Kubernetes and most cloud-native projects follow [RFC 1123](https://tools.ietf.org/html/rfc1123) or [RFC 1035](https://tools.ietf.org/html/rfc1035) for object naming which enforce:
+>
+> * contain at most 63 characters
+> * contain only lowercase alphanumeric characters or '-'
+> * start with an alphabetic character
+> * end with an alphanumeric character
+>
+> Complicating enforcing a naming convention is that Terraform uses an underscore for resource and variable names (e.g., `resource "aws_instance"` and `variable "web_instance_count"`).
+>
+> Given Radius is cloud provider-agnostic, the current restriction will be removed and replaced with RFC 1035. Underscores will not be allowed as it is common practice for Terraform users to substitute underscores for hyphens.
```
ytimocin, line 267: This will not be applied to Core types, right?

architecture/2025-02-user-defined-resource-type-feature-spec.md
```diff
@@ -0,0 +1,1212 @@
...
+              description: "Optional: When true, provisions an L7 ingress gateway mapped to the port named http"
+            hostname:
+              type: string
+              description: "Required when ingress is true: Set the hostname for the service"
+            container:
+              ...
+          required:
+          - environment
+          - container
++          if:
```
ytimocin, line 309: We need a good validator for the resource-type yaml files.

architecture/2025-02-user-defined-resource-type-feature-spec.md
```diff
@@ -0,0 +1,1212 @@
...
+> * We have already prototyped the ability to use Radius resource types in a Terraform recipe via a new Radius provider for Terraform.
+> * There is a need to support TypeSpec for resource type definition. Implementing user-defined resource types in Bicep complicates this objective because it introduces the ability to build functionality into the type definition rather than into the recipe. By modeling user-defined resource types only in YAML and TypeSpec, there is a clear separation of interface and implementation.
+> * Long-term, we envision Radius being polyglot rather than purely Bicep. Not only will Radius support multiple IaC solutions, but could also have language specific SDKs for developers. Imagine a cloud development kit with support for Python, Java, Node, etc. which was built on resource types defined by the platform engineer.
+> * YAML has already been implemented and is getting positive feedback from early users.
+>
+> Thew original proposal for Bicep was based on this user feedback:
```
willtsai, line 123:
```suggestion
> The original proposal for Bicep was based on this user feedback:
```

architecture/2025-02-user-defined-resource-type-feature-spec.md
```diff
@@ -0,0 +1,1212 @@
...
+`````
+
+**User Experience 1 – Using  Terraform configuration**
+
+**`/recipes/webservice/main.tf`**
+
+```
+terraform {
+  required_providers {
+    radius = {
```
ytimocin, line 567: Interesting :) We are writing a Terraform Radius Provider then :)

architecture/2025-02-user-defined-resource-type-feature-spec.md
```diff
@@ -0,0 +1,1212 @@
...
+Radius.Resources         webService
+Radius.Resources         postgreSQL
+```
+
+```bash
+# Show details of a resource type
+rad resource-type show Radius.Resources/postgreSQL
+NAMESPACE                Radius.Resources
+RESOURCE TYPE            postgreSQL
+VERSION                  2025-05-05
```
nithyatsu, line 428: This should perhaps be DEFAULT VERSION

architecture/2025-02-user-defined-resource-type-feature-spec.md
```diff
@@ -0,0 +1,1212 @@
...
+  }
+}
+
+variable "context" {
+  description = "This variable contains Radius recipe context."
+  type = any
+}
+
++ if var.context.resource.properties.ingress {
++  resource "applications_core_gateways" "gateway" {
```
ytimocin, line 679: How would we create UDT resources using Terraform Radius Provider? Each time a new resource type is registered, we would need to update the provider somehow.

architecture/2025-02-user-defined-resource-type-feature-spec.md
```diff
@@ -0,0 +1,1212 @@
...
+```bash
+# Register the PostgreSQL Terraform recipe in the myEnv environment
+rad recipe register \
+  --environment myEnv \
+  --resource-type Radius.Resources/postgreSQL \
+  --recipe-kind terraform \
+  --recipe-path git://github.com/<username>/repo.git//recipes/postgreSQL
+Registering recipe for Radius.Resources/postgreSQL
+The recipe for Radius.Resources/postgreSQL is registered in the myEnv environment
+```
```
ytimocin, line 754: Is there any sort of validation in this flow? Like the validation of the recipe in the given path or at least if there is a recipe in the given path?

architecture/2025-02-user-defined-resource-type-feature-spec.md
```diff
@@ -0,0 +1,1212 @@
...
+>
+> - **Conditional resources.** Building on child resources, provisioning resources based on the value of developer-specified properties was a use case identified. In the web service example, they want to have a single web service resource type which has a boolean specifying whether an ingress gateway is provisioned or not. (More concretely, a boolean to control whether the Kubernetes service is of type cluster IP or load balancer.) 
+>   - This feature spec demonstrates how to do conditional resources via recipes.
+>
+> - **Data manipulation and validation functions**. For example, one user wanted to be able to concatenate two properties which gets passed to the recipe. They need data validation functions with property specific error messages when validation does not pass. Regular expressions were proposed but their feedback was regex may be too complex. 
+>   - One advantage of Bicep-based resource type definitions was bringing the data validation earlier in the deployment. This type of validation is still possible in a Bicep and Terraform recipe.
+
```
willtsai, line 133: What would the guidance be for users when it comes to defining custom Radius resource types? Would it be:
- If they are trying to define simple or non-composite types, use YAML/TypeSpec
- Else if they are trying to define composite types, use Bicep/Terraform by way of Radius Recipe

architecture/2025-02-user-defined-resource-type-feature-spec.md
```diff
@@ -0,0 +1,1212 @@
...
+Furthermore, once portable resource types are refactored into user-defined resource types, the ability to specify a recipe name will be retired. The command will change from ``rad recipe register <recipeName>` to `rad recipe register`. This also has the benefit of not having to build developer documentation for discovering available recipes. 
+
+**Rename of template-kind and template-path** – Radius is not opinionated about which infrastructure as code solution to use. Therefore, it is important that Radius is not biased towards one solution versus another. The parameter `--template-kind` and `--template-path` are biased towards Bicep as Bicep files are called templates and Terraform files are configurations. 
+
+Given this, the `--template-kind` and `--template-path` arguments will be renamed `--recipe-kind` and `--recipe-path`. The previous commands will be retained for backwards compatibility.
```
ytimocin, line 764: 👍

architecture/2025-02-user-defined-resource-type-feature-spec.md
```diff
@@ -0,0 +1,1212 @@
...
+>
+> - **Conditional resources.** Building on child resources, provisioning resources based on the value of developer-specified properties was a use case identified. In the web service example, they want to have a single web service resource type which has a boolean specifying whether an ingress gateway is provisioned or not. (More concretely, a boolean to control whether the Kubernetes service is of type cluster IP or load balancer.) 
+>   - This feature spec demonstrates how to do conditional resources via recipes.
+>
+> - **Data manipulation and validation functions**. For example, one user wanted to be able to concatenate two properties which gets passed to the recipe. They need data validation functions with property specific error messages when validation does not pass. Regular expressions were proposed but their feedback was regex may be too complex. 
+>   - One advantage of Bicep-based resource type definitions was bringing the data validation earlier in the deployment. This type of validation is still possible in a Bicep and Terraform recipe.
```
willtsai, line 132: so if users choose to define a resource type using YAML/TypeScript, then they forgo the ability to do early data validation?

architecture/2025-02-user-defined-resource-type-feature-spec.md
```diff
@@ -0,0 +1,1212 @@
...
+
+* Required properties using the `required` property
+* Read only properties which are set by the recipe denoted with `readOnly: true`
+* String and integer types; Radius also needs to support arrays
+* Enumerations as shown in the `size` property
+* Nested properties as demonstrated with the credentials property
+
+> [!NOTE]
+>
+> Today's implementation is very resource provider-centric. The behavior of `rad resource-type create` for example is different dependent upon if the resource provider exists or not and users are notified about resource providers being created. Since users are defining types and not providers, they have no concept of what a resource provider is. The implementation will change to remove resource providers from the user experience.
```
lakshmimsft, line 248: Let's review what changes are needed. My interpretation of this is that the user experience will change but the underlying constructs still exist.

architecture/2025-02-user-defined-resource-type-feature-spec.md
```diff
@@ -0,0 +1,1212 @@
...
+
+**Summary**
+
+The goal of user-defined resource types is to enable users and the broader Radius community to build a library of resource types which model cloud-native application components rather than cloud infrastructure resources. Some organizations will choose to model their resources with granular resource types such as containers and databases, while others will only offer abstracted resource types such as a microservice in their resource catalog. These abstract resource types will be composed of many resources just an an application is composed of many resources. 
+
+A microservice, for example, could be composed of an application container, an in-memory cache, a reverse proxy, and a container security sidecar. In order to model the microservice resource type, the platform engineer needs the ability to specify a schema for the resource type, as well as the resources that compose the microservice. What makes this user story unique compared to those previously discussed is that the resources composing this microservice are other Radius resource types—not only infrastructure. 
+
+**`webservice-resource-type.bicep`**:
```
willtsai, line 530:
```suggestion
**`webservice-resource-type.yaml`**:
```

architecture/2025-02-user-defined-resource-type-feature-spec.md
```diff
@@ -0,0 +1,1212 @@
...
+  --recipe-path git://github.com/<username>/repo.git//recipes/postgreSQL
+Registering recipe for Radius.Resources/postgreSQL
+The recipe for Radius.Resources/postgreSQL is registered in the myEnv environment
+```
+
+There are two changes here:
+
+**Retirement of named recipes** – Today, Radius supports named recipes. When deploying a portable resource type, developers can override the default recipe by specifying a recipe name in their application definition. This enables developers to punch through the separation of concerns between platform and developers. Based on strong user feedback, user-defined resource types will not implement the ability to punch through and use alternate recipes. 
+
+Furthermore, once portable resource types are refactored into user-defined resource types, the ability to specify a recipe name will be retired. The command will change from ``rad recipe register <recipeName>` to `rad recipe register`. This also has the benefit of not having to build developer documentation for discovering available recipes.
```
willdavsmith, line 760:
```suggestion
Furthermore, once portable resource types are refactored into user-defined resource types, the ability to specify a recipe name will be retired. The command will change from `rad recipe register <recipeName>` to `rad recipe register`. This also has the benefit of not having to build developer documentation for discovering available recipes. 
```

architecture/2025-02-user-defined-resource-type-feature-spec.md
```diff
@@ -0,0 +1,1212 @@
...
+
+**Summary**
+
+The goal of user-defined resource types is to enable users and the broader Radius community to build a library of resource types which model cloud-native application components rather than cloud infrastructure resources. Some organizations will choose to model their resources with granular resource types such as containers and databases, while others will only offer abstracted resource types such as a microservice in their resource catalog. These abstract resource types will be composed of many resources just an an application is composed of many resources. 
+
+A microservice, for example, could be composed of an application container, an in-memory cache, a reverse proxy, and a container security sidecar. In order to model the microservice resource type, the platform engineer needs the ability to specify a schema for the resource type, as well as the resources that compose the microservice. What makes this user story unique compared to those previously discussed is that the resources composing this microservice are other Radius resource types—not only infrastructure. 
+
+**`webservice-resource-type.bicep`**:
+
+`````yaml
```
willtsai, line 532: is this illustrating that composite types can be defined using YAML? would we recommend users against this and point them to defining composite types via Bicep/Terraform Recipes?

architecture/2025-02-user-defined-resource-type-feature-spec.md
zachcasper: That makes sense. Do we have the concept of a default version? I haven't been able to create multiple versions. I can only do 2023-05-01-preview.

architecture/2025-02-user-defined-resource-type-feature-spec.md
```diff
@@ -0,0 +1,1212 @@
...
+
+### **User Story 6 – Conditional resources** 
+
+As a platform engineer, I need to write a recipe which creates additional resources dependent upon the properties set by the developer.
+
+**Summary**
+
+The platform engineer adds a boolean property to the `webservice` resource type from the previous user story.
+
+**`service-resource-type.bicep`**:
```
willtsai, line 637:
```suggestion
**`service-resource-type.yaml`**:
```

architecture/2025-02-user-defined-resource-type-feature-spec.md
```diff
@@ -0,0 +1,1212 @@
...
+// Shared resource for all applications representing an external service
+resource jira 'Radius.Resources/externalService@2023-10-01-preview' = {
+  name: 'jira'
+  properties:{
+    connectionString: 'https://api.atlassian.com/ex/jira/<cloudId>/rest/api/3/<resource-name>'
+    credentials: {
+      type: 'apiKey'
+      apiKey: 'your-api-key'
+    }
+    environment: nimble.id
```
willdavsmith, line 938:
```suggestion
    environment: env.id 
```

architecture/2025-02-user-defined-resource-type-feature-spec.md
zachcasper: My intuition is that large enterprises would create a resource type for everything so they retain control on the schema.

architecture/2025-02-user-defined-resource-type-feature-spec.md
```diff
@@ -0,0 +1,1212 @@
...
+          required:
+          - environment
+          - container
++          if:
++            properties:
++              ingress:
++                const: true
++          then:
++            required: hostname
+```
```
lakshmimsft, line 315: Concern on this. if/then is mostly not recognised by OpenAPI3.0 spec. Something similar may be achieved perhaps with (need to verify): 
```
      # …
      required:
        - environment
        - container
      anyOf:
        - properties:
            ingress:
              const: false
          # when ingress is false, no extra requirement
        - properties:
            ingress:
              const: true
          required:
            - hostname
       ```

File: architecture/2025-02-user-defined-resource-type-feature-spec.md (hunk 47 above)
Comment (zachcasper) [reply architecture/2025-02-user-defined-resource-type-feature-spec.md:132]: Early is relative here. They would be able to validate whatever Bicep validation has. The recipe may do further data validation further down the deployment pipeline.

File: architecture/2025-02-user-defined-resource-type-feature-spec.md (hunk 46 above)
Comment (kachawla) [inline architecture/2025-02-user-defined-resource-type-feature-spec.md:764]: > The previous commands will be retained for backwards compatibility.

Do we eventually plan to phase out older version?

File: architecture/2025-02-user-defined-resource-type-feature-spec.md (hunk 50 above)
Comment (willtsai) [inline architecture/2025-02-user-defined-resource-type-feature-spec.md:760]:
```suggestion
Furthermore, once portable resource types are refactored into user-defined resource types, the ability to specify a recipe name will be retired. The command will change from `rad recipe register \<recipeName\>` to `rad recipe register`. This also has the benefit of not having to build developer documentation for discovering available recipes. 
```

architecture/2025-02-user-defined-resource-type-feature-spec.md
zachcasper: Agree, this statement is focused on the UI mostly. There are some UX things like https://github.com/radius-project/radius/issues/9354#issuecomment-2852978086

architecture/2025-02-user-defined-resource-type-feature-spec.md
```diff
@@ -0,0 +1,1212 @@
...
+// webservice resource, not core container
+resource frontend 'Radius.Resources/webServices@2025-05-05' = {
+  name: 'frontend'
+  properties: {
+    application: myWebApp.id
+    environment: environment
+    connection: {
+      jira:
+        source: jira.id
+    }
```
kachawla, line 910: What functionality does this block enable for users when it's defined?

architecture/2025-02-user-defined-resource-type-feature-spec.md
willtsai, line 760: I worry that taking away the ability to punch through will cause issues for users down the line who want to use this functionality in ways we don't fully understand yet

architecture/2025-02-user-defined-resource-type-feature-spec.md
nithyatsu: Would we not want the created types to be shareable across the community? If so, would we not want to support some kind of registry?

nithyatsu: Many users are not familiar with the workspace-kubecontext-env combinations we offer and it has been a little confusing for the end user. Not sure if we want to overload it further. Might need some discussions to figure if there is a better construct.

architecture/2025-02-user-defined-resource-type-feature-spec.md
```diff
@@ -0,0 +1,1212 @@
...
+>
+> Given Radius is cloud provider-agnostic, the current restriction will be removed and replaced with RFC 1035. Underscores will not be allowed as it is common practice for Terraform users to substitute underscores for hyphens.
+>
+> However, all examples in the documentation will conform to the more restrictive ARM standard to encourage consistency.
+
+> [!CAUTION]
+>
+> Today, Radius enforces the ARM API versioning standard. From the [source code](https://github.com/radius-project/radius/blob/main/pkg/cli/manifest/validation.go#L33) `An API version must be a date in YYYY-MM-DD format, and may optionally have the suffix '-preview'. Example: 2025-01-01"`. However, Kubernetes and most cloud-native project follow a [different versioning scheme](https://kubernetes.io/docs/reference/using-api/). The [Terraform versioning scheme](https://developer.hashicorp.com/terraform/plugin/best-practices/versioning) is similar to Kubernetes. Because Radius is cloud provider-agnostic, Radius will not enforce a specific API versioning scheme on the user.
+>
```
nithyatsu, line 274: It would be good to confirm this. Since UCP has a long term vision of replacing ARM, we might still want radius versioning to be compliant with ARM.

architecture/2025-02-user-defined-resource-type-feature-spec.md
nithyatsu: We do not have support for versioning yet. But once that comes in, all radius types will have older versions + one default with is the latest on that release.

architecture/2025-02-user-defined-resource-type-feature-spec.md
nithyatsu: We authored schema design based on openapi 3.0 which brings in structural schema.  3.1 is not widely adopted in field.

architecture/2025-02-user-defined-resource-type-feature-spec.md
```diff
@@ -0,0 +1,1212 @@
...
+    environment: environment
+  }
+}
+
+// Existing resource in the resource group
+resource jira 'Radius.Resources/externalService@2023-10-01-preview' = existing {
+  name: 'jira'
+}
+
+// webservice resource, not core container
```
nithyatsu, line 901: If webservices deploys multiple containers front-end, reverse-proxy and security scanner, 
how would we know which container to set the environement variable on?

architecture/2025-02-user-defined-resource-type-feature-spec.md
```diff
@@ -0,0 +1,1405 @@
...
+        schema: 
+          type: object
+          properties:
+            environment:
+              type: string
+              description: "Required: The Radius environment; typically set by the rad CLI"
+            application:
+              type: string
+              description: "Optional: The application which the resource is associated with"
+            connectionString:
```
nithyatsu, line 964: This are outputs of recipe, and marked readonly, correct?

architecture/2025-02-user-defined-resource-type-feature-spec.md
```diff
@@ -0,0 +1,1055 @@
...
+}
+```
+
+The following environment variables are automatically created in the container:
+
+| Environment Variable Name                              | Property Name                              |
+| ------------------------------------------------------ | ------------------------------------------ |
+| `CONNECTIONS_ORDERSDB_PROPERTIES_ENVIRONMENT`          | `ordersDB.properties.environment`          |
+| `CONNECTIONS_ORDERSDB_PROPERTIES_APPLICATION`          | `ordersDB.properties.application`          |
+| `CONNECTIONS_ORDERSDB_PROPERTIES_DATABASE`             | `ordersDB.properties.database`             |
```
nithyatsu, line 784: we decided to leave out env id and app id right?

architecture/2025-02-user-defined-resource-type-feature-spec.md
```diff
@@ -0,0 +1,1055 @@
...
+
+### User Story 8 – Connections into Applications.Core/containers
+
+As a developer, I need to read properties from a connected resource using environment variables injected into my container.
+
+**Priority:** p0 – Critical (core functionality)
+
+**Summary**
+
+All resource in Radius will support connections to other resources, no matter the type. When a connection exists from a connecting resource (destination) to another connected resource (source), the source resource's properties will be available within the destination resource. In the case of `Applications.Core/containers` these properties are manifested as environment variables.
```
brooke-hamilton, line 763: Nit
```suggestion
All resources in Radius will support connections to other resources, no matter the type. When a connection exists from a connecting resource (destination) to another connected resource (source), the source resource's properties will be available within the destination resource. In the case of `Applications.Core/containers` these properties are manifested as environment variables.
```

brooke-hamilton (changes requested):
🚀

architecture/2025-02-user-defined-resource-type-feature-spec.md
```diff
@@ -0,0 +1,1055 @@
...
+| `CONNECTIONS_ORDERSDB_PROPERTIES_DATABASE`             | `ordersDB.properties.database`             |
+| `CONNECTIONS_ORDERSDB_PROPERTIES_SIZE`                 | `ordersDB.properties.size`                 |
+| `CONNECTIONS_ORDERSDB_PROPERTIES_HOST`                 | `ordersDB.properties.host`                 |
+| `CONNECTIONS_ORDERSDB_PROPERTIES_PORT`                 | `ordersDB.properties.port`                 |
+| `CONNECTIONS_ORDERSDB_PROPERTIES_CREDENTIALS_USERNAME` | `ordersDB.properties.credentials.username` |
+| `CONNECTIONS_ORDERSDB_PROPERTIES_CREDENTIALS_PASSWORD` | `ordersDB.properties.credentials.password` |
+
+If the developer prefers for these environment variables to not be created, he/she can set the existing `disableDefaultEnvVars` boolean property to TRUE.
+
+### User Story 9 – Connections into recipes
```
kachawla, line 793: Do we have a third scenario for UDT to UDT connections, or does this cover that?

architecture/2025-02-user-defined-resource-type-feature-spec.md
```diff
@@ -0,0 +1,1055 @@
...
+```yaml
+connections: {
+  ordersDB:
+    source: ordersDB.id
+}
+```
+
+The following environment variables are automatically created in the container:
+
+| Environment Variable Name                              | Property Name                              |
```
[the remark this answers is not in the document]
zachcasper: `CONNECTIONS_ORDERSDB_PROPERTIES_CREDENTIALS_USERNAME` is an example.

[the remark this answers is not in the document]
zachcasper: For arrays, I believe in OpenAPI every item has a name correct?

architecture/2025-02-user-defined-resource-type-feature-spec.md
```diff
@@ -0,0 +1,1055 @@
...
+| ------------------------------------------------------ | ------------------------------------------ |
+| `CONNECTIONS_ORDERSDB_PROPERTIES_ENVIRONMENT`          | `ordersDB.properties.environment`          |
+| `CONNECTIONS_ORDERSDB_PROPERTIES_APPLICATION`          | `ordersDB.properties.application`          |
+| `CONNECTIONS_ORDERSDB_PROPERTIES_DATABASE`             | `ordersDB.properties.database`             |
+| `CONNECTIONS_ORDERSDB_PROPERTIES_SIZE`                 | `ordersDB.properties.size`                 |
+| `CONNECTIONS_ORDERSDB_PROPERTIES_HOST`                 | `ordersDB.properties.host`                 |
+| `CONNECTIONS_ORDERSDB_PROPERTIES_PORT`                 | `ordersDB.properties.port`                 |
+| `CONNECTIONS_ORDERSDB_PROPERTIES_CREDENTIALS_USERNAME` | `ordersDB.properties.credentials.username` |
+| `CONNECTIONS_ORDERSDB_PROPERTIES_CREDENTIALS_PASSWORD` | `ordersDB.properties.credentials.password` |
+
```
willdavsmith, line 790: so the list of available env vars are defined in the UDT type spec?

willdavsmith: yes, all properties are automatically exported as connection vars

architecture/2025-02-user-defined-resource-type-feature-spec.md
lakshmimsft, line 763: should this be "destination's resource's properties will be available within the source resource"

architecture/2025-02-user-defined-resource-type-feature-spec.md
nithyatsu: today for portable resources what we inject is CONNECTIONS_ORDERSDB_DATABASE, there is no "PROPERTIES" - Did we want it different for UDTs?

nithyatsu: Based on yesterday's discussion, confirming CONNECTIONS_ORDERSDB_DATABASE for env names and context.resource.connections.ordersDB.database for recipe context. We are leaving out "properties" since it is redundant.

lakshmimsft (approved):

PR #82, opened 2025-02-14
UDT schema structure design

nithyatsu:
UDT schema basics


architecture/2024-07-user-defned-types-schema-design.md
```diff
@@ -0,0 +1,305 @@
...
+}
+```
+
+## Design
+
+When user executes a `rad resource-type create` command, the cli should validate the schema object provided in the manifest and report any errors back to the users before making the API call. 
+
+The server side, upon receiving a resource payload for a resource-type resource, must again validate the schema before it saves the resource into database.
+
+We should implement validations on both client and server since rad cli need not be the only client. It is important to validate payloads since these can be any arbitrary data.
```
kachawla, line 159: What if we implement validations only on the server side? Is there a benefit to doing it twice for client using rad cli?

architecture/2024-07-user-defned-types-schema-design.md
```diff
@@ -0,0 +1,305 @@
...
+
+As a cloud operations engineer, I am responsible for ensuring the deployment of databases. The permissible sizes of the databases vary depending on whether they are intended for development or production environments. It is crucial to prevent development engineers from provisioning databases with excessive resources.
+
+## User Experience (if applicable)
+
+Users author a manifest like the following to defind a user-defined resource type.
+
+**Sample Input:**
+
+```yaml
```
sk593, line 76: Is the YAML definition getting replaced with Bicep as discussed in the feature spec? Or is this YAML for a different use case?

architecture/2024-07-user-defned-types-schema-design.md
```diff
@@ -0,0 +1,305 @@
...
+        - XL
+      host:
+        type: string
+        description: hostname
+        maxLength: 20
+```
+
+* array
+
+`array` must specify a `type` for item
```
sylvainsf, line 203: Is another way of saying this that arrays must be composed of items all with a valid type?

architecture/2024-07-user-defned-types-schema-design.md
```diff
@@ -0,0 +1,305 @@
...
+    type: object
+    properties:
+      name:
+        type: string
+        description: The name of the resource
+      labels:
+        type: object
+        description: A map of labels for the resource
+        additionalProperties:
+          type: string
```
ytimocin, line 234: Is there an example to this map?

architecture/2024-07-user-defned-types-schema-design.md
```diff
@@ -0,0 +1,305 @@
...
+Schemas support maps using `additional properties`. The additionalProperties keyword allows for dynamic keys that are not predefined in the schema
+The Schema is allowed to have either user-defined properties or additional properties which specifies a type. The type can be `Object` or another scalar.
+
+# Arrays
+
+Arrays must specify a type for `item`.
+
+# References
+
+Schemas are allowed to reference reusable built-in schemas of Radius using the `$ref` construct of OpenAPI.
```
sylvainsf, line 253: Is it a non-goal to support other sources of schemas? I could see this being a lot like private bicep or terraform registries.

architecture/2024-07-user-defned-types-schema-design.md
```diff
@@ -0,0 +1,305 @@
...
+        additionalProperties:
+          type: string
+```
+
+
+# Scalars
+
+Schemas will support every construct OpenAPI provides for a scalar field. That includes all of the modifiers like `enum` and `format` and also all of the validation attributes like `minLength`.
+
+# Maps
```
ytimocin, line 242: An example would be nice. Is it string:string type of map? What does the type of additionalProperties represent? Type of the key or the value of the map?

architecture/2024-07-user-defned-types-schema-design.md
```diff
@@ -0,0 +1,305 @@
...
+* **Author**: @nithyatsu
+
+## Overview
+
+The [user-defined-types](https://github.com/radius-project/design-notes/blob/main/architecture/2024-07-user-defined-types.md) feature enables end-users to define their own resource types as part of their tenant of Radius. User-defined types have the same set of capabilities and enable the same features (connections, app graph, recipes, etc) as system-defined types in Radius. 
+
+User-defined types are created with rad resource-type create command, which takes a resource type manifest file as input. Here is a very simple resource manifest -
+
+ ```yaml
+name: 'Mycompany.Messaging'
```
Reshrahim, line 12: Earlier this was namespace, I think we changed this to name not to be confusing with Kubernetes namespace. Now I think it causes confusion if its resource type name or something else. 

I think `namespace` might be better and  closer to the what it actually means

architecture/2024-07-user-defned-types-schema-design.md
```diff
@@ -0,0 +1,305 @@
...
+*Please read the [Radius API](https://docs.radapp.io/concepts/technical/api/) conceptual documentation. This document will heavily use the terminology and concepts defined there.*
+
+|Term| Definition|                                                              
+| ----------------------------- | ------------------------------------------|
+| User-defined type | A resource-type that can be defined or modified by end-users.|
+| OpenAPI/Schema | A format for documenting HTTP-based APIs. In this document we're primarily concerned with the parts of OpenAPI pertaining to request/response bodies.   |                                                                     
+
+## Objectives
+
+https://dev.azure.com/azure-octo/Incubations/_workitems/edit/13731
```
zachcasper, line 52: Since this is not publicly accessible, we shouldn't include it here.

architecture/2024-07-user-defned-types-schema-design.md
```diff
@@ -0,0 +1,305 @@
...
+
+All the limitations here are because at this point, we want to limit complexity and keep our tooling simple. Based on feedback, we would likely add support for some of the more complex OpenAPI features. 
+
+1. We are not supporting inheritence and polymorphism. Objects may not use the following constructs:
+
+- `allOf`
+- `anyOf`
+- `oneOf`
+- `not`
+- `discriminator`
```
ytimocin, line 270: Are these going to be supported in the upcoming versions of UDT?

architecture/2024-07-user-defned-types-schema-design.md
sylvainsf: I see this addressed below

architecture/2024-07-user-defned-types-schema-design.md
```diff
@@ -0,0 +1,305 @@
...
+                  $ref: 'https://radapp.io/schemas/v1#RecipeStatus'
+                  
+          required:
+          - size
+
+    capabilities: ["SupportsRecipes"]
+```
+**Sample Output:**
+
+Users can use this schema to generate a Bicep extension providing strongly-typed editor support. Here's an example of Bicep code that matches this API definition.
```
nicolejms, line 121: are we doing this generation for them?

architecture/2024-07-user-defned-types-schema-design.md
```diff
@@ -0,0 +1,305 @@
...
+
+- `allOf`
+- `anyOf`
+- `oneOf`
+- `not`
+- `discriminator`
+
+2. Objects may not set both `additionalProperties` as well as  define their own properties.
+
+3. Schemas are not allowed to use `$ref` to reference other than what we provide in Radius. This reduces concept count and simplifies our tooling. We can reconsider this based on feedback. This also prevents the definition of recursive types and circular references.
```
ytimocin, line 274: Are we thinking about supporting 3rd party schemas in future? For example, like UDT, my company may have custom schemas and I would like to point to them in my UDTs.

architecture/2024-07-user-defned-types-schema-design.md
```diff
@@ -0,0 +1,305 @@
...
+# References
+
+Schemas are allowed to reference reusable built-in schemas of Radius using the `$ref` construct of OpenAPI. 
+
+The example above uses `$ref: 'https://radapp.io/schemas/v1#RecipeStatus'` to reference the schema for a "recipe status". This aids with consistency and reduces boilerplate for users. 
+
+The exact URL and set of types that can be referenced is TBD.
+
+
+# Limitations
```
kachawla, line 260: Are we enforcing these through validation?

architecture/2024-07-user-defned-types-schema-design.md
```diff
@@ -0,0 +1,305 @@
...
+            openAPIV3Schema:
+              type: object
+              properties:
+                host:
+                  type: string
+                  description: hostname 
+                port:
+                  type: string
+                  description: port
+              required:
```
Reshrahim, line 27: For required, Ryan pointed me to this docs earlier saying the required must be array -https://json-schema.org/draft/2020-12/json-schema-validation#name-required

architecture/2024-07-user-defned-types-schema-design.md
```diff
@@ -0,0 +1,305 @@
...
+
+#### Structure
+
+Users provide us with the schema of UDT by defining `properties` in `openAPIV3Schema`. `openAPIV3Schema` is an `object` that holds properties of UDT. There is no support for defining custom fields outside of `properties`.
+
+Each of the property is a user defined property of the UDT. It has a type and optional description and format. The type can be 
+
+* scalar
+
+We support all of the scalars supported by OpenAPI.
```
brooke-hamilton, line 179: Is this the right link? I don't see scalar types defined there, but maybe I'm reading it wrong.

brooke-hamilton (commented):
Great document!

architecture/2024-07-user-defned-types-schema-design.md
```diff
@@ -0,0 +1,305 @@
...
+
+# Arrays
+
+Arrays must specify a type for `item`.
+
+# References
+
+Schemas are allowed to reference reusable built-in schemas of Radius using the `$ref` construct of OpenAPI. 
+
+The example above uses `$ref: 'https://radapp.io/schemas/v1#RecipeStatus'` to reference the schema for a "recipe status". This aids with consistency and reduces boilerplate for users.
```
brooke-hamilton, line 255: What is the example this refers to?

architecture/2024-07-user-defned-types-schema-design.md
```diff
@@ -0,0 +1,305 @@
...
+
+* **Author**: @nithyatsu
+
+## Overview
+
+The [user-defined-types](https://github.com/radius-project/design-notes/blob/main/architecture/2024-07-user-defined-types.md) feature enables end-users to define their own resource types as part of their tenant of Radius. User-defined types have the same set of capabilities and enable the same features (connections, app graph, recipes, etc) as system-defined types in Radius. 
+
+User-defined types are created with rad resource-type create command, which takes a resource type manifest file as input. Here is a very simple resource manifest -
+
+ ```yaml
```
zachcasper, line 11: Can we also show the example in Bicep and note that the YAML format is only for very simple use cases.

architecture/2024-07-user-defned-types-schema-design.md
brooke-hamilton, line 253: Are schemas able to reference other UDT schemas?

architecture/2024-07-user-defned-types-schema-design.md
kachawla: The example shows a YAML sequence which evaluates to an array.

architecture/2024-07-user-defned-types-schema-design.md
```diff
@@ -0,0 +1,305 @@
...
+                    username:
+                      type: string
+                    secret:
+                      type: string
+                      schema:
+                        type: 'object'
+                        properties:
+                          password:
+                            type: 'string'
+                recipe:
```
zachcasper, line 111: The recipe is environment-specific while the resource type is not. Why is recipe here?

architecture/2024-07-user-defned-types-schema-design.md
```diff
@@ -0,0 +1,305 @@
...
+  name: 'sample-db'
+  properties: {
+    size: 'L' // Bicep editor can validate this field
+  }
+}
+```
+
+## Design
+
+When user executes a `rad resource-type create` command, the cli should validate the schema object provided in the manifest and report any errors back to the users before making the API call.
```
Reshrahim, line 155: we should also validate if the user has also the required properties in the manifest.

architecture/2024-07-user-defned-types-schema-design.md
```diff
@@ -0,0 +1,305 @@
...
+Schemas are allowed to reference reusable built-in schemas of Radius using the `$ref` construct of OpenAPI. 
+
+The example above uses `$ref: 'https://radapp.io/schemas/v1#RecipeStatus'` to reference the schema for a "recipe status". This aids with consistency and reduces boilerplate for users. 
+
+The exact URL and set of types that can be referenced is TBD.
+
+
+# Limitations
+
+All the limitations here are because at this point, we want to limit complexity and keep our tooling simple. Based on feedback, we would likely add support for some of the more complex OpenAPI features.
```
lakshmimsft, line 262: Do we know of any other limitations that biceptype generation may have in compatibility with vOpenAPIv3

architecture/2024-07-user-defned-types-schema-design.md
kachawla: not sure if I follow, required properties aren't required in the manifest

architecture/2024-07-user-defned-types-schema-design.md
kachawla: I would lean towards doing that when we starting looking into the technical design of bicep. None of the investigation work has been done yet.

nithyatsu: Bicep will be supported, but not for phase 1, I will cover the details there as a seperate doc.

architecture/2024-07-user-defned-types-schema-design.md
nithyatsu: I will confirm this based on openapi 3.x

architecture/2024-07-user-defned-types-schema-design.md
Reshrahim: I meant validate the presence of top level fields like the namespace, types, APIversion etc. Those I believe are must haves for the manifest.

architecture/2024-07-user-defned-types-schema-design.md
```diff
@@ -0,0 +1,305 @@
...
+                        properties:
+                          password:
+                            type: 'string'
+                recipe: 
+                  $ref: 'https://radapp.io/schemas/v1#RecipeStatus'
+                  
+          required:
+          - size
+
+    capabilities: ["SupportsRecipes"]
```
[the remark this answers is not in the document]
nithyatsu: allowsRecipes: false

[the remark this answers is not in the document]
nithyatsu: not for phase 1.

architecture/2024-07-user-defned-types-schema-design.md
nithyatsu: I will investigate bicep compatibility aspects.

architecture/2024-07-user-defned-types-schema-design.md
nithyatsu: We could avoid unneccessary server calls by catching issues which can be caught in cli. All cli commands have a validate function which captures issues that can be addressed before making api calls, we can plug in schema validation into this.

architecture/2024-07-user-defned-types-schema-design.md
nithyatsu: We will explore bicep, but not for first iteration.

architecture/2024-07-user-defned-types-schema-design.md
nithyatsu: yes. it enforces a static type on each element.

architecture/2024-07-user-defned-types-schema-design.md
nithyatsu: created https://github.com/radius-project/design-notes/pull/82#discussion_r1962198019 to address the change.

architecture/2024-07-user-defned-types-schema-design.md
nithyatsu: yes, we have initial version of the tooling to do this.

architecture/2024-07-user-defned-types-schema-design.md
nithyatsu: Deepending on field requirements we will support 3rd party schemas as well as any other openapi features for future.

architecture/2024-07-user-defned-types-schema-design.md
nithyatsu: Yes, we have to enforce them through validation.

architecture/2024-07-user-defned-types-schema-design.md
nithyatsu: clarified the format is good for yaml and the above example is json.

architecture/2024-07-user-defned-types-schema-design.md
nithyatsu: added

architecture/2024-07-user-defned-types-schema-design.md
nithyatsu: not in the first draft but based on field rqeuirements we might implement them as part of next iteration.

architecture/2024-07-user-defned-types-schema-design.md
nithyatsu: They arent part of user defined schema but are validated today.

architecture/2024-07-user-defned-types-schema-design.md
nithyatsu: Karishma, Lakshmi and I discussed this and went over the implementation we have today. We dont need to have the user input these fields specifically. These are needed since we have a path for supporting"status" without requiring the user to explicitly define it.

architecture/2024-07-user-defned-types-schema-design.md
nithyatsu: rewrote the example. But the idea was to convey we can use #ref to reuse schemas that Radius provides, but not for user defined schemas.

architecture/2024-07-user-defned-types-schema-design.md
```diff
@@ -0,0 +1,339 @@
...
+        type: array
+        description: "ports this resource binds to"
+        item:
+          type: integer
+          format: uint32
+```
+
+* map
+
+We support map through `adddiionalProperties`. This is useful when the resource type allows for dynamic(user defined) keys. We still must specify a type for the value of the property.
```
lakshmimsft, line 234: nit: additional

ytimocin (approved):

kachawla (approved):

PR #84, opened 2025-02-27
Feature Spec: Radius Control Plane Upgrades

willtsai:
Add feature spec for Radius Control Plane Upgrades

Issue reference: https://github.com/radius-project/roadmap/issues/52

### DIFF
--- a/tools/2025-02-control-plane-upgrades-feature-spec.md
+++ b/tools/2025-02-control-plane-upgrades-feature-spec.md
@@ -1,0 +1,73 @@
+# Topic: Radius Control Plane Upgrades
+
+* **Author**: Will Tsai (@WillTsai)
+
+## Topic Summary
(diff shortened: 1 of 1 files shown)


tools/2025-02-control-plane-upgrades-feature-spec.md
```diff
@@ -0,0 +1,73 @@
...
+
+## Details of user problem
+When I try to upgrade the Radius control plane, I face challenges related to downtime and complexity. The current process is manual and error-prone, leading to potential disruptions in service. These issues result in increased operational overhead and reduced reliability of my Radius deployment.
+
+## Desired user experience outcome
+After this scenario is implemented, I can perform control plane upgrades seamlessly and without downtime. As a result, I can maintain high availability and reliability of my Radius deployment, reducing operational overhead and ensuring a smooth upgrade process.
+
+### Detailed user experience
+1. User initiates the upgrade process using Helm.
+2. The system performs compatibility checks to ensure the new version is compatible with existing configurations.
```
ytimocin, line 61: I think the two most important things to check are:
- If there is a DB migration and, if so, apply that (not sure how this would work backwards as in downgrades)
- If there is a change in one of the APIs that could break the resources of the user in production

willtsai:
resolving in favor of https://github.com/radius-project/design-notes/pull/85

PR #85, opened 2025-03-05
Feature specification for in-place upgrade of Radius

ytimocin:


architecture/2025-03-upgrade-feature-spec.md
```diff
@@ -0,0 +1,221 @@
...
+
+**Dependencies:**
+
+- Dapr
+- Contour
+- Postgres
+
+**Key Features:**
+
+- **Compatibility:** Ensures compatibility with new versions.
```
willdavsmith, line 26: what if we can't ensure compatibility with new versions? I'm not sure we can commit to "no breaking changes" while we aren't at 1.0.0

architecture/2025-03-upgrade-feature-spec.md
```diff
@@ -0,0 +1,221 @@
...
+- **Data Safety:** Takes snapshots before the upgrade to safeguard user data, with automatic rollback in case of failure.
+- **Seamless Experience:** Provides a smooth and reliable upgrade process with minimal downtime.
+
+This feature will significantly improve the user experience by automating the upgrade process, reducing the risk of errors, and maintaining system stability.
+
+### Top level goals
+
+- Deliver a seamless and intuitive upgrade experience for users.
+- Ensure a smooth and reliable upgrade process for the components and dependencies of Radius.
+- Minimize downtime and disruption during upgrades, ensuring that the Radius CLI is available as soon as possible.
```
willdavsmith, line 36: should this be control plane or CLI?

architecture/2025-03-upgrade-feature-spec.md
```diff
@@ -0,0 +1,221 @@
...
+  - Uninstall the existing Radius installation by running `rad uninstall kubernetes`. This command doesn't delete any user data or the `radius-system` namespace.
+  - Download and install the desired version locally. Ensure it is the correct version by running `rad version`.
+  - Run `rad install kubernetes` or `rad init` to finalize the upgrade.
+- This process is also not as seamless as a `rad upgrade kubernetes --version desired.version` command could be.
+
+### Positive user outcome
+
+- By running the `rad upgrade kubernetes --version desired.version` command, users will be able to initiate the upgrade process seamlessly.
+  - The desired version must be higher than the current version.
+- By setting a global flag indicating that an upgrade is in progress, we ensure that users do not start data-changing commands like `rad deploy app.bicep` or `rad delete app` in a different terminal.
```
willdavsmith, line 72: good idea

architecture/2025-03-upgrade-feature-spec.md
sylvainsf: That has to be answered by the versioning story. This requirement should be removed from this spec.

architecture/2025-03-upgrade-feature-spec.md
```diff
@@ -0,0 +1,221 @@
...
+rad upgrade kubernetes --version v0.44.0
+
+# Upgrade with custom configuration values
+rad upgrade kubernetes --version v0.44.0 --set global.monitoring.enabled=true
+
+# Perform a dry-run to simulate the upgrade without making changes
+rad upgrade kubernetes --version v0.44.0 --dry-run
+
+# Upgrade with extended timeout for large deployments
+rad upgrade kubernetes --version v0.44.0 --timeout 600
```
willdavsmith, line 93: this might be unnecessary, we will only be upgrading control plane components I think

architecture/2025-03-upgrade-feature-spec.md
ytimocin: We can discuss what we mean by breaking changes. I think, if we are not removing a property from a resource type which means a migration will update the data store we are using, `rad upgrade kubernetes` should work fine.

Can you let me know what other scenarios you were thinking of?

architecture/2025-03-upgrade-feature-spec.md
```diff
@@ -0,0 +1,221 @@
...
+**Dependencies:**
+
+- Dapr
+- Contour
+- Postgres
+
+**Key Features:**
+
+- **Compatibility:** Ensures compatibility with new versions.
+- **Data Safety:** Takes snapshots before the upgrade to safeguard user data, with automatic rollback in case of failure.
```
sylvainsf, line 27: We should probably call out that this is backend dependent and plan to ship with support for etcd and postgresql

architecture/2025-03-upgrade-feature-spec.md
```diff
@@ -0,0 +1,221 @@
...
+- Deliver a seamless and intuitive upgrade experience for users.
+- Ensure a smooth and reliable upgrade process for the components and dependencies of Radius.
+- Minimize downtime and disruption during upgrades, ensuring that the Radius CLI is available as soon as possible.
+- Ensure user data safety during the upgrade process by taking snapshots and providing rollback capabilities.
+- Provide clear documentation and guidance for users performing upgrades.
+- Maintain system performance during the upgrade process to avoid significant impact on running applications.
+
+### Non-goals (out of scope)
+
+- Downgrades are not supported. The upgrade process is designed to move forward to newer versions only.
```
kachawla, line 43: The rollbacks mentioned above is effectevely going to be a downgrade back to the old version though?

architecture/2025-03-upgrade-feature-spec.md
```diff
@@ -0,0 +1,221 @@
...
+rad upgrade kubernetes --version v0.44.0 --timeout 600
+```
+
+The upgrade process will:
+
+1. **Pre-flight checks**: Detect existing installation, check versions, ensure the user is not attempting a downgrade, etc.
+2. **Fetch available chart versions**: Provide a list of known chart versions so the desired version that the users select is a valid one.
+3. **Dry-run** (when requested): Simulate the upgrade, logging steps without making changes. Also making sure that the upgrade will work. Helm has this feature available in the `helm upgrade` command: <https://helm.sh/docs/helm/helm_upgrade/>.
+4. **Snapshot**: Automatically back up current data (e.g., etcd, resources in the API server, or Postgres) before making changes.
+5. **Upgrade**: Apply necessary Helm changes (including timeouts, set args, etc.), optionally perform database migrations if needed.
```
willdavsmith, line 102: how do we know when to perform database migrations? if we introduce a breaking change to one of our schemas, is there an automatic way to detect and upgrade? probably not

architecture/2025-03-upgrade-feature-spec.md
ytimocin: I was thinking about a scenario where the upgrade gets stuck and doesn't move forward. That could make Radius unavailable for a while with no way to go back. But, if there is a timeout, the process can be rolled back if it is not successful in the given time.

architecture/2025-03-upgrade-feature-spec.md
```diff
@@ -0,0 +1,221 @@
...
+### Success Metrics & Validation
+
+- **Upgrade Success Rate:** Define metrics to track successful vs. failed upgrades
+- **Performance Impact:** Measure and minimize performance degradation during upgrades
+- **Downtime Duration:** Establish acceptable limits for component unavailability during the process
+- **Validation Testing:** Define comprehensive test cases to verify proper operation post-upgrade
+
+### Research Plan
+
+- Analyze upgrade approaches from similar platforms (Dapr, Crossplane, Istio) to adopt best practices
```
willdavsmith, line 156: Flux would be a good one too

architecture/2025-03-upgrade-feature-spec.md
```diff
@@ -0,0 +1,221 @@
...
+- This approach is cumbersome and does not provide a good user experience. Additionally, it does not automatically install dependencies like Contour and Dapr, requiring users to handle these installations manually.
+- Another way to upgrade Radius to the desired version involves the following steps:
+  - Uninstall the existing Radius installation by running `rad uninstall kubernetes`. This command doesn't delete any user data or the `radius-system` namespace.
+  - Download and install the desired version locally. Ensure it is the correct version by running `rad version`.
+  - Run `rad install kubernetes` or `rad init` to finalize the upgrade.
+- This process is also not as seamless as a `rad upgrade kubernetes --version desired.version` command could be.
+
+### Positive user outcome
+
+- By running the `rad upgrade kubernetes --version desired.version` command, users will be able to initiate the upgrade process seamlessly.
```
sylvainsf, line 70: I'd like to add --version latest as an option

architecture/2025-03-upgrade-feature-spec.md
ytimocin: We have two states: current version and desired version.

Rollbacks will only occur if there is a failure during the `rad upgrade kubernetes` process and they are going to roll back to the current version. So, it is not going to go back to an older version.

It just means: "You wanted to upgrade to version 0.44 from 0.25 but there is a failure during that process, we are taking you back to 0.25"

architecture/2025-03-upgrade-feature-spec.md
kachawla, line 72: Is the plan to block any mutating operations on the server side? Where is this flag stored?

architecture/2025-03-upgrade-feature-spec.md
```diff
@@ -0,0 +1,221 @@
...
+
+```bash
+# Basic upgrade to a specific version
+rad upgrade kubernetes --version v0.44.0
+
+# Upgrade with custom configuration values
+rad upgrade kubernetes --version v0.44.0 --set global.monitoring.enabled=true
+
+# Perform a dry-run to simulate the upgrade without making changes
+rad upgrade kubernetes --version v0.44.0 --dry-run
```
sylvainsf, line 90: What's the goal? What would the output be?

architecture/2025-03-upgrade-feature-spec.md
```diff
@@ -0,0 +1,221 @@
...
+
+## Key scenarios
+
+### Scenario 1: Upgrading the control plane using the Radius CLI
+
+After the implementation of this feature, the Radius CLI will provide a streamlined approach to upgrade the control plane components with minimal disruption. Users can execute various upgrade scenarios through the `rad upgrade kubernetes` command with appropriate flags.
+
+```bash
+# Basic upgrade to a specific version
+rad upgrade kubernetes --version v0.44.0
```
kachawla, line 84: What's benefit of including `kubernetes` as an argument here?

architecture/2025-03-upgrade-feature-spec.md
```diff
@@ -0,0 +1,221 @@
...
+
+- As of March 2025, Radius does not support in-place upgrades or upgrades via the Radius CLI. Users can achieve this using Helm, but this method does not update the local Radius CLI version to the desired version. The current steps are:
+  - `helm install radius oci://ghcr.io/radius-project/helm-chart/radius --version 0.43.0`
+  - `helm upgrade radius oci://ghcr.io/radius-project/helm-chart/radius --version 0.44.0`
+- This approach is cumbersome and does not provide a good user experience. Additionally, it does not automatically install dependencies like Contour and Dapr, requiring users to handle these installations manually.
+- Another way to upgrade Radius to the desired version involves the following steps:
+  - Uninstall the existing Radius installation by running `rad uninstall kubernetes`. This command doesn't delete any user data or the `radius-system` namespace.
+  - Download and install the desired version locally. Ensure it is the correct version by running `rad version`.
+  - Run `rad install kubernetes` or `rad init` to finalize the upgrade.
+- This process is also not as seamless as a `rad upgrade kubernetes --version desired.version` command could be.
```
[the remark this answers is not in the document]
ytimocin: I was thinking that we can. Not put a lot of thought into it though but I believe that can be achievable.

architecture/2025-03-upgrade-feature-spec.md
sylvainsf, line 102: Shouldn't this apply all upgrades in order? A non-goal should be supporting arbitrary version jumps which skip versions. Otherwise we need to resolve the delta of db migrations

architecture/2025-03-upgrade-feature-spec.md
ytimocin: Latest as in edge or latest as in 0.44 (latest we have now) :)

architecture/2025-03-upgrade-feature-spec.md
```diff
@@ -0,0 +1,221 @@
...
+### User persona(s)
+
+- **System Administrator**: Responsible for managing the infrastructure and ensuring the smooth operation of the Radius platform.
+- **DevOps Engineer**: Focused on automating deployment processes and maintaining the CI/CD pipeline.
+
+### Challenge(s) faced by the user
+
+- As of March 2025, Radius does not support in-place upgrades or upgrades via the Radius CLI. Users can achieve this using Helm, but this method does not update the local Radius CLI version to the desired version. The current steps are:
+  - `helm install radius oci://ghcr.io/radius-project/helm-chart/radius --version 0.43.0`
+  - `helm upgrade radius oci://ghcr.io/radius-project/helm-chart/radius --version 0.44.0`
```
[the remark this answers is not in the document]
ytimocin: You get a warning saying that you have Radius installed and you should use `--reinstall`.

architecture/2025-03-upgrade-feature-spec.md
```diff
@@ -0,0 +1,221 @@
...
+
+1. **Pre-flight checks**: Detect existing installation, check versions, ensure the user is not attempting a downgrade, etc.
+2. **Fetch available chart versions**: Provide a list of known chart versions so the desired version that the users select is a valid one.
+3. **Dry-run** (when requested): Simulate the upgrade, logging steps without making changes. Also making sure that the upgrade will work. Helm has this feature available in the `helm upgrade` command: <https://helm.sh/docs/helm/helm_upgrade/>.
+4. **Snapshot**: Automatically back up current data (e.g., etcd, resources in the API server, or Postgres) before making changes.
+5. **Upgrade**: Apply necessary Helm changes (including timeouts, set args, etc.), optionally perform database migrations if needed.
+6. **Rollback** (on failure): If something goes wrong, use the snapshot to restore the prior state.
+7. **Post-upgrade checks**: Validate that new control plane components are healthy and confirm the upgrade was successful.
+
+### Scenario 2: Upgrading the control plane using Helm
```
sylvainsf, line 106: This should just be a non-goal, not a scenario to design for

architecture/2025-03-upgrade-feature-spec.md
```diff
@@ -0,0 +1,221 @@
...
+```
+
+The upgrade process will:
+
+1. **Pre-flight checks**: Detect existing installation, check versions, ensure the user is not attempting a downgrade, etc.
+2. **Fetch available chart versions**: Provide a list of known chart versions so the desired version that the users select is a valid one.
+3. **Dry-run** (when requested): Simulate the upgrade, logging steps without making changes. Also making sure that the upgrade will work. Helm has this feature available in the `helm upgrade` command: <https://helm.sh/docs/helm/helm_upgrade/>.
+4. **Snapshot**: Automatically back up current data (e.g., etcd, resources in the API server, or Postgres) before making changes.
+5. **Upgrade**: Apply necessary Helm changes (including timeouts, set args, etc.), optionally perform database migrations if needed.
+6. **Rollback** (on failure): If something goes wrong, use the snapshot to restore the prior state.
```
kachawla, line 103: If mutations are not allowed when upgrade is in progress then is restoring from snapshot needed?

architecture/2025-03-upgrade-feature-spec.md
```diff
@@ -0,0 +1,221 @@
...
+- **Dependency Management:** We use a specific version of Contour and Dapr (<https://github.com/radius-project/radius/blob/main/pkg/cli/helm/cluster.go#L34>) and they are not overridable with the configuration. This ensures consistent dependency management. There is a case where the default chart versions may be updated in a newer version of Radius but that means that those versions of Contour and Dapr are probably tested with that new version of Radius and shouldn't provide problems during the upgrade from the current version and the desired version which is the newer version that we just discussed about.
+- **Resource Requirements:** The upgrade process won't exceed available cluster resources during the transition period when both old and new components may be running simultaneously.
+
+### Technical Questions to Resolve
+
+- **Downgrade Support:**
+  - Should we support downgrading to previous versions? If yes, what are the limitations?
+  - How should we handle cases where users attempt to downgrade to versions that don't support the upgrade feature itself?
+- **Version Skipping:**
```
sylvainsf, line 136: Having tackled this problem before we should not support this at this time.

architecture/2025-03-upgrade-feature-spec.md
ytimocin: Yes, the plan is to let other users (terminals, etc.) know that there is an upgrade going on so that they shouldn't kick off any operations that would change the data.

Not sure where the flag will be stored as of now but shouldn't be a big problem. (I hope)

architecture/2025-03-upgrade-feature-spec.md
```diff
@@ -0,0 +1,221 @@
...
+  - Can users skip multiple versions in a single upgrade (e.g., v0.40 → v0.44), or should we enforce incremental upgrades?
+- **Failure Recovery:**
+  - What recovery mechanisms should be implemented if an upgrade fails mid-process?
```
sylvainsf, line 139: We need to account for the failure mode of rad-cli losing connection during an upgrade.

architecture/2025-03-upgrade-feature-spec.md
ytimocin: That you can actually upgrade to the version you want without problems. I would want that simulation capability as a user.

architecture/2025-03-upgrade-feature-spec.md
ytimocin: Like `rad install kubernetes` we are specifying that we are upgrading the Radius on the Kubernetes cluster.

architecture/2025-03-upgrade-feature-spec.md
```diff
@@ -0,0 +1,221 @@
...
+
+## Key investments
+
+### Feature 1
+
+Implement pre-upgrade and post-upgrade checks to ensure compatibility and verify the success of the upgrade. These checks are:
+
+1. **Pre-Upgrade Checks:**
+   1. **Version Compatibility:** Ensure the current version is behind the desired version to prevent downgrades.
+   2. **Snapshot Creation:** Automatically create a snapshot of the current data (e.g., etcd, resources in the API server, or Postgres) to ensure user data safety in case of rollback.
```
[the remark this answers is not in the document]
sylvainsf: We should notify gracefully that this is taking place rather than timeouts on people attempting these operations. If you implement a DB level lock that would be the result.

architecture/2025-03-upgrade-feature-spec.md
ytimocin: If the new Radius installation changes etcd (or any other data store) during the upgrade, yes, we need a way to go back to the older state.

kachawla (commented):
How do we handle in progress deployments when an upgrade is initiated?

architecture/2025-03-upgrade-feature-spec.md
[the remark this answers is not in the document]
ytimocin: Upgrade may be writing some data to the data store during the process.

architecture/2025-03-upgrade-feature-spec.md
```diff
@@ -0,0 +1,226 @@
...
+- Applications Resource Provider
+- Dynamic Resource Provider
+- Dashboard
+- Controller
+
+**Dependencies:**
+
+- Dapr
+- Contour
+- Postgres
```
willtsai, line 22: Dapr and Postgres removed as of `v0.46`
```suggestion
- Contour
```

architecture/2025-03-upgrade-feature-spec.md
```diff
@@ -0,0 +1,226 @@
...
+- **Dependency Management:** We use a specific version of Contour and Dapr (<https://github.com/radius-project/radius/blob/main/pkg/cli/helm/cluster.go#L34>) and they can not be updated with the configuration. This ensures consistent dependency management. There is a case where the default chart versions may be updated in a newer version of Radius but that means that those versions of Contour and Dapr are probably tested with that new version of Radius and shouldn't provide problems during the upgrade from the current version and the desired version which is the newer version that we just discussed about.
+- **Resource Requirements:** The upgrade process won't exceed available cluster resources during the transition period when both old and new components may be running simultaneously.
+
+### Technical Questions to Resolve
+
+- **Downgrade Support:**
+  - Should we support downgrading to previous versions? If yes, what are the limitations?
+  - How should we handle cases where users attempt to downgrade to versions that don't support the upgrade feature itself?
```
willtsai, line 140: I think it's reasonable to only allow downgrades up to the first version that supported upgrades

architecture/2025-03-upgrade-feature-spec.md
```diff
@@ -0,0 +1,226 @@
...
+- **Data Integrity:** The upgrade process will maintain data integrity and prevent corruption during migration operations (particularly for database migrations) by using snapshots and the ability to restore in case of failure.
+- **Dependency Management:** We use a specific version of Contour and Dapr (<https://github.com/radius-project/radius/blob/main/pkg/cli/helm/cluster.go#L34>) and they can not be updated with the configuration. This ensures consistent dependency management. There is a case where the default chart versions may be updated in a newer version of Radius but that means that those versions of Contour and Dapr are probably tested with that new version of Radius and shouldn't provide problems during the upgrade from the current version and the desired version which is the newer version that we just discussed about.
+- **Resource Requirements:** The upgrade process won't exceed available cluster resources during the transition period when both old and new components may be running simultaneously.
+
+### Technical Questions to Resolve
+
+- **Downgrade Support:**
+  - Should we support downgrading to previous versions? If yes, what are the limitations?
```
willtsai, line 139: I think we should support them to the same extent of the mechanisms used to support upgrades (e.g. preflight checks, rollback if something goes wrong, etc.)

architecture/2025-03-upgrade-feature-spec.md
```diff
@@ -0,0 +1,226 @@
...
+
+The upgrade process will:
+
+1. **Pre-flight checks**: Detect existing installation, check versions, ensure the user is not attempting a downgrade, etc.
+2. **Fetch available chart versions**: Provide a list of known chart versions so the desired version that the users select is a valid one.
+3. **Dry-run** (when requested): Simulate the upgrade, logging steps without making changes. Also making sure that the upgrade will work. Helm has this feature available in the `helm upgrade` command: <https://helm.sh/docs/helm/helm_upgrade/>.
+4. **Snapshot**: Automatically back up current data (e.g., etcd, resources in the API server, or Postgres) before making changes.
+5. **Upgrade**: Apply necessary Helm changes (including timeouts, set args, etc.), optionally perform database migrations if needed.
+6. **Rollback** (on failure): If something goes wrong, use the snapshot to restore the prior state.
+7. **Post-upgrade checks**: Validate that new control plane components are healthy and confirm the upgrade was successful.
```
willtsai, line 109: We've gotten some feedback from community users that they might have to roll back the upgrade even after the post-upgrade checks have passed (perhaps their own tests fail in the upgraded version) -- is it possible to do a rollback on the snapshot even after the upgrade successfully completes?

architecture/2025-03-upgrade-feature-spec.md
ytimocin: These questions and their answers are being discussed in the design document.

architecture/2025-03-upgrade-feature-spec.md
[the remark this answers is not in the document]
ytimocin: It is going to be supported in future versions.

architecture/2025-03-upgrade-feature-spec.md
ytimocin: Will be supported in the upcoming versions.

architecture/2025-03-upgrade-feature-spec.md
ytimocin: I removed this section.

willtsai (approved):

architecture/2025-03-upgrade-feature-spec.md
```diff
@@ -0,0 +1,218 @@
...
+- Upgrading major versions of dependencies like Contour is not covered by this process. These upgrades should be handled separately following their respective guidelines.
+- Upgrading the Radius control plane using Helm directly. We can run `helm upgrade` on the Radius Helm installation but that is not going to put all the necessary pieces together for the control plane to work. Making this work is not in the scope of this work.
+- Zero-downtime control plane upgrades. While we aim to minimize disruption, guaranteeing absolutely no downtime for control plane components is not a goal for this initial release.
+- Automatic CLI upgrades. Users must manually update their local CLI version after upgrading the control plane.
+- Direct GitOps workflow integration for version 1. While users who manage Radius through HelmReleases in their GitOps pipeline will be able to update Helm charts, the complete upgrade process (including preflight checks, locking, and health verification) requires the `rad upgrade kubernetes` command in this initial version. Future versions will provide better GitOps integration options.
```
willtsai, line 52: per @willdavsmith this is in scope now

PR #86, opened 2025-03-05
Autoscaling Feature spec 

Reshrahim:


architecture/2025-02-application-autoscaling.md
```diff
@@ -0,0 +1,239 @@
...
+1. Autoscale config is run-time specific and doesn't have to be standardized across runtimes as different platforms have different autoscaling mechanisms. 
+1. Enterprises pick platforms based on their application workloads and the underlying platform provides default autoscaling policies based on the workloads. Radius provides a clear abstraction to leverage default autoscaling mechanisms from the underlying platform.
+
+Cons:
+1. More configuration to manage across different runtimes.
+2. The developer needs to know the underlying platform and the autoscaling policies required for the platform.
+
+#### Option 2: Adding it as part extensions
+
+`extensions` have been the way to provide punch through capabilities which do not change the behaviors of the resource. The following is an example of adding autoscaling as part of the extensions.
```
willtsai, line 112: minor correction: `runtimes` is the punch through

architecture/2025-02-application-autoscaling.md
```diff
@@ -0,0 +1,239 @@
...
+  }
+}
+```
+
+Pros: consistent with existing manual scaling config
+
+Cons: not straightforward or intuitive as user doesn't know which autoscale config is required for the platform ? needs a platform specific discriminator
+May be, this is not a developer problem and should be handled by the platform operator. 
+
+#### Option 3: Adding it as top level common property
```
willtsai, line 139: I think Option 3 makes the most sense IF hpa and/or http auto-scaling and their corresponding properties are common across container platforms.

architecture/2025-02-application-autoscaling.md
```diff
@@ -0,0 +1,239 @@
...
+        httpConcurrency: 50
+      }  
+    }
+  }
+}
+``` 
+
+Cons:same as above, not straightforward or intuitive as user doesn't know which autoscale config is required for the platform ? May be, this is not a developer problem and should be handled by the platform operator.
+
+### Configuring Autoscaling in Radius Environment
```
willtsai, line 163: Would autoscaling policies specified in the container override the autoscaling policies set for the environment?

architecture/2025-02-application-autoscaling.md
```diff
@@ -0,0 +1,239 @@
...
+}
+```
+
+Pros:
+1. Autoscaling is an infrastructure problem, the platform engineer has the ability to configure the autoscaling policies in the environment configuration thus separating the concerns of the platform engineer and the developer. 
+1. Dev, test environments wouldn't require autoscaling policies and the platform engineer has more flexibility to configure the scaling policies based on the environment.
+1. The platform engineer can configure the scaling policies once and all the applications deployed in the environment will inherit the scaling policies.
+1. Developer can focus on the application modelling and not worry about the scaling policies. If needed the developer can override the scaling policies in the application resource definition. Overriding scenario could be for example, if the developer wants to scale based on some events from the Kafka or RabbitMQ modelled in the application definition 
+
+### Configuring Autoscaling as a core resource type
```
willtsai, line 194: I don't follow this completely - I see Radius resources as objects it needs to provision or create at deploy time, but autoscaling is not an object, but rather a configuration. I think this kind of pattern might be better handled through bicep params instead?

PR #87, opened 2025-03-18
Radius in-place upgrade design document

ytimocin:
**Note: The base branch will be changed.**


architecture/2025-03-upgrade-design-doc.md
```diff
@@ -0,0 +1,677 @@
...
+
+    // Releases the upgrade lock, allowing other commands to proceed
+    ReleaseLock(ctx context.Context) error
+
+    // Checks if an upgrade is currently in progress
+    IsUpgradeInProgress(ctx context.Context) (bool, error)
+}
+```
+
+We can utilize Kubernetes Lease objects (coordination.k8s.io/v1) for implementing the distributed locking mechanism (open to discussion and suggestions). Leases are purpose-built for this use case, providing built-in lease duration and automatic expiration capabilities. For more information, see: <https://kubernetes.io/docs/concepts/architecture/leases/>.
```
sylvainsf, line 380: I'm hesitant to take a hard dependency on Kubernetes for this safety check. We don't have a need for a cross-cluster locks. Both etc and Postgres offer db level locks, though we would want to gracefully handle locks instead of timeout errors.

architecture/2025-03-upgrade-design-doc.md
```diff
@@ -0,0 +1,677 @@
...
+6. **Storage availability**: The Kubernetes cluster has sufficient persistent storage capacity for backup operations during the upgrade process.
+
+## Objectives
+
+> **Issue Reference:** <https://github.com/radius-project/radius/issues/8095>
+
+### Goals
+
+- **Simplify upgrade process**: Provide a single CLI command (`rad upgrade kubernetes`) to upgrade Radius without manual reinstallation.
+- **Ensure data safety**: Implement automatic user data backups before upgrades and restore capability if failures occur.
```
ytimocin, line 55: if we leave this out for the version 1, we will still have a functioning `rad upgrade kubernetes` and implementation will be simpler.

architecture/2025-03-upgrade-design-doc.md
```diff
@@ -0,0 +1,677 @@
...
+| **Pre-flight Checks**    | Validation steps performed before an upgrade to ensure prerequisites are met and the system is in a valid state for upgrading.                                                                                                                                                                                                                                               |
+| **Version Skipping**     | The ability to upgrade directly from one version to a newer non-adjacent version (e.g., v0.40 → v0.44) without installing intermediate versions.                                                                                                                                                                                                                             |
+
+## Assumptions
+
+1. **User permissions**: Users running the upgrade command have enough permissions on both the Kubernetes cluster and the Radius installation.
+
+2. **Resource requirements**: The Kubernetes cluster has sufficient compute resources (CPU, memory) to run both the existing and new version components during the rolling upgrade process.
```
kachawla, line 27: What's the expected user experience during upgrade if resources are constrained?

architecture/2025-03-upgrade-design-doc.md
```diff
@@ -0,0 +1,677 @@
...
+Checks will include:
+
+1. Version compatibility verification
+2. Existing installation detection
+3. Database connectivity
+4. Custom configuration validation
+
+**User Data Backup and Restore System:**
+
+Rather than taking complete snapshots of the underlying databases (etcd/PostgreSQL), we'll implement a more targeted approach that backs up only the user application metadata and configuration that Radius manages:
```
sylvainsf, line 413: Can you call out what would not be backed up in this approach?

architecture/2025-03-upgrade-design-doc.md
```diff
@@ -0,0 +1,677 @@
...
+
+## Objectives
+
+> **Issue Reference:** <https://github.com/radius-project/radius/issues/8095>
+
+### Goals
+
+- **Simplify upgrade process**: Provide a single CLI command (`rad upgrade kubernetes`) to upgrade Radius without manual reinstallation.
+- **Ensure data safety**: Implement automatic user data backups before upgrades and restore capability if failures occur.
+- **Minimize downtime**: Use rolling upgrades where possible to keep Radius control plane available during the upgrade process.
```
ytimocin, line 56: We can use rolling upgrade feature of Kubernetes to be able to revert the upgrade process if there is an issue with the new pods because older one will not be deleted until the new ones are healthy.

architecture/2025-03-upgrade-design-doc.md
```diff
@@ -0,0 +1,677 @@
...
+   - You CAN upgrade to Control Plane v0.44 using the v0.42 CLI (if v0.42 CLI includes the upgrade feature)
+   - After upgrading, your CLI (v0.42) will still work with basic operations against your v0.44 Control Plane
+   - However, you won't be able to use any new v0.44 features from your v0.42 CLI
+   - For full functionality, you should upgrade your CLI to match the Control Plane version after the upgrade completes
+
+4. **Network connectivity**: The upgrade process requires internet connectivity to pull container images from registries and validate versions. The process is not intended to run in air-gapped environments.
+
+   - An air-gapped environment is one where systems are physically isolated from unsecured networks like the public internet.
+   - These environments are common in high-security scenarios (military, financial, healthcare, government) where external network connectivity is restricted.
+   - Future implementations will need to address version validation without GitHub API access and provide mechanisms for offline image management.
```
willdavsmith, line 42: I don't think we need to mention this

architecture/2025-03-upgrade-design-doc.md
```diff
@@ -0,0 +1,677 @@
...
+| **Pre-flight Checks**    | Validation steps performed before an upgrade to ensure prerequisites are met and the system is in a valid state for upgrading.                                                                                                                                                                                                                                               |
+| **Version Skipping**     | The ability to upgrade directly from one version to a newer non-adjacent version (e.g., v0.40 → v0.44) without installing intermediate versions.                                                                                                                                                                                                                             |
+
+## Assumptions
+
+1. **User permissions**: Users running the upgrade command have enough permissions on both the Kubernetes cluster and the Radius installation.
```
nicolejms, line 25: I'm ok with this assumption if this document discusses what happens when permissions are not adequate (i.e permission checking and error messages/behavior).

architecture/2025-03-upgrade-design-doc.md
```diff
@@ -0,0 +1,677 @@
...
+2. Existing installation detection
+3. Database connectivity
+4. Custom configuration validation
+
+**User Data Backup and Restore System:**
+
+Rather than taking complete snapshots of the underlying databases (etcd/PostgreSQL), we'll implement a more targeted approach that backs up only the user application metadata and configuration that Radius manages:
+
+```go
+type UserDataBackup interface {
```
sylvainsf, line 416: Are these interfaces implemented at a storage level? So one for etcd, one for Postgres?

architecture/2025-03-upgrade-design-doc.md
nicolejms, line 27: Will this be validated before the upgrade happens? If not, what is the behavior if there are insufficient resources?

architecture/2025-03-upgrade-design-doc.md
```diff
@@ -0,0 +1,677 @@
...
+
+1. **User permissions**: Users running the upgrade command have enough permissions on both the Kubernetes cluster and the Radius installation.
+
+2. **Resource requirements**: The Kubernetes cluster has sufficient compute resources (CPU, memory) to run both the existing and new version components during the rolling upgrade process.
+
+3. **CLI version compatibility**: User has a Radius CLI version that includes the `rad upgrade kubernetes` feature. While older CLIs can't perform upgrades, newer CLIs maintain backward compatibility with older control planes.
+
+   Example: If you have CLI v0.42 and Control Plane v0.42:
+
+   - You CAN upgrade to Control Plane v0.44 using the v0.42 CLI (if v0.42 CLI includes the upgrade feature)
```
kachawla, line 33: Are there any edge that we should think about with this approach? A lot of the operations are currently performed on the client side, so I wonder if this would break in ways we haven't thought through.

architecture/2025-03-upgrade-design-doc.md
```diff
@@ -0,0 +1,677 @@
...
+
+1. **Additional Complexity**: Implementing backup/restore functionality adds complexity to the codebase
+1. **Limited Control**: Users have less granular control compared to manually upgrading components
+1. **Resource Requirements**: The upgrade process temporarily requires additional resources during the transition period
+1. **Upgrade Path Constraints**: Some version combinations may require intermediate upgrades, limiting direct jump capability
+1. **CLI Version Mismatch**: Potential confusion for users when their CLI version doesn't match the server version
+
+#### Proposed Option
+
+I recommend implementing the helm-based upgrade approach with automatic user data backup/restore functionality (we can even discuss if we would like this functionality in version 1 or not) because:
```
sylvainsf, line 501: What alternatives are there to helm? Could we abstract this in a way where if we later run the control plane on a serverless compute we could implement upgrades differently from helm?

architecture/2025-03-upgrade-design-doc.md
```diff
@@ -0,0 +1,677 @@
...
+### Goals
+
+- **Simplify upgrade process**: Provide a single CLI command (`rad upgrade kubernetes`) to upgrade Radius without manual reinstallation.
+- **Ensure data safety**: Implement automatic user data backups before upgrades and restore capability if failures occur.
+- **Minimize downtime**: Use rolling upgrades where possible to keep Radius control plane available during the upgrade process.
+- **Preserve application continuity**: Ensure that user applications continue running without interruption throughout the Radius upgrade process.
+
+### Non goals
+
+- **Downgrade support**: The upgrade process is designed to move forward to newer versions only. Downgrading to previous versions is not supported.
```
brooke-hamilton, line 61: Would we provide guidance on how to manually roll back to a previous version? This could be just documentation.

architecture/2025-03-upgrade-design-doc.md
```diff
@@ -0,0 +1,677 @@
...
+2. **Resource requirements**: The Kubernetes cluster has sufficient compute resources (CPU, memory) to run both the existing and new version components during the rolling upgrade process.
+
+3. **CLI version compatibility**: User has a Radius CLI version that includes the `rad upgrade kubernetes` feature. While older CLIs can't perform upgrades, newer CLIs maintain backward compatibility with older control planes.
+
+   Example: If you have CLI v0.42 and Control Plane v0.42:
+
+   - You CAN upgrade to Control Plane v0.44 using the v0.42 CLI (if v0.42 CLI includes the upgrade feature)
+   - After upgrading, your CLI (v0.42) will still work with basic operations against your v0.44 Control Plane
+   - However, you won't be able to use any new v0.44 features from your v0.42 CLI
+   - For full functionality, you should upgrade your CLI to match the Control Plane version after the upgrade completes
```
nicolejms, line 36: can you upgrade cli before control plane?

architecture/2025-03-upgrade-design-doc.md
ytimocin: All data other than user data that is in etcd or Postgres. I will add this to the doc.

architecture/2025-03-upgrade-design-doc.md
willtsai, line 27: Make sure to publish these requirements in the Radius docs.

architecture/2025-03-upgrade-design-doc.md
```diff
@@ -0,0 +1,677 @@
...
+
+4. **Network connectivity**: The upgrade process requires internet connectivity to pull container images from registries and validate versions. The process is not intended to run in air-gapped environments.
```
kachawla, line 38: To confirm, this assumption applies to Radius installation as well today, right?

architecture/2025-03-upgrade-design-doc.md
sylvainsf: I think we'd have to revisit the whole concept of radius install anyway, so maybe think about this but it's not something we need to solve now.

architecture/2025-03-upgrade-design-doc.md
ytimocin: Yes. We will need to implement this for all the data store implementations we have.

architecture/2025-03-upgrade-design-doc.md
nicolejms, line 61: If upgrade is "bad" we need an option to rollback

architecture/2025-03-upgrade-design-doc.md
ytimocin: I added that as an assumption but we can add the logic to check for enough resources (predefined values) and not start the process if cluster has lower than that.

architecture/2025-03-upgrade-design-doc.md
```diff
@@ -0,0 +1,677 @@
...
+| **Version Skipping**     | The ability to upgrade directly from one version to a newer non-adjacent version (e.g., v0.40 → v0.44) without installing intermediate versions.                                                                                                                                                                                                                             |
+
+## Assumptions
+
+1. **User permissions**: Users running the upgrade command have enough permissions on both the Kubernetes cluster and the Radius installation.
+
+2. **Resource requirements**: The Kubernetes cluster has sufficient compute resources (CPU, memory) to run both the existing and new version components during the rolling upgrade process.
+
+3. **CLI version compatibility**: User has a Radius CLI version that includes the `rad upgrade kubernetes` feature. While older CLIs can't perform upgrades, newer CLIs maintain backward compatibility with older control planes.
```
lakshmimsft, line 29: Can we discuss further if/how we are proposing backward-compatibility?

architecture/2025-03-upgrade-design-doc.md
ytimocin: Yes. We first download a version of Radius to our local computer and then run `rad install kubernetes` which installs the control plane for that version.

architecture/2025-03-upgrade-design-doc.md
```diff
@@ -0,0 +1,677 @@
...
+   - Integrate with version validation and custom configuration handling.
+   - Add comprehensive unit tests for this functionality.
+   - This task can be worked on in parallel with items 1-2, 4. It is a blocker for item 6.
+
+4. **User Data Backup and Restore Interfaces**
+
+   - Define two new interfaces in the `components/database` package:
+     - `UserDataBackup`: Responsible for creating backups of user data before the upgrade.
+     - `UserDataRestore`: Responsible for restoring data from the backup in case of rollback.
+   - Design versioned backup formats to handle schema migrations between versions.
```
sylvainsf, line 613: etcd has snapshots with a lock, but doesn't have a concept of a transaction lock where we could do a snapshot and perform the upgrade in the same lock so we would want to block writes at a higher level. In Postgres we can do a transaction with rollback preventing the need for a backup at all, or enable an optional backup but it seems redundant.

architecture/2025-03-upgrade-design-doc.md
```diff
@@ -0,0 +1,677 @@
...
+- **Unit Tests**: Cover all new code paths, especially backup and restore logic, upgrade logic, and error handling.
+- **Functional Tests**: Validate end-to-end upgrade scenarios, including successful upgrades, upgrades with custom configurations, failure scenarios, and rollback procedures.
+- **Compatibility Tests**: Verify compatibility between different Radius CLI versions and control plane components.
+
+## Open Questions
+
+- **Rollback reliability**: What specific scenarios could cause rollback to fail, and how should we mitigate these risks?
+  - Do we need rollbacks in version 1?
+  - If the answer to the item above is yes, then do we need to add the implementation of the interfaces for Postgres in version 1?
+- **Version skipping limits**: Should we enforce incremental upgrades for certain major version jumps, or always allow direct version skipping?
```
sylvainsf, line 661: We should require incremental upgrades, this will be tricky to skip versions in Postgres.

architecture/2025-03-upgrade-design-doc.md
ytimocin: I was also thinking about this option of not adding the backup functionality by trying to solve the problem in the data layer. We can discuss further.

architecture/2025-03-upgrade-design-doc.md
```diff
@@ -0,0 +1,677 @@
...
+  ✓ Universal Control Plane
+  ✓ Deployment Engine
+  ✓ Applications Resource Provider
+  ✓ Controller
+Performing post-upgrade verification...
+  ✓ All components healthy
+  ✓ Database schema updated successfully
+
+Upgrade complete! Radius has been successfully upgraded to v0.44.0.
+Note: Your local Radius CLI is still v0.40.0. To upgrade your CLI, download the latest version.
```
willtsai, line 251: consider pointing users to the relevant docs page for upgrading their CLI.

architecture/2025-03-upgrade-design-doc.md
```diff
@@ -0,0 +1,677 @@
...
+1. System acknowledges custom configuration parameters
+1. Pre-flight checks validate both upgrade path and configuration
+1. Control plane components are upgraded with custom settings
+1. Verification confirms both health and configuration application
+
+**Exceptions:**
+
+1. If custom configuration parameters are invalid (or should it be ignored?)
+
+#### Scenario 3: Handling upgrade failure and recovery
```
nicolejms, line 174: How do we handle rollback after a schema change?

architecture/2025-03-upgrade-design-doc.md
ytimocin: Because of migrations?

architecture/2025-03-upgrade-design-doc.md
```diff
@@ -0,0 +1,677 @@
...
+
+## Open Questions
+
+- **Rollback reliability**: What specific scenarios could cause rollback to fail, and how should we mitigate these risks?
+  - Do we need rollbacks in version 1?
+  - If the answer to the item above is yes, then do we need to add the implementation of the interfaces for Postgres in version 1?
+- **Version skipping limits**: Should we enforce incremental upgrades for certain major version jumps, or always allow direct version skipping?
+- **Upgrade notifications**: How should we notify users clearly about the CLI version mismatch after upgrading the control plane?
+- **Resource constraints**: How do we handle scenarios where the cluster lacks sufficient resources to perform a rolling upgrade?
+
```
willdavsmith, line 664: One thing that I'm not seeing mentioned is how we handle breaking changes. we're still in 0.x versions, and sometimes breaking changes might happen in the resource specs or in other places

architecture/2025-03-upgrade-design-doc.md
willtsai: but in general I like this warning to notify them of the mismatched versions

architecture/2025-03-upgrade-design-doc.md
ytimocin: Depending on how we handle the schema changes, we are going to do a rollback. For example, some db migration tools require two files: up.sql and down.sql. We will call down.sql to do the rollback.

architecture/2025-03-upgrade-design-doc.md
brooke-hamilton, line 413: What other data won't be backed up?

architecture/2025-03-upgrade-design-doc.md
```diff
@@ -0,0 +1,677 @@
...
+**User Experience:**
+
+```bash
+# Check current version (significantly behind latest)
+> rad version
+RELEASE   VERSION   BICEP     COMMIT
+0.40.0    v0.40.0   0.31.93   1dd17270ec9bc9e725f314fa62c249406034edda
+
+# Upgrade directly to latest version
```
nicolejms, line 230: Need to discuss the max number of versions that can be skipped. This becomes problematic if schemas change frequently. Also rollback becomes trickier with every version skip. how do we handle breaking changes?

architecture/2025-03-upgrade-design-doc.md
```diff
@@ -0,0 +1,677 @@
...
+  ✓ All components healthy
+  ✓ Database schema updated successfully
+
+Upgrade complete! Radius has been successfully upgraded to v0.44.0.
+Note: Your local Radius CLI is still v0.40.0. To upgrade your CLI, download the latest version.
+```
+
+**Result:**
+
+1. System detects a multi-version upgrade scenario
```
nicolejms, line 256: how do we know this?

architecture/2025-03-upgrade-design-doc.md
```diff
@@ -0,0 +1,677 @@
...
+1. Pre-flight checks validate the upgrade is possible
+1. System automatically creates a user data backup for recovery
+1. All control plane components are upgraded in sequence
+1. Post-upgrade verification confirms system health
+1. User is notified about the CLI version mismatch
+
+**Exceptions:**
+
+1. If version check fails because the target is lower than current version
+2. If components fail health checks after upgrade
```
kachawla, line 127: How does a user recover from this?

architecture/2025-03-upgrade-design-doc.md
```diff
@@ -0,0 +1,677 @@
...
+**Result:**
+
+1. System acknowledges custom configuration parameters
+1. Pre-flight checks validate both upgrade path and configuration
+1. Control plane components are upgraded with custom settings
+1. Verification confirms both health and configuration application
+
+**Exceptions:**
+
+1. If custom configuration parameters are invalid (or should it be ignored?)
```
kachawla, line 172: What's the benefit of ignoring invalid config?

architecture/2025-03-upgrade-design-doc.md
```diff
@@ -0,0 +1,677 @@
...
+```go
+// VersionValidator interface
+type VersionValidator interface {
+    ValidateTargetVersion(currentVersion, targetVersion string) error
+    GetLatestVersion() (string, error)
+    IsValidVersion(version string) bool
+}
+```
+
+This interface will be implemented (or existing will be improved) to handle version comparisons, prevent downgrades, and resolve the "latest" version tag to a specific version number. The implementation will (probably) connect to the GitHub API to fetch available release versions when needed.
```
nicolejms, line 361: How will we track information about "compatibility" between versions?

architecture/2025-03-upgrade-design-doc.md
```diff
@@ -0,0 +1,677 @@
...
+```
+
+Checks will include:
+
+1. Version compatibility verification
+2. Existing installation detection
+3. Database connectivity
+4. Custom configuration validation
+
+**User Data Backup and Restore System:**
```
brooke-hamilton, line 411: What is the underlying implementation of an etcd backup? Where does the backup data go? Can it be found and deleted by the user? Backups can create a security issue, depending on what is contained in the backup.

brooke-hamilton (commented):
🥇

architecture/2025-03-upgrade-design-doc.md
ytimocin: I think we can add a breaking change detection logic and I believe at that point instead of an upgrade we can offer the user to uninstall and reinstall. What do you think the flow be if there is a breaking change?

willtsai:
Would upgrades via GitOps trigger this upgrade flow? I think it should :)

architecture/2025-03-upgrade-design-doc.md
ytimocin: We know the version in the cluster and we know the version that is passed in the CLI command.

architecture/2025-03-upgrade-design-doc.md
```diff
@@ -0,0 +1,677 @@
...
+ERROR: Upgrade failed during Applications Resource Provider update.
+Initiating automatic rollback to v0.43.0...
+  ✓ Restoring from backup (Not sure if this is needed)
+  ✓ Universal Control Plane reverted
+  ✓ Deployment Engine reverted
+  ✓ System verification complete
+  ✓ ...
+
+Rollback complete. System has been restored to v0.43.0.
+Review Kubernetes events and logs for more details on the failure.
```
kachawla, line 204: Do we want the user to switch back and forth between rad cli and kubectl? Can this be achieved through `rad debug-logs`?

architecture/2025-03-upgrade-design-doc.md
ytimocin: Everything other than Radius user related data.

architecture/2025-03-upgrade-design-doc.md
sylvainsf: We should think about this when we implement RBAC

architecture/2025-03-upgrade-design-doc.md
nicolejms, line 380: Since we know we want to (eventually) move the control plane off of K8s, should we take a dependency on another feature?

ytimocin:
> Would upgrades via GitOps trigger this upgrade flow? I think it should :)

I can use @willdavsmith 's help on this question :)

architecture/2025-03-upgrade-design-doc.md
```diff
@@ -0,0 +1,677 @@
...
+**Result:**
+
+1. System detects failure during the upgrade process
+1. Automatic restore is initiated using the pre-upgrade backup
+1. All components are restored to their previous state
+1. User is informed of the failure and suggested next steps
+
+**Exceptions:**
+
+1. If the user data backup restoration fails (rare but possible)
```
kachawla, line 216: What's the path to recovery from this state?

architecture/2025-03-upgrade-design-doc.md
ytimocin: @sylvainsf mentioned the same thing and I believe that is a fair point. We can do this in the DB level.

architecture/2025-03-upgrade-design-doc.md
sylvainsf: Most of the work is around proper locks, we can discuss more

architecture/2025-03-upgrade-design-doc.md
```diff
@@ -0,0 +1,677 @@
...
+    Run(ctx context.Context) (bool, string, error)
+    Name() string
+    // Error, Warning, Info
+    Severity() CheckSeverity
+}
+```
+
+Checks will include:
+
+1. Version compatibility verification
```
nicolejms, line 406: wouldn't we also want to validate permissions, resources, etc (i.e. all your assumptions are true)?

architecture/2025-03-upgrade-design-doc.md
```diff
@@ -0,0 +1,677 @@
...
+
+1. System detects a multi-version upgrade scenario
+1. Pre-flight checks validate that direct upgrade is supported
+1. All components are upgraded to the target version
+1. Verification confirms system health with the new version
+
+**Exceptions:**
+
+1. If direct upgrade path isn't supported between versions
+1. If database migrations encounter issues (this may be the case when we introduce Postgres as the data store)
```
kachawla, line 264: Why is this specific to postgres?

architecture/2025-03-upgrade-design-doc.md
sylvainsf: Wouldn't this require us to have the ability to roll back the database state changes? Otherwise new pods wouldn't be healthy until the change is made, and rolling back pods wouldn't work until the change is reverted.

architecture/2025-03-upgrade-design-doc.md
```diff
@@ -0,0 +1,677 @@
...
+      Controller["Controller"]
+      DynamicRP["Dynamic RP"]
+  end
+  ControlPlane -->|Stores Metadata| Database["etcd"]
+  ControlPlane -->|Interacts with| Dependencies["Dependencies (Dapr, Contour)"]
+  CLI -->|Logs Progress| User["User"]
+  CLI -->|Performs Pre-flight Checks| PreFlight["Pre-flight Checks"]
+  PreFlight -->|Validates| KubernetesAPI
+  CLI -->|Creates User Data Backup| Backup["User Data Backup"]
+  Backup -->|Restores on Failure| Restore["Restore Mechanism"]
```
kachawla, line 289: These should happen on the server side, client doesn't store the data today.

architecture/2025-03-upgrade-design-doc.md
```diff
@@ -0,0 +1,677 @@
...
+      ARP["Applications RP"]
+      Controller["Controller"]
+      DynamicRP["Dynamic RP"]
+  end
+  ControlPlane -->|Stores Metadata| Database["etcd"]
+  ControlPlane -->|Interacts with| Dependencies["Dependencies (Dapr, Contour)"]
+  CLI -->|Logs Progress| User["User"]
+  CLI -->|Performs Pre-flight Checks| PreFlight["Pre-flight Checks"]
+  PreFlight -->|Validates| KubernetesAPI
+  CLI -->|Creates User Data Backup| Backup["User Data Backup"]
```
kachawla, line 288: Where is the backup stored?

architecture/2025-03-upgrade-design-doc.md
sylvainsf: There's a distinct difference between a rollback and down migration. Rollback is automatic when the database change fails, down migration is something a user does to undo a database change.

architecture/2025-03-upgrade-design-doc.md
```diff
@@ -0,0 +1,677 @@
...
+
+const (
+    StatusHealthy     HealthStatus = "Healthy"
+    StatusDegraded    HealthStatus = "Degraded"
+    StatusUnavailable HealthStatus = "Unavailable"
+    StatusUnknown     HealthStatus = "Unknown"
+)
+```
+
+#### Advantages (of each option considered)
```
nicolejms, line 483: what other options were considered?

architecture/2025-03-upgrade-design-doc.md
ytimocin: Right now install (including reinstall) and uninstall happen using Helm. I also think that we should revisit all when we need to in future. Adding something other than Helm should be out of scope for this feature.

architecture/2025-03-upgrade-design-doc.md
```diff
@@ -0,0 +1,677 @@
...
+# Upgrade directly to latest version
+> rad upgrade kubernetes --version latest
+
+Initiating Radius upgrade from v0.40.0 to v0.44.0 (latest stable)...
+Pre-flight checks:
+  ✓ Valid version target
+  ✓ Multiple version jump detected (v0.40.0 → v0.44.0)
+  ✓ Compatible upgrade path confirmed
+  ✓ Database schema changes detected
+Creating backup of current user data...
```
lakshmimsft, line 239: Need more details on what userdata is backedup.
Okay.. i see this will be addressed below

architecture/2025-03-upgrade-design-doc.md
sylvainsf: a helper function that lists git tags would tell us all releases between installed and the CLI attempting the upgrade.

architecture/2025-03-upgrade-design-doc.md
ytimocin: Not sure how UCP can go and ask Kubernetes about its remaining resources. I can check that.

architecture/2025-03-upgrade-design-doc.md
```diff
@@ -0,0 +1,677 @@
...
+
+### Error Handling
+
+The upgrade process will implement the following error handling strategies:
+
+1. **Pre-flight Validation**: Catch incompatibility issues before starting the upgrade
+2. **Graceful Timeouts**: All operations will respect user-defined or default timeouts
+3. **Automatic Rollback**: Failed upgrades trigger automatic restoration of previous state
+4. **Detailed Error Reporting**: Clear error messages with troubleshooting guidance
+5. **Idempotent Operations**: Commands can be safely retried after addressing issues
```
nicolejms, line 544: does this mean we'll rollback on every failure or that the system can have varying degrees of upgrade completed?

architecture/2025-03-upgrade-design-doc.md
```diff
@@ -0,0 +1,677 @@
...
+6. **Health Verification**: Component readiness and health check mechanisms
+
+All components will follow Radius coding standards and include comprehensive unit tests.
+
+### Error Handling
+
+The upgrade process will implement the following error handling strategies:
+
+1. **Pre-flight Validation**: Catch incompatibility issues before starting the upgrade
+2. **Graceful Timeouts**: All operations will respect user-defined or default timeouts
```
nicolejms, line 541: how do we deal with timeouts?  always rollback?

architecture/2025-03-upgrade-design-doc.md
kachawla, line 413: What type of data is not included in this?

architecture/2025-03-upgrade-design-doc.md
```diff
@@ -0,0 +1,677 @@
...
+    ReleaseLock(ctx context.Context) error
+
+    // Checks if an upgrade is currently in progress
+    IsUpgradeInProgress(ctx context.Context) (bool, error)
+}
+```
+
+We can utilize Kubernetes Lease objects (coordination.k8s.io/v1) for implementing the distributed locking mechanism (open to discussion and suggestions). Leases are purpose-built for this use case, providing built-in lease duration and automatic expiration capabilities. For more information, see: <https://kubernetes.io/docs/concepts/architecture/leases/>.
+
+Other CLI commands (`rad deploy app.bicep`, `rad delete app my-app` or other data-changing commands) that modify data will check for this lock before proceeding:
```
kachawla, line 382: Locks should be enforced on the server side to prevent race conditions, as client-side locking may not effectively handle concurrent requests where client mutations and upgrades overlap.

architecture/2025-03-upgrade-design-doc.md
ytimocin: Update our design template by adding `how to do upgrade`

architecture/2025-03-upgrade-design-doc.md
ytimocin: I am leaning towards doing incremental upgrades. I will update the doc accordingly.

architecture/2025-03-upgrade-design-doc.md
ytimocin: This was discussed in the meeting. We don't have a lot of options available other than Helm. Since we are using Helm now, it is the easiest and we are going to go with that. Updated the section titles per request.

architecture/2025-03-upgrade-design-doc.md
ytimocin: Yes, we need to be able to pull the images during installation.

architecture/2025-03-upgrade-design-doc.md
ytimocin: Probably. The outputs will be updated. These are just placeholders for now.

architecture/2025-03-upgrade-design-doc.md
ytimocin: It is going to be moved to the data store layer.

architecture/2025-03-upgrade-design-doc.md
ytimocin: Good point. Users may hit mid-upgrade failures with no clear information. To fix this, we can:
- Update the Assumptions section (in docs or somewhere that the users can see) to call out about the minimum requirements for the `rad upgrade kubernetes` process
- Add ResourceAvailability pre-flight check that can measure the resources in the cluster and abort the process if there is not enough
- Improve the CLI UX so that on failure it can mention that there is not enough resources

We can apply all these steps and I will update the doc accordingly.

ytimocin: We can add a pre-flight check that checks the resource availability.

architecture/2025-03-upgrade-design-doc.md
ytimocin: In one of the future versions, we will be keeping a list of most recent successful Radius installations and and introduce a new command for rollback to the most recent happy version.

architecture/2025-03-upgrade-design-doc.md
ytimocin: I had added that as a note to myself. After playing around with the code, we shouldn't ignore and just error out.

architecture/2025-03-upgrade-design-doc.md
ytimocin: What I am talking about here is that we will have a migrations folder that will have all of our migrations (up and down files). For example: add_resources_table.up, add_resources_table.down, add_new_column_to_resources_table.up, and add_new_column_to_resources_table.down etc. We will need to find a way to do this in etcd.

architecture/2025-03-upgrade-design-doc.md
ytimocin: There may be 3 possible ways to recover from this state:

1. Manual rollback
2. Troubleshoot and retry

Let's please keep in mind that the user data will still be available in the data store. This is just the backup restoration failure we are talking about. It is not the actual data.

architecture/2025-03-upgrade-design-doc.md
ytimocin: Kubernetes objects may be too platform specific. But for now, I believe we are going to keep them in Kubernetes object like a PVC (https://kubernetes.io/docs/concepts/storage/persistent-volumes/)

architecture/2025-03-upgrade-design-doc.md
ytimocin: Added that in the development plan

architecture/2025-03-upgrade-design-doc.md
superbeeny: this can get really thorny if the db schema changes and data has to be rolled back as well

architecture/2025-03-upgrade-design-doc.md
```diff
@@ -0,0 +1,793 @@
...
+
+1. If direct upgrade path isn't supported between versions
+1. If database migrations encounter issues (this may be the case when we introduce Postgres as the data store)
+1. If intermediate upgrades are required first
+
+## Design
+
+### High-Level Design Diagram
+
+```mermaid
```
brooke-hamilton, line 270: This diagram appears to indicate that the user data backup runs in parallel to preflight checks but does not have to complete before the helm upgrade is applied. Is that the intended flow?

architecture/2025-03-upgrade-design-doc.md
nicolejms: I don't agree this is a non-goal. We need the ability to rollback even if the upgrade succeeds.  If we introduce a regression, users must be able to rollback to a last known good state.

architecture/2025-03-upgrade-design-doc.md
```diff
@@ -0,0 +1,793 @@
...
+### Goals
+
+- **Simplify upgrade process**: Provide a single CLI command (`rad upgrade kubernetes`) to upgrade Radius without manual reinstallation.
+- **Ensure data safety**: Implement automatic user data backups before upgrades and restore capability if failures occur.
+- **Minimize downtime**: Use rolling upgrades where possible to keep Radius control plane available during the upgrade process.
+- **Preserve application continuity**: Ensure that user applications continue running without interruption throughout the Radius upgrade process.
+
+### Non goals
+
+- **Downgrade support**: The upgrade process is designed to move forward to newer versions only. Downgrading to previous versions is not supported.
```
willtsai, line 60: one of the use cases I've heard is that an upgrade may go successfully from a Radius control plane perspective, but then subsequently result in test failures in other parts of the stack (e.g. if Radius is built into a developer platform that consists of many components downstream from Radius). how will we address or advise users for this?

ytimocin: In one of the future iterations, we will add a new command to rollback to the most recent successful version of Radius. A command like `rad rollback kubernetes`. Would that answer your question?

architecture/2025-03-upgrade-design-doc.md
```diff
@@ -0,0 +1,793 @@
...
+    // IsUpgradeInProgress returns true if a valid (non‑stale) lock is held by another process.
+    IsUpgradeInProgress(ctx context.Context) (bool, error)
+}
+```
+
+**Timeouts:** callers must supply a context with a finite deadline (e.g. 2 min) to avoid blocking forever.
+**Stale‑lock detection:** each lock has a TTL/heartbeat; expired leases are auto‑cleaned before AcquireLock.
+Force cleanup: --force flag allows manual removal of stale/orphaned locks.
+
+Usage in CLI commands:
```
kachawla, line 384: https://github.com/radius-project/design-notes/pull/87#discussion_r2032108184 would still be an issue if the we are checking the lock in the CLI.

architecture/2025-03-upgrade-design-doc.md
```diff
@@ -0,0 +1,793 @@
...
+
+   - You CAN upgrade to Control Plane v0.44 using the v0.42 CLI (if v0.42 CLI includes the upgrade feature)
+   - After upgrading, your CLI (v0.42) will still work with basic operations against your v0.44 Control Plane
+   - However, you won't be able to use any new v0.44 features from your v0.42 CLI
+   - For full functionality, you should upgrade your CLI to match the Control Plane version after the upgrade completes
+
+4. **Network connectivity**: The upgrade process requires internet connectivity to pull container images from registries and validate versions. The process is not intended to run in air-gapped environments.
+
+   - An air-gapped environment is one where systems are physically isolated from unsecured networks like the public internet.
+   - These environments are common in high-security scenarios (military, financial, healthcare, government) where external network connectivity is restricted.
```
willtsai, line 41: what would be our stance for upgrading Radius in an air-gapped environment? will there be an option to point a radius upgrade to use container images from predownloaded or custom registries?

architecture/2025-03-upgrade-design-doc.md
```diff
@@ -0,0 +1,793 @@
...
+```go
+inProgress, _ := upgradeLock.IsUpgradeInProgress(ctx)
+if inProgress {
+    return errors.New("An upgrade is currently in progress. Please try again after the upgrade completes.")
+}
+```
+
+**Pre-flight Check System:**
+
+Pre-flight checks run before any changes are made to ensure the upgrade can proceed safely.
```
kachawla, line 416: Where will these checks run?

architecture/2025-03-upgrade-design-doc.md
```diff
@@ -0,0 +1,793 @@
...
+    // Error, Warning, Info
+    Severity() CheckSeverity
+}
+```
+
+Checks will include:
+
+1. Version compatibility verification
+2. Existing installation detection
+3. Database connectivity
```
kachawla, line 431: What do we mean by this?

architecture/2025-03-upgrade-design-doc.md
```diff
@@ -0,0 +1,793 @@
...
+
+1. **User permissions**: Users running the upgrade command have enough permissions on both the Kubernetes cluster and the Radius installation.
+
+2. **Resource requirements**: The Kubernetes cluster has sufficient compute resources (CPU, memory) to run both the existing and new version components during the rolling upgrade process.
+
+3. **CLI version compatibility**: User has a Radius CLI version that includes the `rad upgrade kubernetes` feature. While older CLIs can't perform upgrades, newer CLIs maintain backward compatibility with older control planes.
+
+   Example: If you have CLI v0.42 and Control Plane v0.42:
+
+   - You CAN upgrade to Control Plane v0.44 using the v0.42 CLI (if v0.42 CLI includes the upgrade feature)
```
superbeeny, line 33: Can we update the CLI as part of the process?

architecture/2025-03-upgrade-design-doc.md
```diff
@@ -0,0 +1,793 @@
...
+    Severity() CheckSeverity
+}
+```
+
+Checks will include:
+
+1. Version compatibility verification
+2. Existing installation detection
+3. Database connectivity
+4. Custom configuration validation
```
kachawla, line 432: Can you expand a bit on what is custom configuration?

architecture/2025-03-upgrade-design-doc.md
```diff
@@ -0,0 +1,793 @@
...
+  PreFlight -->|Validates| KubernetesAPI
+  CLI -->|Creates User Data Backup| Backup["User Data Backup"]
+  Backup -->|Restores on Failure| Restore["Restore Mechanism"]
+```
+
+- **Important Note:** As of April 2025, Postgres is not fully implemented yet as the data store of Radius. We use etcd in production.
+
+### Architecture Diagram
+
+```mermaid
```
brooke-hamilton, line 295: Could this and the previous diagram be merged into one component diagram that shows relationships and responsibilities?

architecture/2025-03-upgrade-design-doc.md
ytimocin: Yes, exactly. We will be responsible for adding all the necessary parameters to the necessary commands so that users can point to their own registries and/or specific images.

architecture/2025-03-upgrade-design-doc.md
```diff
@@ -0,0 +1,793 @@
...
+    ComponentUpgrade --> Controller
+    ComponentUpgrade --> Dashboard
+
+    UCP -->|"Reads/Writes"| Database["etcd / PostgreSQL"]
+    UCP -.->|"Only holds metadata"| UserApps["User Applications"]
+```
+
+### Detailed Design
+
+```mermaid
```
brooke-hamilton, line 328: This is great - very clear process.

architecture/2025-03-upgrade-design-doc.md
```diff
@@ -0,0 +1,793 @@
...
+2. Existing installation detection
+3. Database connectivity
+4. Custom configuration validation
+
+**User Data Backup and Restore System:**
+
+Rather than taking complete snapshots of the underlying databases (etcd/PostgreSQL), we'll implement a more targeted approach that backs up only the user application metadata and configuration that Radius manages:
+
+- **Included in backup**: User application, environment, recipe definitions, and all other resources that the user has deployed/added via Radius.
+- **Not included in backup**: Anything other than user data in the data store.
```
kachawla, line 439: Can you share example of this data that's not user data?

architecture/2025-03-upgrade-design-doc.md
ytimocin: Yes, I added that later in the document. I will remove this non-goal. As discussed with you yesterday, we will introduce a `rad rollback kubernetes` in one of the future iterations of this work.

architecture/2025-03-upgrade-design-doc.md
```diff
@@ -0,0 +1,793 @@
...
+
+```go
+type UserDataBackup interface {
+    // Creates a backup of all user application metadata and configurations
+    BackupUserData(ctx context.Context) (BackupID string, err error)
+    // Lists available user data backups with metadata
+    ListBackups(ctx context.Context) ([]BackupInfo, error)
```
kachawla, line 446: What is the purpose of list?

architecture/2025-03-upgrade-design-doc.md
```diff
@@ -0,0 +1,793 @@
...
+```go
+// VersionValidator interface
+type VersionValidator interface {
+    ValidateTargetVersion(currentVersion, targetVersion string) error
+    GetLatestVersion() (string, error)
+    IsValidVersion(version string) bool
+}
+```
+
+This interface will be implemented (or existing will be improved) to handle version comparisons, prevent downgrades, and resolve the "latest" version tag to a specific version number. The implementation will (probably) connect to the GitHub API to fetch available release versions when needed.
```
brooke-hamilton, line 360: We need a way to check the version in an airgapped environment without the GitHub API. Maybe use the currently configured OCI container repo?

architecture/2025-03-upgrade-design-doc.md
ytimocin: Like the work we have been doing here: https://github.com/radius-project/radius/pull/9189. And this is the sh file to run `rad install` w/o internet access: https://gist.github.com/ytimocin/8887d95ab1409562f4646fd30edb101c.

architecture/2025-03-upgrade-design-doc.md
```diff
@@ -0,0 +1,793 @@
...
+RELEASE   VERSION   BICEP     COMMIT
+0.40.0    v0.40.0   0.31.93   1dd17270ec9bc9e725f314fa62c249406034edda
+
+# Upgrade directly to latest version
+> rad upgrade kubernetes --version latest
+
+Initiating Radius upgrade from v0.40.0 to v0.44.0 (latest stable)...
+Pre-flight checks:
+  ✓ Valid version target
+  ✓ Multiple version jump detected (v0.40.0 → v0.44.0)
```
superbeeny, line 235: would the simplest option here be applying each release in turn?

architecture/2025-03-upgrade-design-doc.md
```diff
@@ -0,0 +1,793 @@
...
+
+6. **Storage availability**: The Kubernetes cluster has sufficient persistent storage capacity for backup operations during the upgrade process.
+
+## Objectives
+
+> **Issue Reference:** <https://github.com/radius-project/radius/issues/8095>
+
+### Goals
+
+- **Simplify upgrade process**: Provide a single CLI command (`rad upgrade kubernetes`) to upgrade Radius without manual reinstallation.
```
willtsai, line 53: Will there be an option to upgrade via GitOps if the user manages their Radius installation via Flux?

brooke-hamilton (commented):
🚀 Nice document 🚀

architecture/2025-03-upgrade-design-doc.md
```diff
@@ -0,0 +1,793 @@
...
+    // Validates that an upgrade to the target version is possible
+    ValidateUpgrade(ctx context.Context, targetVersion string) error
+}
+
+type UpgradeOptions struct {
+    Version string // Target version to upgrade to
+    Values map[string]interface{} // Custom configuration values
+    Timeout time.Duration // Maximum time allowed for upgrade
+
+    EnableUserDataBackup  bool // Whether automatic user data backup is enabled
```
kachawla, line 476: Could you share a scenario when this will be disabled?

architecture/2025-03-upgrade-design-doc.md
ytimocin: To be honest, I haven't explored this option. If GitOps is using Helm upgrade behind the scenes, it may be not that difficult to add. But I would discuss this with @willdavsmith and decide what to do next.

architecture/2025-03-upgrade-design-doc.md
ytimocin: No, that is not right. I will update it.

architecture/2025-03-upgrade-design-doc.md
ytimocin: I will try to do that too. I am a beginner in using MermaidJS.

architecture/2025-03-upgrade-design-doc.md
ytimocin: We can list the Helm releases and get the deployed one from that.

architecture/2025-03-upgrade-design-doc.md
```diff
@@ -0,0 +1,793 @@
...
+
+1. It provides the best balance between user experience simplicity and technical safety
+2. Leverages existing Helm functionality while adding Radius-specific safety features
+3. The automated user data backup/restore mechanism offers critical protection against data loss
+4. Clear progress indicators and health checks give users confidence in the process
+5. This approach aligns with how users currently install/reinstall Radius, providing consistency
+
+### API design (if applicable)
+
+No specific REST API addition is necessary.
```
kachawla, line 537: https://github.com/radius-project/design-notes/blob/9cd00ecbd2898641e68055ad57c01337a680b109/architecture/2025-03-upgrade-design-doc.md?plain=1#L462-L468

I gathered that these will be new APIs, did I misunderstand it?

architecture/2025-03-upgrade-design-doc.md
```diff
@@ -0,0 +1,793 @@
...
+
+Rather than taking complete snapshots of the underlying databases (etcd/PostgreSQL), we'll implement a more targeted approach that backs up only the user application metadata and configuration that Radius manages:
+
+- **Included in backup**: User application, environment, recipe definitions, and all other resources that the user has deployed/added via Radius.
+- **Not included in backup**: Anything other than user data in the data store.
+
+```go
+type UserDataBackup interface {
+    // Creates a backup of all user application metadata and configurations
+    BackupUserData(ctx context.Context) (BackupID string, err error)
```
lakshmimsft, line 444: will there be a pattern/format for this ID. how. in other words, does it tie to a specific backup call run and do we need to identify it later?

architecture/2025-03-upgrade-design-doc.md
```diff
@@ -0,0 +1,793 @@
...
+### Error Handling
+
+The upgrade process will implement the following error handling strategies:
+
+1. **Pre-flight Validation**: Catch incompatibility issues before starting the upgrade
+2. **Graceful Timeouts**: All operations will respect user-defined or default timeouts
+3. **Automatic Rollback**: Failed upgrades trigger automatic restoration of previous state
+4. **Detailed Error Reporting**: Clear error messages with troubleshooting guidance
+5. **Idempotent Operations**: Commands can be safely retried after addressing issues
+6. **Resource Cleanup**: Temporary resources created during the upgrade are properly removed
```
kachawla, line 571: What are the temporary resources created for?

architecture/2025-03-upgrade-design-doc.md
```diff
@@ -0,0 +1,793 @@
...
+
+## Test Plan
+
+### Unit Tests
+
+- Test each interface implementation independently
+
+### Integration Tests
```
kachawla, line 579: Are they going to be run differently than end-to-end tests?

architecture/2025-03-upgrade-design-doc.md
```diff
@@ -0,0 +1,793 @@
...
+### End-to-End Tests
+
+- Perform upgrades across consecutive versions (e.g., v0.43 → v0.44)
+- Test version skipping scenarios (e.g., v0.40 → v0.44)
+- Verify custom configuration is properly applied
+- Simulate failures and verify automatic rollback
+
+## Security
+
+No changes to the existing security model needed.
```
kachawla, line 594: More details about where the backup is stored, who can access it, when it gets deleted are things that would be relevant here

superbeeny:
Nice, very comprehensive and lots of food for thought

architecture/2025-03-upgrade-design-doc.md
sylvainsf: This should be easy to do, for etcd:
https://etcd.io/docs/v3.5/op-guide/recovery/

We just need to take a snapshot and perhaps label it the old version. 

Postgres would be similar. We should document them here as this is distinctly different from a rollback migration. There's some considerations as it will reflect the app graph at the time of the last deployment so we may want to document that a deployment after rollback is required.

sylvainsf (approved):
Thanks for all the work on this!

PR #88, opened 2025-03-26
ACI

brooke-hamilton:
Note: there is a separate extensibility design at PR #91. This document covers adding ACI support.


architecture/2025-04-compute-platform-extensibility.md
```diff
@@ -0,0 +1,505 @@
...
+
+This document describes the initial integration with ACI, which is built upon an earlier demo capability. Our longer-term approach will provide an extensibility capability to Radius for deploying to other compute platforms, which will be described in a separate document.
+
+## Terms and definitions
+
+* ACI: Azure Container Instances, a service that allows users to run Docker containers on Azure without managing servers.
+
+## Objectives
+
+* To integrate Azure Container Instances (ACI) with Radius, enabling users to deploy and manage their applications and containers in ACI environments.
```
ytimocin, line 17: What is the difference between an application and a container? An application will run in a container, right? Are these terms different in ACI?

architecture/2025-04-compute-platform-extensibility.md
```diff
@@ -0,0 +1,505 @@
...
+* To ensure that the initial features are fully functional and ready for demonstration.
+
+> **Issue Reference:** <!-- (If appropriate) Reference an existing issue that describes the feature or bug. -->
+
+### Goals
+
+* Implement the capability for the Radius control plane hosted in Kubernetes to deploy to ACI.
+* Enable users to set up a Radius Environment targeted at ACI using environment definition files (e.g., env.bicep).
+* Allow users to deploy Radius applications to ACI using application definition files (e.g., app.bicep).
+* Support the deployment of application containers to ACI, ensuring that container definitions are platform-agnostic.
```
ytimocin, line 28: We can talk about deploying an application and deploying an application container and their differences.

architecture/2025-04-compute-platform-extensibility.md
```diff
@@ -0,0 +1,505 @@
...
+
+### User scenarios (optional)
+
+#### User story 1
+
+As a user, I can use the instance of the Radius control plane hosted in my Kubernetes cluster to set up an ACI Environment and target that ACI environment to deploy my Applications and Containers.
+
+#### User story 2
+
+As a user, I can define a new Radius Environment that specifies ACI as the underlying compute platform by creating a new Environment definition file (e.g., env.bicep) and specifying the necessary settings before deploying the environment using rad deploy env.bicep. Applications and container resources that I deploy into this Environment will be targeted at my ACI cluster.
```
sk593, line 49: Can we specify the difference between ACI Environment and Radius environment running with an ACI compute (if there is one)? To me, these are the same thing since ACI wouldn't really have it's own environment.

architecture/2025-04-compute-platform-extensibility.md
```diff
@@ -0,0 +1,505 @@
...
+
+## Terms and definitions
+
+* ACI: Azure Container Instances, a service that allows users to run Docker containers on Azure without managing servers.
+
+## Objectives
+
+* To integrate Azure Container Instances (ACI) with Radius, enabling users to deploy and manage their applications and containers in ACI environments.
+* To provide a seamless user experience for deploying Radius applications to ACI, including environment setup, application definition, and deployment.
+* To ensure that the initial features are fully functional and ready for demonstration.
```
kachawla, line 19: What do we mean by initial features?

architecture/2025-04-compute-platform-extensibility.md
```diff
@@ -0,0 +1,505 @@
...
+### Goals
+
+* Implement the capability for the Radius control plane hosted in Kubernetes to deploy to ACI.
+* Enable users to set up a Radius Environment targeted at ACI using environment definition files (e.g., env.bicep).
+* Allow users to deploy Radius applications to ACI using application definition files (e.g., app.bicep).
+* Support the deployment of application containers to ACI, ensuring that container definitions are platform-agnostic.
+* Allow users to apply ACI-specific configurations to containers using extension properties.
+* Adding and configuring Dapr sidecars to ACI containers.
+* A default implementation of Azure Gateways on ACI.
+* Defining and creating Secret Stores for use in ACI containers. (The implementation is TBD: KeyVault or Dapr.)
```
ytimocin, line 32: Are the secrets going to be kept on the Kubernetes cluster next to the Radius installation?

architecture/2025-04-compute-platform-extensibility.md
sk593: I don't think "ACI cluster" is the right terminology. The cluster would still be Kubernetes, but the containers would be ACI

architecture/2025-04-compute-platform-extensibility.md
```diff
@@ -0,0 +1,505 @@
...
+
+### Non goals
+
+* Hosting the Radius control plane in ACI. The control plane will continue to be hosted in Kubernetes.
+* Defining and creating custom ingress for use in ACI containers.
+* Defining Extenders/UDT for use in ACI containers.
```
ytimocin, line 38: Would this be difficult considering we do the other resource types?

architecture/2025-04-compute-platform-extensibility.md
kachawla, line 17: > in ACI environments

Just to clarify - this is not referencing to Radius environment in ACI, but only applications deployment on ACI?

architecture/2025-04-compute-platform-extensibility.md
ytimocin, line 49: I see now that you are using applications and containers as in Applications.Core/applications and Applications.Core/containers.

architecture/2025-04-compute-platform-extensibility.md
```diff
@@ -0,0 +1,505 @@
...
+
+#### User story 3
+
+As a user, I can define a new Radius Application by creating a new Application definition file (e.g., app.bicep) and use this same application definition to deploy to an Environment with either Kubernetes or ACI as its underlying compute platform. Via the extension property in the Application definition, the user can optionally configure custom settings that are specific to ACI and get applied only when deploying to ACI. For any extension properties that are not applicable to ACI, they will be ignored.
+
+#### User story 4
+
+As a user, I can declare a Radius Container within the application definition (e.g., app.bicep) including all the platform-agnostic container properties, e.g., image, env, etc., and deploy the application container to an ACI cluster. Radius will create the necessary resources in ACI for the application, including container instances, NGroups, container group profiles, load balancers, DNS, subnets, etc. Note that the container definitions should be container runtime platform-agnostic, thus this same container definition can be deployed to both Kubernetes and ACI.
```
sk593, line 57: The assumption here is that the environment the application is tied to would have to have been defined with an ACI compute, correct?

brooke-hamilton: Yes - the environment would be specified as ACI.

willdavsmith, line 57: you can define stuff in `extensions` that's platform specific though right? like `kubernetesMetadata`?

architecture/2025-04-compute-platform-extensibility.md
```diff
@@ -0,0 +1,505 @@
...
+
+### Goals
+
+* Implement the capability for the Radius control plane hosted in Kubernetes to deploy to ACI.
+* Enable users to set up a Radius Environment targeted at ACI using environment definition files (e.g., env.bicep).
+* Allow users to deploy Radius applications to ACI using application definition files (e.g., app.bicep).
+* Support the deployment of application containers to ACI, ensuring that container definitions are platform-agnostic.
+* Allow users to apply ACI-specific configurations to containers using extension properties.
+* Adding and configuring Dapr sidecars to ACI containers.
+* A default implementation of Azure Gateways on ACI.
```
kachawla, line 31: I didn't follow this, could you please elaborate?

architecture/2025-04-compute-platform-extensibility.md
```diff
@@ -0,0 +1,505 @@
...
+
+#### User story 5
+
+As a user, upon the successful deployment of my application to my ACI environment, I can view the ACI containers in the CLI using the rad app graph command or via the Application Graph in the Radius Dashboard.
+
+#### User story 6
+
+As a user, I may set ACI-specific configurations for my containers via the extensions or runtimes properties. This allows me to make use of ACI-specific features like confidential containers or spot instances that may not be available on other container compute platforms. Note that any properties that are not applicable to the compute platform in the targeted deployment environment will be ignored.
+
```
sk593, line 66: I think we should add a scenario about specifying a Radius environment with ACI compute during `rad init`. Right now, any default environment created would be Kubernetes compute

architecture/2025-04-compute-platform-extensibility.md
```diff
@@ -0,0 +1,505 @@
...
+## Objectives
+
+* To integrate Azure Container Instances (ACI) with Radius, enabling users to deploy and manage their applications and containers in ACI environments.
+* To provide a seamless user experience for deploying Radius applications to ACI, including environment setup, application definition, and deployment.
+* To ensure that the initial features are fully functional and ready for demonstration.
+
+> **Issue Reference:** <!-- (If appropriate) Reference an existing issue that describes the feature or bug. -->
+
+### Goals
```
willtsai, line 23: might be implicit in the goals below, but perhaps it'd be good to call out that making the comprehensive set of CLI functionality compatible with ACI concepts is a goal too? e.g. `rad deploy, rad workspace, rad env` etc.

architecture/2025-04-compute-platform-extensibility.md
```diff
@@ -0,0 +1,505 @@
...
+
+As a user, I can use the instance of the Radius control plane hosted in my Kubernetes cluster to set up an ACI Environment and target that ACI environment to deploy my Applications and Containers.
+
+#### User story 2
+
+As a user, I can define a new Radius Environment that specifies ACI as the underlying compute platform by creating a new Environment definition file (e.g., env.bicep) and specifying the necessary settings before deploying the environment using rad deploy env.bicep. Applications and container resources that I deploy into this Environment will be targeted at my ACI cluster.
+
+#### User story 3
+
+As a user, I can define a new Radius Application by creating a new Application definition file (e.g., app.bicep) and use this same application definition to deploy to an Environment with either Kubernetes or ACI as its underlying compute platform. Via the extension property in the Application definition, the user can optionally configure custom settings that are specific to ACI and get applied only when deploying to ACI. For any extension properties that are not applicable to ACI, they will be ignored.
```
ytimocin, line 53: Does this mean that we will need to have two environments: one that supports Kubernetes and one that supports ACI?

If we can change the extension property in the Application definition, why would we need the specific Environment?

architecture/2025-04-compute-platform-extensibility.md
```diff
@@ -0,0 +1,505 @@
...
+
+As a user, I may set ACI-specific configurations for my containers via the extensions or runtimes properties. This allows me to make use of ACI-specific features like confidential containers or spot instances that may not be available on other container compute platforms. Note that any properties that are not applicable to the compute platform in the targeted deployment environment will be ignored.
+
+## User Experience (if applicable)
+
+The integration of Azure Container Instances (ACI) with Radius will bring several changes to the user experience, particularly in how users define and deploy their applications. The key changes are as follows:
+
+1. Environment Setup: Users will now be able to set up a Radius Environment targeted at ACI using environment definition files (e.g., env.bicep). This will involve specifying ACI-specific configurations and parameters in the Bicep templates.
+1. Application Definition: Users will define their applications using application definition files (e.g., app.bicep). These files can include ACI-specific properties and configurations, also know as “punch-through”, allowing users to tailor their deployments to the ACI environment.
```
sk593, line 72: We should discuss at what level the punchthroughs will exist. environment level vs app level, etc

architecture/2025-04-compute-platform-extensibility.md
kachawla, line 38: Not sure what we mean by this

architecture/2025-04-compute-platform-extensibility.md
ytimocin, line 57: I think we should talk about (or give examples of) environment definition, application definition, and container definition. Which ones support which runtimes? We should also talk about all the different scenarios that can occur.

architecture/2025-04-compute-platform-extensibility.md
kachawla, line 49: What is the difference between these two user scenarios?

architecture/2025-04-compute-platform-extensibility.md
```diff
@@ -0,0 +1,505 @@
...
+}
+```
+
+#### Output
+
+The above example will deploy on either Kubernetes or ACI.
+
+## Design
+
+### High Level Design
```
willtsai, line 123: Will there need to be any testing considerations to be called out? e.g. unlike Kubernetes, I'm not sure if ACI has "local" setups available, or will all ACI functional testing need to be done on the cloud?

architecture/2025-04-compute-platform-extensibility.md
```diff
@@ -0,0 +1,505 @@
...
+
+#### User story 5
+
+As a user, upon the successful deployment of my application to my ACI environment, I can view the ACI containers in the CLI using the rad app graph command or via the Application Graph in the Radius Dashboard.
+
+#### User story 6
+
+As a user, I may set ACI-specific configurations for my containers via the extensions or runtimes properties. This allows me to make use of ACI-specific features like confidential containers or spot instances that may not be available on other container compute platforms. Note that any properties that are not applicable to the compute platform in the targeted deployment environment will be ignored.
```
kachawla, line 65: Seems similar to user story 4, could you help me understand the difference?

architecture/2025-04-compute-platform-extensibility.md
```diff
@@ -0,0 +1,505 @@
...
+1. Container Definitions: The container definitions within the application definition files will allow users to deploy their containers to both Kubernetes and ACI without significant changes.
+1. Extension Properties, aka "punch-through": The Radius codebase will be enhanced to support extension properties in the application definition files. These properties will allow users to apply ACI-specific configurations to their containers, such as custom resource settings, networking configurations, and other ACI-specific parameters.
+1. Error Handling and Logging: The error handling and logging mechanisms within the Radius codebase will be updated to account for ACI-specific scenarios. This includes handling errors related to ACI resource creation, configuration, and deployment, as well as providing detailed logs for troubleshooting and diagnostics.
+1. Gateways: Radius will provision the necessary Azure resources to establish gateways for routing internet traffic to services hosted in ACI. This includes defining and creating gateways and secret stores for use in ACI containers.
```
sk593, line 132: We should discuss if we're going to make an opinionated choice of using Azure Application Gateways

architecture/2025-04-compute-platform-extensibility.md
```diff
@@ -0,0 +1,505 @@
...
+### Sample Input:Radius Environment*
+
+```bicep
+param aciscope string = '/subscriptions/<subscription id/resourceGroups/<resource group name>'
+resource env 'Applications.Core/environments@2023-10-01-preview' = {
+  name: 'aci-env'
+  properties: {
+    compute: {
+      kind: 'aci'
+      resourceGroup: aciscope
```
kachawla, line 85: Is this option for resource group available only if kind is set to `aci` or were you thinking of enforcing this on server side instead? I'm wondering if we should consider taking the approach of providers schema below in this bicep

```
compute: {
    aci: {
        resourceGroup:
     }
}
```

architecture/2025-04-compute-platform-extensibility.md
```diff
@@ -0,0 +1,505 @@
...
+
+The rendering system is a key part of the architecture with several important relationships:
+
+1. Multiplexed Rendering
+    * The `mux.Renderer` acts as a router that directs rendering requests to the appropriate platform-specific renderer
+    * It examines the `EnvironmentComputeKind` (Kubernetes or ACI) from the environment configuration and selects the correct renderer chain
+1. Layered Rendering
+    * Renderers are organized in a layered pattern using composition:
+   KubernetesCompute: kubernetesmetadata → manualscale → daprextension → container
+   ACICompute: aci_manualscale → aci
```
willtsai, line 169: is the goal to have renderers organized into identical layers for all compute types? Or do we expect each compute to be composed of different layers?

architecture/2025-04-compute-platform-extensibility.md
```diff
@@ -0,0 +1,505 @@
...
+
+1. Azure Container Instances (ACI) Integration
+    * `corerp/renderers/aci` implements ACI-specific resource rendering logic
+    * `corerp/renderers/aci/gateway` provides gateway functionality for ACI
+    * `corerp/renderers/aci/manualscale` handles scaling for ACI containers
+1. Renderer Multiplexer
+    * `corerp/renderers/mux` routes requests to appropriate renderers based on compute environment
+    * Selects between Kubernetes and ACI renderers based on environment configuration
+
+This architecture supports multi-cloud deployments with a unified programming model, where applications can be defined once and deployed to either Kubernetes or Azure Container Instances environments based on the environment configuration.
```
sk593, line 155: I think we're missing a handlers section here. We will likely have explicit handlers for different components of ACI that need to get deployed as output resources

architecture/2025-04-compute-platform-extensibility.md
```diff
@@ -0,0 +1,505 @@
...
+  properties: {
+    environment: environment
+    application: application
+  }
+}
+```
+
+#### Output
+
+The above example will deploy on either Kubernetes or ACI.
```
ytimocin, line 119: I guess this is depending on the environment. Because the environment example above is ACI specific.

architecture/2025-04-compute-platform-extensibility.md
lakshmimsft: Do we know specifics of what ACI-specific properties/configurations will be allowed to be configured? Will that list expand over time?

architecture/2025-04-compute-platform-extensibility.md
```diff
@@ -0,0 +1,505 @@
...
+1. Core Interface Hierarchy
+    * `renderers.Renderer` is the central interface that defines the rendering contract
+    * All concrete renderers implement this interface, providing a consistent API
+1. Multiplexer Implementation
+    * `mux.Renderer` implements the `renderers.Renderer` interface but delegates to inner renderers
+    * It maintains a map of inner renderers keyed by `EnvironmentComputeKind` (Kubernetes or ACI)
+    * During rendering, it examines the environment compute kind and delegates to the appropriate renderer.
+1. Concrete Renderer Implementations
+    * Platform-specific renderers like `aci.Renderer` and `container.Renderer` (Kubernetes) implement the same interface
+    * Resource-specific renderers like `gateway.Renderer` and `aci_gateway.Renderer` handle specific resource types
```
willtsai, line 183: I assume these need to be renamed as a part of this work? Or perhaps as a part of the "Radius core platform extensibility" work?

architecture/2025-04-compute-platform-extensibility.md
```diff
@@ -0,0 +1,505 @@
...
+        scope: aciscope
+      }
+    }
+  }
+}
+```
+
+### Sample Input: Portable Resources
+
+[Portable Resources](https://docs.radapp.io/guides/author-apps/portable-resources/overview/) are expected to work on ACI with no changes from the way they would be configured for Kubernetes:
```
kachawla, line 98: Will the connection from Radius Container to portable resource work the same way as well between Kubernetes and ACI?

architecture/2025-04-compute-platform-extensibility.md
```diff
@@ -0,0 +1,505 @@
...
+The above example will deploy on either Kubernetes or ACI.
+
+## Design
+
+### High Level Design
+
+The integration with ACI will be achieved by extending the existing Radius control plane, which will continue to be hosted in Kubernetes, to support ACI as a target deployment environment. The following summarizes the key changes and enhancements that will be made to the Radius codebase to support ACI deployments:
+
+1. Environment Definition and Setup: The Radius control plane will be updated to recognize and process environment definition files (env.bicep) that specify ACI as the target compute platform. This will involve adding support for ACI-specific configurations and parameters in the environment setup process.
+1. Application Definition and Deployment: The application definition files (app.bicep) will be extended to include ACI-specific properties and configurations. The Radius deployment engine will be enhanced to interpret these properties and deploy applications to ACI environments. This includes creating and managing ACI container groups, configuring networking, and handling ACI-specific resource settings.
```
kachawla, line 128: > The application definition files (app.bicep) will be extended to include ACI-specific properties and configurations.

Which part of app.bicep will be extended?

> The Radius deployment engine will be enhanced to interpret these properties and deploy applications to ACI environments

Could you please elaborate on why the deployment engine needs to be customized for this?

architecture/2025-04-compute-platform-extensibility.md
```diff
@@ -0,0 +1,505 @@
...
+A basic test scenario will be added to Radius that will be included in the cloud functional tests.
+
+## Security
+
+The Radius [Secret Store](https://docs.radapp.io/reference/resource-schema/core-schema/secretstore/) portable type will create an Azure KeyVault when deploying to ACI.
+
+## Compatibility
+
+* When a resource type is not supported by the compute kind, a descriptive error will be emitted.
+* When a "punch through" specification is added to a bicep file to provide platform specific configuration, configurations that do not apply to the target environment will be ignored and not emit an error.
```
willtsai, line 460: I suggest emitting an informational log every time something that doesn't apply is ignored. something like "Configuration <X> was ignored as it does not apply to targeted deployment environment <Y>."

architecture/2025-04-compute-platform-extensibility.md
```diff
@@ -0,0 +1,505 @@
...
+1. Application Definition and Deployment: The application definition files (app.bicep) will be extended to include ACI-specific properties and configurations. The Radius deployment engine will be enhanced to interpret these properties and deploy applications to ACI environments. This includes creating and managing ACI container groups, configuring networking, and handling ACI-specific resource settings.
+1. Container Definitions: The container definitions within the application definition files will allow users to deploy their containers to both Kubernetes and ACI without significant changes.
+1. Extension Properties, aka "punch-through": The Radius codebase will be enhanced to support extension properties in the application definition files. These properties will allow users to apply ACI-specific configurations to their containers, such as custom resource settings, networking configurations, and other ACI-specific parameters.
```
kachawla, line 130: Could you please clarify the difference between this and #2 above (`The application definition files (app.bicep) will be extended to include ACI-specific properties and configurations.`)?

architecture/2025-04-compute-platform-extensibility.md
```diff
@@ -0,0 +1,505 @@
...
+
+No changes to monitoring and logging.
+
+## Development plan
+
+### Milestone 1: Add the ACI feature to Radius
+
+Exit Criteria
+
+* The ACI integration is merged with the main branch in the Radius repo.
```
willtsai, line 472: And also the existing ACI integration (prototype, hardcoded) is able to handle "punch-through" configurations like Confidential Containers.

architecture/2025-04-compute-platform-extensibility.md
```diff
@@ -0,0 +1,505 @@
...
+1. Handle multiple compute environments through a single interface
+1. Dynamically select the appropriate renderer at runtime
+1. Share common interfaces across different platform implementations
+1. Easily extend the system with new compute environments by adding renderers to the map
+
+The design follows the Strategy Pattern, where the rendering strategy is selected based on the environment configuration, while maintaining a consistent interface for the rest of the system.
+
+### Options Considered
+
+#### Proposed Option: Two Milestones
```
sk593, line 202: I think this option makes most sense. The demo already has areas where extensibility would be fairly simple to implement vs the second option proposed

architecture/2025-04-compute-platform-extensibility.md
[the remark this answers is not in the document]
brooke-hamilton: This is an Azure Gateway

architecture/2025-04-compute-platform-extensibility.md
```diff
@@ -0,0 +1,505 @@
...
+
+Include separate high level architecture diagram and component specific diagrams, wherever appropriate.
+-->
+
+### Detailed Design
+
+#### Key Components and Relationships
+
+1. Azure Container Instances (ACI) Integration
+    * `corerp/renderers/aci` implements ACI-specific resource rendering logic
```
ytimocin, line 148: Would this folder structure require for the existing folder structure to change?

Like `corerp/renderers/kubernetes` and/or `corerp/renderers/aci`?

architecture/2025-04-compute-platform-extensibility.md
```diff
@@ -0,0 +1,505 @@
...
+1. Container Definitions: The container definitions within the application definition files will allow users to deploy their containers to both Kubernetes and ACI without significant changes.
+1. Extension Properties, aka "punch-through": The Radius codebase will be enhanced to support extension properties in the application definition files. These properties will allow users to apply ACI-specific configurations to their containers, such as custom resource settings, networking configurations, and other ACI-specific parameters.
+1. Error Handling and Logging: The error handling and logging mechanisms within the Radius codebase will be updated to account for ACI-specific scenarios. This includes handling errors related to ACI resource creation, configuration, and deployment, as well as providing detailed logs for troubleshooting and diagnostics.
```
kachawla, line 131: What type of updates do you envision in error handling? I'd assume Radius would just propagate back whatever ACI returned.

architecture/2025-04-compute-platform-extensibility.md
kachawla, line 132: Will this be explicitly defined by users or implicitly provisioned by Radius?

architecture/2025-04-compute-platform-extensibility.md
```diff
@@ -0,0 +1,505 @@
...
+1. Gateways: Radius will provision the necessary Azure resources to establish gateways for routing internet traffic to services hosted in ACI. This includes defining and creating gateways and secret stores for use in ACI containers.
+1. TBD: Secret Management: Radius will provision an Azure KeyVault resource on behalf of the user to store and manage secrets. These secrets can then be referenced in the application definition files to inject secret values as environment variables in ACI containers or to authenticate into services.
```
kachawla, line 133: Is this different from currently supported `Applications.Core/SecretStore` resource? If not, then this is really enabling another type of secret manager for the Radius secret store resource, rest of the functionality remains the same as it is today regardless of where the app gets deployed to.

architecture/2025-04-compute-platform-extensibility.md
```diff
@@ -0,0 +1,505 @@
...
+    * `corerp/renderers/aci` implements ACI-specific resource rendering logic
+    * `corerp/renderers/aci/gateway` provides gateway functionality for ACI
+    * `corerp/renderers/aci/manualscale` handles scaling for ACI containers
+1. Renderer Multiplexer
+    * `corerp/renderers/mux` routes requests to appropriate renderers based on compute environment
+    * Selects between Kubernetes and ACI renderers based on environment configuration
+
+This architecture supports multi-cloud deployments with a unified programming model, where applications can be defined once and deployed to either Kubernetes or Azure Container Instances environments based on the environment configuration.
+
+#### Understanding the Radius/ACI Architecture Relationships
```
kachawla, line 157: Any changes needed in the deployment processor?

architecture/2025-04-compute-platform-extensibility.md
```diff
@@ -0,0 +1,505 @@
...
+#### Option 2: Design then build
+
+This option is a traditional development cycle of design --> build.
+
+* Design an extensibility solution.
+* Implement the solution.
+
+### API design (if applicable)
+
+No changes from current design.
```
willtsai, line 227: Not sure if it's part of Milestone 1 or Milestone 2, but should there be considerations for how some of the environment and container schemas might change to accommodate compute extensibility? e.g. env compute properties for ACI currently is `kind: 'aci'` with `resourceGroup` required, while for k8s is `kind: 'kubernetes'` with `namespace` required. Would we want to evaluate whether their respective required properties can be collapsed into a since field that encompasses both `namespace` and `resourceGroup` whose value would just be dependent on the target environment?

architecture/2025-04-compute-platform-extensibility.md
```diff
@@ -0,0 +1,505 @@
...
+### Milestone 2: Serverless Extensibility
+
+Exit Criteria
+
+* An extensibility interface that allows compute platforms to be added.
+* Documentation on how to create a compute environment.
+
+## Open Questions
+
+* Can we use Dapr for secret stores on ACI, or should we hard code KeyVault?
```
sk593, line 485: We already have an implementation for Azure KeyVault where we can mount a keyvault as a volume on a container. I'm guessing there would be some overlap in scenarios here that might be useful for this

architecture/2025-04-compute-platform-extensibility.md
```diff
@@ -0,0 +1,505 @@
...
+
+No changes from current design.
+
+### CLI Design (if applicable)
+
+No changes from current design.
+
+### Implementation Details
+
+The implementation involves conditional logic based on the compute kind. (A longer-term solution provide extensibility instead of conditional logic hard coded for each platform.)
```
lakshmimsft, line 235: Maybe I'm missing this but can you point me to notes on implementation of punch-through mechanism?

architecture/2025-04-compute-platform-extensibility.md
```diff
@@ -0,0 +1,505 @@
...
+* We will have a customer-ready demo faster.
+* Higher-quality design informed by the demo work
+
+##### Disadvantages
+
+* Longer calendar schedule
+
+#### Option 2: Design then build
+
+This option is a traditional development cycle of design --> build.
```
kachawla, line 220: The detailed design above is what's in the demo code today? Could you share a high level overview of what would be different if we were to go with option 2?

architecture/2025-04-compute-platform-extensibility.md
ytimocin: I was also thinking why we can't do it this way:

- `corerp/renderers/gateway/aci`
- `corerp/renderers/gateway/kubernetes`

I mean like implementation for both lives under gateway (or any other resource) and the process is decided based on what user wants. In very simple definition with an if case (if runtime = k8s or aci).

architecture/2025-04-compute-platform-extensibility.md
```diff
@@ -0,0 +1,505 @@
...
+The rendering system is a key part of the architecture with several important relationships:
+
+1. Multiplexed Rendering
+    * The `mux.Renderer` acts as a router that directs rendering requests to the appropriate platform-specific renderer
+    * It examines the `EnvironmentComputeKind` (Kubernetes or ACI) from the environment configuration and selects the correct renderer chain
+1. Layered Rendering
+    * Renderers are organized in a layered pattern using composition:
+   KubernetesCompute: kubernetesmetadata → manualscale → daprextension → container
+   ACICompute: aci_manualscale → aci
+    * Each layer handles a specific aspect of rendering, with inner renderers focusing on core functionality and outer renderers adding specialized features
```
ytimocin, line 170: This is why instead of doing the routing in the router level (routing to aci or kubernetes implementations) we can do it in the renderer level by just checking what the environment type is.

architecture/2025-04-compute-platform-extensibility.md
```diff
@@ -0,0 +1,505 @@
...
+#### UCP (if applicable)
+#### Bicep (if applicable)
+#### Deployment Engine (if applicable)
+#### Core RP (if applicable)
+#### Portable Resources / Recipes RP (if applicable)
+-->
+
+### Error Handling
+
+The approach to error handling will stay the same, with two additional features:
```
kachawla, line 447: what are the two additional features? :)

architecture/2025-04-compute-platform-extensibility.md
```diff
@@ -0,0 +1,505 @@
...
+
+## Test plan
+
+A basic test scenario will be added to Radius that will be included in the cloud functional tests.
```
kachawla, line 451: Will it include extensions and supported configs?

architecture/2025-04-compute-platform-extensibility.md
sk593: Just a callout, but we also might need to reevaluate how we handle namespace mapping with Kubernetes. We might run into issues with a namespace not being needed with an ACI environment

architecture/2025-04-compute-platform-extensibility.md
ytimocin: After reading the rest, I think that is also what you decided to go with. Happy to discuss further if you'd like.

architecture/2025-04-compute-platform-extensibility.md
kachawla, line 485: Not sure why we need to hardcode, unerlying secret manager for Applications.Core/SecretStore is selected by the user.

Also what are the pros and cons of each?

architecture/2025-04-compute-platform-extensibility.md
```diff
@@ -0,0 +1,505 @@
...
+
+Exit Criteria
+
+* An extensibility interface that allows compute platforms to be added.
+* Documentation on how to create a compute environment.
+
+## Open Questions
+
+* Can we use Dapr for secret stores on ACI, or should we hard code KeyVault?
+* What are we missing in terms of features/design/coding that will have to be done for the release of ACI capabilities?
```
kachawla, line 486: In Radius or ACI?

architecture/2025-04-compute-platform-extensibility.md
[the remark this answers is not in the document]
sk593: We wouldn't create an ACI resource unless a container is defined. The environment-only deployment might set up backing resource needed for ACI to work (i.e. networking resources)

architecture/2025-04-compute-platform-extensibility.md
```diff
@@ -0,0 +1,505 @@
...
+
+### Next Design: Extensibility for Other Platforms
```
kachawla, line 488: This likely needs its own design doc :)

architecture/2025-04-compute-platform-extensibility.md
```diff
@@ -0,0 +1,505 @@
...
+
+## Test plan
+
+A basic test scenario will be added to Radius that will be included in the cloud functional tests.
+
+## Security
+
+The Radius [Secret Store](https://docs.radapp.io/reference/resource-schema/core-schema/secretstore/) portable type will create an Azure KeyVault when deploying to ACI.
+
+## Compatibility
```
sk593, line 457: We should add an entry for how we keep the SDK up to date. We can't use the public SDK so we'd have to generate the SDK based off the swagger specs and would need some way to keep the specs up to date+compatible

architecture/2025-04-compute-platform-extensibility.md
sk593: We should also think about places where we have some pretty heavy Kubernetes-centric coupling. An example would be the update filters that we have defined on the applications controller where we create app namespaces: https://github.com/radius-project/radius/blob/6b9df3b1c1ed6053531c12e9a96595a3ed86a0b5/pkg/corerp/setup/setup.go#L76

architecture/2025-04-compute-platform-extensibility.md
brooke-hamilton: Yes - Radius will run in Kubernetes. I added clarifying text.

architecture/2025-04-compute-platform-extensibility.md
brooke-hamilton: A managed offering would be a first-party Azure service based on Radius that provides an application model for ACI.

architecture/2025-04-compute-platform-extensibility.md
brooke-hamilton: Updated

architecture/2025-04-compute-platform-extensibility.md
brooke-hamilton: Yes ACI hosts containers only. I will remove the application phrase to make it more clear.

architecture/2025-04-compute-platform-extensibility.md
brooke-hamilton: Removing this objective - it is unclear.

architecture/2025-04-compute-platform-extensibility.md
brooke-hamilton: This has changed - we will use extenders for confidential compute configuration. I will remove it as a non-goal.

architecture/2025-04-compute-platform-extensibility.md
brooke-hamilton: Consolidating

brooke-hamilton: Updated

architecture/2025-04-compute-platform-extensibility.md
brooke-hamilton: It would be the same environment resource type used for ACI and Kubernetes.

architecture/2025-04-compute-platform-extensibility.md
brooke-hamilton: How about a list of which resource types are supported/not supported by this ACI implementation?

brooke-hamilton: Yes

architecture/2025-04-compute-platform-extensibility.md
brooke-hamilton: Yes we will create a list of supported resource types.

architecture/2025-04-compute-platform-extensibility.md
brooke-hamilton: We had a few conversations on this and I think where we landed was leaving `rad init` unchanged for this release, which means that a user would add an ACI environment after running `rad init`. @willtsai and @sk593 please correct me if I am wrong on this.

architecture/2025-04-compute-platform-extensibility.md
brooke-hamilton: We are making the choice to use Azure Application Gateways.

architecture/2025-04-compute-platform-extensibility.md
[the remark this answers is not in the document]
brooke-hamilton: We are not having Radius create a managed identity, but let's confirm with @willtsai.

architecture/2025-04-compute-platform-extensibility.md
brooke-hamilton: We don't plan to rename these right now, but I'm open to discuss doing so.

architecture/2025-04-compute-platform-extensibility.md
brooke-hamilton: Option 2 results in the same long-term design and implementation as option 1. The difference is that Option 2 would eliminate the early delivery of a production-ready implementation that is based on the current demo code.

architecture/2025-04-compute-platform-extensibility.md
brooke-hamilton: Adding a note here to confirm that working with @willtsai and @sk593 we will be working together on a small design to represent schema changes for milestone 1, which will include namespace mapping.

architecture/2025-04-compute-platform-extensibility.md
brooke-hamilton: User story 4 (which is now renamed to user story 3) is about the ability to use the platform-agnostic resource types to define a container to deploy to ACI, while this user story describes the ability to set ACI-specific properties.

architecture/2025-04-compute-platform-extensibility.md
brooke-hamilton: Testing will be done in the cloud.

architecture/2025-04-compute-platform-extensibility.md
brooke-hamilton: Thanks for pointing this out - at the time of writing I believed we would need to modify the deployment engine but that has proved to be unnecessary. I am removing that statement. 

Regarding the bicep schemas, the scope of changes has also decreased. We have added an additional design activity to clearly define the schema changes and that will be delivered in a separate document.

architecture/2025-04-compute-platform-extensibility.md
brooke-hamilton: Extensions will be used to support confidential containers and other ACI-specific configurations. We will produce a separate design to clarify how it will be implemented.

architecture/2025-04-compute-platform-extensibility.md
brooke-hamilton: Gateways will be defined by users using the Gateway core type, which will provision an Azure Application Gateway.

architecture/2025-04-compute-platform-extensibility.md
brooke-hamilton: Users will need clear error messages when they have used a Radius feature that is not supported by an ACI deployment. I will update the text here to clarify that.

architecture/2025-04-compute-platform-extensibility.md
brooke-hamilton: Yes, we would implement an Azure KeyVault deployment for the `Applications.Core/SecretStore`. Doing so is not in scope for the first release.

architecture/2025-04-compute-platform-extensibility.md
brooke-hamilton: The current demo implementation is like this: `corerp/renderers/aci/gateway`, but we will change the structure when we get to the second milestone.

architecture/2025-04-compute-platform-extensibility.md
brooke-hamilton: Yes - the deployment processor requires a resource group property to be set for ACI. I will update the doc.

architecture/2025-04-compute-platform-extensibility.md
brooke-hamilton: This is a good question and I think that when we get to the next phase of extensions we will make that decision.

architecture/2025-04-compute-platform-extensibility.md
brooke-hamilton: We will support punch through using the regular extension mechanism, except that ACI extensions will be mapped to ACI-specific properties. We are doing a small design focused on that area and will publish it for review.

architecture/2025-04-compute-platform-extensibility.md
brooke-hamilton: thank you - fixed it :)

architecture/2025-04-compute-platform-extensibility.md
brooke-hamilton: Yes, and I updated the language to specify that we will add cloud functional tests for all supported ACI features.

architecture/2025-04-compute-platform-extensibility.md
brooke-hamilton: I added a bullet point to this section to cover the SDK integration.

architecture/2025-04-compute-platform-extensibility.md
brooke-hamilton: I added that. Thank you.

architecture/2025-04-compute-platform-extensibility.md
brooke-hamilton: Added

architecture/2025-04-compute-platform-extensibility.md
brooke-hamilton: Here's more context on the question. The `Applications.Core/secretStores` resource type currently only supports Kubernetes secret stores. We received a question from a stakeholder on whether we should have  `Applications.Core/secretStores` deploy Dapr secret stores for ACI (which support KeyVault). But we also have `Applications.Dapr/secretStores`, which could perform the same function.

architecture/2025-04-compute-platform-extensibility.md
brooke-hamilton: I'm not sure - I think this question is too general and I'm going to remove it.

architecture/2025-04-compute-platform-extensibility.md
brooke-hamilton: Yes - this definitely needs its own design. It will be posted in a separate document.

architecture/2025-04-compute-platform-extensibility.md
brooke-hamilton: Good point and I'm adding this to the question.

architecture/2025-04-compute-platform-extensibility.md
brooke-hamilton: I added a task to the development plan for documentation of these things.

architecture/2025-04-compute-platform-extensibility.md
brooke-hamilton: I added a handlers section. Let's talk about the filters. I'm not sure about that one.

architecture/2025-04-aci-support.md
```diff
@@ -0,0 +1,567 @@
...
+
+### Sample Input: Radius Environment
+
+```diff
+param aciscope string = '/subscriptions/<subscription id/resourceGroups/<resource group name>'
+resource env 'Applications.Core/environments@2023-10-01-preview' = {
+  name: 'aci-env'
+  properties: {
+    compute: {  
+-        kind: 'kubernetes'
```
sk593, line 78: We ended up sticking with the original implementation of kind + resource group with the expectation that it'd get replaced with extensibility

sk593 (approved):
Some minor comments about the schema. Since we decided to move to compute extensibility with UDTs, we kept the POC schema.

sylvainsf (approved):
:shipit: looking forward to deprecating this soon! :)

sylvainsf (approved):
🚢

PR #89, opened 2025-03-26
GitHub workflow changes

brooke-hamilton:
This design is a set of principles we can use when editing our workflows.


nithyatsu:
This sounds so cool 👍

tools/2025-03-workflow-changes.md
```diff
@@ -0,0 +1,255 @@
...
+### Existing GitHub Workflows
+
+Existing workflows will evolve to implement the design principles. For example, if a new step is being added to an existing workflow, the step would be added via a Make command, which adheres to the principle that all logic is runnable on a developer machine.
+
+### New GitHub Workflows
+
+New GitHub workflows will adhere to the design principles.
+
+## Open Questions and Actions
+
```
willdavsmith, line 249: one open question I have is if anyone's tried this before. I'm happy to be the first to do it (sounds like a great blog post opportunity) but I wonder if we could find an example of a project that handles automation like this

tools/2025-03-workflow-changes.md
```diff
@@ -0,0 +1,255 @@
...
+:warning: When using GitHub actions that are published on other repositories, we are [placing our trust](https://arstechnica.com/information-technology/2025/03/supply-chain-attack-exposing-credentials-affects-23k-users-of-tj-actions/) in the authors of that repo that they will prevent malicious code from executing. Choose wisely, and consider forking and customizing a GitHub action instead of calling it directly.
+
+#### Reusable logic exists in custom GitHub actions instead of copy/paste to multiple workflows
+
+Apply the DRY principle (don't repeat yourself).
+
+#### Core logic is runnable from Make commands
+
+Any core logic that is currently in GitHub workflows will be moved to Make commands that are runnable on a developer machine (with some environment setup).
```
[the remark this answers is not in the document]
brooke-hamilton: Excellent point - I will add an example to the doc

tools/2025-03-workflow-changes.md
brooke-hamilton: Good point - I will look for some.

tools/2025-03-workflow-changes.md
```diff
@@ -0,0 +1,255 @@
...
+
+* This design does not include details on how each Radius workflow would be modified to align with the patterns defined here.
+* Not every workflow must be updated to the principles defined in this design. We will prioritize refactoring workflows based on the level of effort each workflow is requiring from the team when we need to make changes or when the workflows fail and need attention.
+* Replacement of security tokens, e.g. PATs, with GitHub applications.
+
+### User scenarios
+
+#### User story: Forks
+
+As a Radius developer, I can fork a repo and run any workflow on the fork with minimal setup, so that I can quickly test changes and confidently submit a PR for those changes.
```
lakshmimsft, line 53: This is soo needed.

tools/2025-03-workflow-changes.md
```diff
@@ -0,0 +1,255 @@
...
+Existing workflows will evolve to implement the design principles. For example, if a new step is being added to an existing workflow, the step would be added via a Make command, which adheres to the principle that all logic is runnable on a developer machine.
+
+### New GitHub Workflows
+
+New GitHub workflows will adhere to the design principles.
+
+## Open Questions and Actions
+
+* Which actions cannot be performed by the GitHub CLI? How will identity be provided to them?
+* Provide guidance on which git/GitHub CLI commands allow a PR reviewer to check out a PR to their own fork.
```
lakshmimsft, line 251: Thinking aloud here. I think we've done this for existing PRs. We can/do checkout other PR in repos and include it in our fork on occasion. The checking out of the repo includes checking out of the workflow file(s). The ability to run the workflow in a dev context is what is would be new.

tools/2025-03-workflow-changes.md
```diff
@@ -0,0 +1,255 @@
...
+* We need a prototype of moderate complexity that implements these principles.
+* Can we separate the scheduling from the workflows so that forks do not automatically run the workflows?
```
lakshmimsft, line 253: Would workflows be triggered only on demand in forks and ,say, be disabled for when user is not in active development?

tools/2025-03-workflow-changes.md
brooke-hamilton: I found some things on using .env files to set local variables, using [reusable workflows](https://docs.github.com/en/actions/sharing-automations/reusing-workflows). Also adding branch triggers and using [metadata in actions](https://docs.github.com/en/actions/sharing-automations/creating-actions/metadata-syntax-for-github-actions).

tools/2025-03-workflow-changes.md
brooke-hamilton: Yes - there are two elements to this. (1) every workflow should have a `workflow_dispatch:` trigger to allow manual runs, and (2) the chron configuration should exit the workflow if not running on the upstream repo. Something like this in the conditional expression: `github.repository_owner == 'upstream-owner'`.

willdavsmith (dismissed):

tools/2025-03-workflow-changes.md
```diff
@@ -0,0 +1,260 @@
...
+
+#### Use GitHub workflows for CI/CD setup and workflow layout
+
+We will use GitHub workflows for these things:
+
+* Identity and Security
+* Test runner setup
+* Control flow, including matrix operations
+
+GitHub workflows will not contain core logic (unless the workflow is very simple).
```
kachawla, line 119: Could you add an example of "core logic" and the rationale behind this principle?

tools/2025-03-workflow-changes.md
```diff
@@ -0,0 +1,260 @@
...
+
+### High Level Workflow Design Principles
+
+#### Use GitHub workflows for CI/CD setup and workflow layout
+
+We will use GitHub workflows for these things:
+
+* Identity and Security
+* Test runner setup
+* Control flow, including matrix operations
```
kachawla, line 117: Will be helpful to understand how did we narrow it down to this list and a high level description what each of them mean.

tools/2025-03-workflow-changes.md
```diff
@@ -0,0 +1,260 @@
...
+
+GitHub workflows will not contain core logic (unless the workflow is very simple).
+
+#### Core logic is debuggable and testable on a developer machine with minimal setup
+
+```mermaid
```
kachawla, line 123: - Where is this core logic expected to be implemented since the previous principle mentions that core logic is not a part of the workflow?

- What does "dev machine setup involve"?

tools/2025-03-workflow-changes.md
```diff
@@ -0,0 +1,260 @@
...
+#### Use GitHub actions for setup, cleanup, and matrix operations
+
+The use of actions should be limited to setup and cleanup that is unique to GitHub workflows or runners because any logic that exists in a GitHub action cannot be executed on a developer machine. Some examples are below:
+
+* Cloning repos
+* Installing tools
+* Retrieving stored secrets
+* Publishing results
+
+> NOTE: The above examples are also needed on a developer machine, but the Implementation is different.
```
kachawla, line 176: > but the Implementation is different

Can you please elaborate on this?

tools/2025-03-workflow-changes.md
```diff
@@ -0,0 +1,260 @@
...
+
+:warning: When using GitHub actions that are published on other repositories, we are [placing our trust](https://arstechnica.com/information-technology/2025/03/supply-chain-attack-exposing-credentials-affects-23k-users-of-tj-actions/) in the authors of that repo that they will prevent malicious code from executing. Choose wisely, and consider forking and customizing a GitHub action instead of calling it directly.
```
kachawla, line 178: > Choose wisely, and consider forking and customizing a GitHub action instead of calling it directly.

Could we add more specific guidance here?

tools/2025-03-workflow-changes.md
```diff
@@ -0,0 +1,260 @@
...
+
+#### Reusable logic exists in custom GitHub actions instead of copy/paste to multiple workflows
+
+Apply the DRY principle (don't repeat yourself).
```
kachawla, line 182: Are you proposing creating new GitHub actions for things repeated across multiple workflows?

tools/2025-03-workflow-changes.md
```diff
@@ -0,0 +1,260 @@
...
+
+#### Core logic is runnable from Make commands
```
kachawla, line 184: Description of what "core logic" means in this context will help.

tools/2025-03-workflow-changes.md
```diff
@@ -0,0 +1,260 @@
...
+
+Any core logic that is currently in GitHub workflows will be moved to Make commands that are runnable on a developer machine (with some environment setup).
+
+Developer automation is already provided through Make. Some of the GitHub workflows also use the same Make commands, e.g. `make build` is invoked from GitHub workflows.
+
+Some of the existing Make commands invoke scripts. This pattern will continue in order to keep all significant logic isolated to scripts because we do not want to move the complexity problem from GitHub workflows into Make. Simple composite commands should remain within the Make files as they are today.
```
kachawla, line 190: Any downsides of this approach that we need to consider?

tools/2025-03-workflow-changes.md
```diff
@@ -0,0 +1,260 @@
...
+
+#### Configuration is provided through environment variables
+
+Environment variables can be set on developer machines (and stored in `.env` files). GitHub workflows will set environment variables during setup steps. Core logic will read the environment variables.
```
kachawla, line 194: Just confirming - Is this different from what Radius workflows do today?

tools/2025-03-workflow-changes.md
```diff
@@ -0,0 +1,260 @@
...
+
+## Security
+
+### Least Privilege
+
+Workflows will continue to adhere to the least privilege feature in which they declare which permissions are required by the workflow.
+
+### Identity
+
+Where possible, the automation logic will use the GitHub CLI, which has identity built in. Developers would log into the CLI when running code on their machines, which would allow them to run the logic on their forks. When running as part of at GitHub workflow, the GitHub CLI context is already set up and no further configuration is required.
```
kachawla, line 223: Same question as above, is this different from what Radius workflows do today?

tools/2025-03-workflow-changes.md
```diff
@@ -0,0 +1,260 @@
...
+A new folder exists in the `radius/docs` folder that provides a `README.md` file with guidance on creating and editing GitHub workflows and other repo automation like `make`.
+
+### PR Template
+
+Modify the PR template checklist for each repository that contains workflows to include items similar to the following:
+
+Workflow modifications have been tested on a repo fork and the changes conform to the workflow development guidance (add hyperlink to the guidance )  
+
+* [ ] This PR contains no modifications to GitHub workflows or repo automation.  
+* [ ] Changes to GitHub workflows, actions, and other repo automation have been tested on a repo fork.
```
kachawla, line 238: Enforcing this is only possible if there are clear instructions on how to run each of the existing workflows on forks. Currenlty it is difficulte to replicate the secrets/setup used for a few workflows.

tools/2025-03-workflow-changes.md
```diff
@@ -0,0 +1,260 @@
...
+
+### Existing GitHub Workflows
+
+Existing workflows will evolve to implement the design principles. For example, if a new step is being added to an existing workflow, the step would be added via a Make command, which adheres to the principle that all logic is runnable on a developer machine.
+
+### New GitHub Workflows
+
+New GitHub workflows will adhere to the design principles.
+
+## Open Questions and Actions
```
kachawla, line 248: I missed the design meeting for this doc - Can you add answers to this questions (assuming they were answered in the design discussion) before checking this in?

tools/2025-03-workflow-changes.md
```diff
@@ -0,0 +1,260 @@
...
+
+* Which actions cannot be performed by the GitHub CLI? How will identity be provided to them?
+* Provide guidance on which git/GitHub CLI commands allow a PR reviewer to check out a PR to their own fork.
+* We need a prototype of moderate complexity that implements these principles.
+* Can we separate the scheduling from the workflows so that forks do not automatically run the workflows?
+
+## Design Review Notes
+
+* Add an example of moving logic into a Make command and show the before and after.
```
kachawla, line 257: Where will this example be added?

tools/2025-03-workflow-changes.md
```diff
@@ -0,0 +1,260 @@
...
+* Consider [reusable workflows](https://docs.github.com/en/actions/sharing-automations/reusing-workflows).
```
kachawla, line 258: Would be good to expand on this - why/where etc.

tools/2025-03-workflow-changes.md
```diff
@@ -0,0 +1,260 @@
...
+* Create a checklist for reviewing workflows.
```
kachawla, line 259: Should this be in the development plan above?

tools/2025-03-workflow-changes.md
```diff
@@ -0,0 +1,260 @@
...
+
+### Least Privilege
+
+Workflows will continue to adhere to the least privilege feature in which they declare which permissions are required by the workflow.
+
+### Identity
+
+Where possible, the automation logic will use the GitHub CLI, which has identity built in. Developers would log into the CLI when running code on their machines, which would allow them to run the logic on their forks. When running as part of at GitHub workflow, the GitHub CLI context is already set up and no further configuration is required.
+
+## Development plan
```
kachawla, line 225: When will this be executed on?

tools/2025-03-workflow-changes.md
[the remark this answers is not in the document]
kachawla: @brooke-hamilton could you point me to example you have added?

tools/2025-03-workflow-changes.md
brooke-hamilton: We have a workflow edit to make in the current sprint, assigned to me. I will apply the design principles where possible, and create issues for follow-up work.

For the rest of the development plan, we don't have this scheduled.

kachawla: sounds great, thank you!

tools/2025-03-workflow-changes.md
brooke-hamilton: I believe there are some instances where git commands are used where a GH CLI command would be simpler, but this is an overall principle rather than a change to a specific instance in the current workflows.

tools/2025-03-workflow-changes.md
kachawla: Never mind, I see the definition for `core logic` in terms and definitions and some details on why later in this doc. I think it will help to add brief context here in this section as well.

tools/2025-03-workflow-changes.md
brooke-hamilton: In the workflows this is not different from what we do today. The new part is the ability to provide an `.env` file on the developer machine, which enables logic to run locally. That is to support local machine testing of workflow logic.

tools/2025-03-workflow-changes.md
brooke-hamilton: Editing Make files isn't what I think of as fun because they can be complex, but putting anything that is moderatly complex into a shell script that is invoked from Make is helpful to isolate the complexity.

The advantage of using Make and shell scripts is that it leverages a pattern we already have in place.

tools/2025-03-workflow-changes.md
brooke-hamilton: I agree, and we should be able to simplify setup of testing on forks by providing some automation to developers, like a shell script that configures a fork with the right environment variables. We also need to ensure that repo names are configurable, e.g., `radius-project/radius` should not be hard coded in a repo in a way that prevents testing in a fork.

The whole thing should be simple for devs, like 
1. Fork the repo
2. Apply repo configuration to the fork (probably just environment variables and secrets)
3. Apply local machine configuration (also environment variables)
4. Test the workflow on the fork
5. Test the logic on the local machine
6. Change the code
7. Test on the local machine
8. Push changes to the fork and test the fork
9. Create PR

tools/2025-03-workflow-changes.md
brooke-hamilton: I added a sample checklist to the development plan.

tools/2025-03-workflow-changes.md
brooke-hamilton: [Reusable workflows](https://docs.github.com/en/actions/sharing-automations/reusing-workflows) and [composite actions](https://docs.github.com/en/actions/sharing-automations/creating-actions/creating-a-composite-action) are helpful in [reducing duplication](https://docs.github.com/en/actions/sharing-automations/avoiding-duplication). The GitHub documentation does a good job explaining, and I added these links to the document.

tools/2025-03-workflow-changes.md
brooke-hamilton: I added an example, but how about adding a real-world example into the documentation? The workflow updates planned in the current sprint will almost certainly provide several good examples.

tools/2025-03-workflow-changes.md
brooke-hamilton: I updated the list to be a table with actions taken.

tools/2025-03-workflow-changes.md
brooke-hamilton: Yes, and also possibly [reusable workflows](https://docs.github.com/en/actions/sharing-automations/reusing-workflows) and [composite actions](https://docs.github.com/en/actions/sharing-automations/creating-actions/creating-a-composite-action).

tools/2025-03-workflow-changes.md
brooke-hamilton: I added a statement about how to think about which actions to trust.

tools/2025-03-workflow-changes.md
brooke-hamilton: I reworded the note to be more clear. If this is causing confusion I can remove it. It is kind of a non-sequitur, but I am attempting to point out that there are some things that can only be done in a GitHub workflow because their implementation is different there. But devs have the same need to set things up, and do that in their own way. Both workflows and devs need efficient and simple automation to accomplish the work.

tools/2025-03-workflow-changes.md
brooke-hamilton: The core logic would be in make commands that call shell scripts instead of embedded into workflows as a large set of shell commands that cannot be tested on dev machines or on forks. As make commands and shell scripts, the logic can be tested locally and as a GitHub workflow on a repo fork.

I clarified the definition of "core logic" in the document.

Dev machine setup is typically the combination of available tools and environment variables that are required to run the code that would be in a workflow.

tools/2025-03-workflow-changes.md
brooke-hamilton: These are the things that can only be done in GitHub workflows. They don't have meaning on a developer machine.

tools/2025-03-workflow-changes.md
brooke-hamilton: I expanded the definition of core logic to clarify.

tools/2025-03-workflow-changes.md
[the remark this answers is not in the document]
brooke-hamilton: I added an example.

tools/2025-03-workflow-changes.md
```diff
@@ -0,0 +1,301 @@
...
+
+* Significantly reduce the time required to create/modify/fix GitHub workflows.
+* Increase the reliability of workflows.
+* Reduce time dedicated to addressing workflow failures by the on-call engineer.
+
+### Non-goals
+
+* This design does not include details on how each Radius workflow would be modified to align with the patterns defined here.
+* Not every workflow must be updated to the principles defined in this design. We will prioritize refactoring workflows based on the level of effort each workflow is requiring from the team when we need to make changes or when the workflows fail and need attention.
+* Replacement of security tokens, e.g. PATs, with GitHub applications.
```
sylvainsf, line 47: If we're defining our best practices going forward this should be included even if on an individual workflow basis we may decide not to do this.

kachawla (approved):

kachawla (commented):
Thanks for addressing the comments. Merging it.

tools/2025-03-workflow-changes.md
sylvainsf: I think we could add a line here about how providing an unpopulated env file makes a clear unambiguous example of what to configure and everything will work.

sylvainsf (approved):
Looks great, really a direction which will make everything smoother.

PR #91, opened 2025-04-09
Extensibility of the compute platform

brooke-hamilton:
This document was originally part of #88. It is now in its own PR.


architecture/2025-04-compute-extensibility.md
```diff
@@ -0,0 +1,469 @@
...
+
+> NOTE: The resource types shown here are new versions, e.g., `environments@2025-05-01-preview`, not changes to the existing versions of the resource types. This allows us to add recipe support to the core types without breaking existing deployments.
+
+The `compute` platform is removed so that we do not have hard-coded support for specific platforms. The functionality enabled by the `compute.identity` property would be implemented via recipes so that we do not need hard-coded support for specific platforms. We could consider adding a recipe to the `environment` type if there are features enabled by `compute` that we could not achieve using recipes on the other core types.
+
+#### Initialize a workspace
+
+As a platform engineer I can use `rad init` (or the equivalent rad `workspace`, `group` and `environment` commands), plus `rad recipe register`, to set up a non-Kubernetes compute platform.
+
+`rad init` would be unchanged. Platform engineers would have to add recipes for each core type and UDT they plan to use.
```
sk593, line 97: We'd need to edit the default environment set up during `rad init`. Right now, this defaults to a Kubernetes compute environment

architecture/2025-04-compute-extensibility.md
```diff
@@ -0,0 +1,469 @@
...
+-       namespace: 'myapp'
+-     }
+-     {
+-       kind: 'kubernetesMetadata'
+-       labels: {
+-         'team.contact.name': 'frontend'
+-       }
+-     }
+-   ]
++   recipe: {
```
sk593, line 133: Why would an application need this information?

architecture/2025-04-compute-extensibility.md
```diff
@@ -0,0 +1,469 @@
...
++      // Name a specific Recipe to use
++      name: 'azure-aci-environment'
++      // Set parameters that will be passed to the recipe.
++      parameters: {
++        subscriptionId: 'xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx'
++        resourceGroup: 'myrg'
++      }
++    }
+  }
```
willtsai, line 142: I think you would specify a recipe here that is used to provision the application, correct? i.e. package up an entire application into a recipe, though I'm not sure what kind of use case would require this.

architecture/2025-04-compute-extensibility.md
```diff
@@ -0,0 +1,469 @@
...
+Describe how you will deliver your features. This includes aligning work items
+to features, scenarios, or requirements, defining what deliverable will be
+checked in at each point in the product and estimating the cost of each work
+item. Don’t forget to include the Unit Test and functional test in your
+estimates.
+-->
+
+## Open Questions
+
+* Should we convert Kubernetes deployments to recipes with this work, after this work, or not at all?
```
sk593, line 444: I'm not sure I understand this question. You can already define Kubernetes deployments using recipes

architecture/2025-04-compute-extensibility.md
zachcasper, line 133: This breaks the separate of developer and platform engineer. Why can the platform engineer not just register a recipe for Applications.Core/containers in an environment?

architecture/2025-04-compute-extensibility.md
```diff
@@ -0,0 +1,469 @@
...
+
+* **Author**: Brooke Hamilton (@Brooke-Hamilton)
+
+## Overview
+
+Radius will enable users to deploy and manage their applications to multiple compute platforms, in addition to Kubernetes, by adding support for recipes to the core resource types. Recipes will enable the Radius team to provide default implementations for multiple platforms, and allow customers to modify or provide their own deployment recipes.
+
+Modifications to Radius include:
+
+* The [core resource types](https://docs.radapp.io/reference/resource-schema/core-schema/) of application, container, gateway, and secret store will support recipes.
```
sk593, line 11: Why applications? I see the use cases for the others

architecture/2025-04-compute-extensibility.md
```diff
@@ -0,0 +1,469 @@
...
+-           {
+-             name: 'log-collector'
+-             image: 'ghcr.io/radius-project/fluent-bit:2.1.8'
+-           }
+-         ]
+-         hostNetwork: true
+-       }
+-     }
+-   }
++   recipe: {
```
zachcasper, line 198: What functionality would the platform engineer lose if they chose to use a recipe rather than the built-in deployment logic for a container? What about:

* Connections
* Environment variable injection

What else?

architecture/2025-04-compute-extensibility.md
```diff
@@ -0,0 +1,469 @@
...
+checked in at each point in the product and estimating the cost of each work
+item. Don’t forget to include the Unit Test and functional test in your
+estimates.
+-->
+
+## Open Questions
+
+* Should we convert Kubernetes deployments to recipes with this work, after this work, or not at all?
+* Is the Radius group concept affected by this design?
+* Do we need to extend [`environment.recipeConfig`](https://docs.radapp.io/reference/resource-schema/core-schema/environment-schema/#recipeconfig) to include Bicep configuration? (There is already a Terraform configuration.)
```
vishwahiremat, line 446: we already have bicep config properties. but looks like we have not added documentation for that.
https://github.com/radius-project/radius/blob/f89f424a2db9bd4cb0aaff5efa163b0c281cf670/pkg/corerp/datamodel/recipe_types.go#L25

willtsai (commented):
Good ideas and design principles. I think it could use some clarifications on what exactly is being proposed and respective problems to solve. I see it as two problems:
1. There needs to be a way to extend Radius core so that adding another supported compute platform outside of Kubernetes is possible and easy according to the design principles listed.
2. There is a need provision Radius core resources (namely Environments, Applications, Containers, SecretStores, Gateways) using Recipes.

We should explore what the experience for `rad deploy env.bicep` and `rad deploy app.bicep` would look like.

architecture/2025-04-compute-extensibility.md
```diff
@@ -0,0 +1,469 @@
+# Compute Platform Extensibility
+
+* **Author**: Brooke Hamilton (@Brooke-Hamilton)
+
+## Overview
+
+Radius will enable users to deploy and manage their applications to multiple compute platforms, in addition to Kubernetes, by adding support for recipes to the core resource types. Recipes will enable the Radius team to provide default implementations for multiple platforms, and allow customers to modify or provide their own deployment recipes.
```
zachcasper, line 7: Could we not implement all core types with recipes? We could ship a set of recipes for like we do today for Kubernetes (today's local-dev), Azure-Kubernetes, Azure-ACI, AWS-Kubernetes, and AWS-ECS. Then users could modify these recipes as needed.

architecture/2025-04-compute-extensibility.md
brooke-hamilton: I will remove recipes from applications.

architecture/2025-04-compute-extensibility.md
brooke-hamilton: I think we could. Let's try it.

architecture/2025-04-compute-extensibility.md
brooke-hamilton: That makes sense. I am removing recipe from applications.

brooke-hamilton: That makes sense. I am removing recipes from the applications type.

architecture/2025-04-compute-extensibility.md
brooke-hamilton: Based on comments from others I'm removing recipe from the applications core type, but we could easily add it back if we find a use for it.

architecture/2025-04-compute-extensibility.md
brooke-hamilton: The `containers` type allows setting environment variables and connections. But you are right, if we do a recipe deployment it would be up to the recipe to inject these into the containers at runtime. We could provide connections and env vars as a reusable module for Bicep and Terraform so that recipe authors don't have to re-implement this for Kubernetes. However, other platform recipes may have to provide their own implementation.

architecture/2025-04-compute-extensibility.md
brooke-hamilton: Yes this is vaguely worded. I'll add context to the question.

architecture/2025-04-compute-extensibility.md
brooke-hamilton: Thank you!

architecture/2025-04-compute-extensibility.md
```diff
@@ -0,0 +1,469 @@
...
+-        }
+-      }
+-    ]
+  }
+}
+```
+
+> NOTE: The resource types shown here are new versions, e.g., `environments@2025-05-01-preview`, not changes to the existing versions of the resource types. This allows us to add recipe support to the core types without breaking existing deployments.
+
+The `compute` platform is removed so that we do not have hard-coded support for specific platforms. The functionality enabled by the `compute.identity` property would be implemented via recipes so that we do not need hard-coded support for specific platforms. We could consider adding a recipe to the `environment` type if there are features enabled by `compute` that we could not achieve using recipes on the other core types.
```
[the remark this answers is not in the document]
brooke-hamilton: Yes - the recipes set for a type in a given environment would deploy to a given compute platform.

architecture/2025-04-compute-extensibility.md
brooke-hamilton: I am thinking of `rad init` as a default fast start setup, so defaulting to Kubernetes is good. Someone can test their deployment on a local cluster using `rad init`. They can later add new environments with new recipes for additional platforms.

architecture/2025-04-compute-extensibility.md
```diff
@@ -0,0 +1,474 @@
...
+|------|------------|
+| **Compute platform** | An environment where applications can be deployed and run, such as Kubernetes, Azure Container Instances (ACI), etc. |
+| **Core types** | Built-in resource types provided by Radius, including `containers`, `gateways`, `secretStores`, `environments`, and `applications`. |
+| **User-defined type (UDT)** | A custom resource type defined separately from Radius core types and loaded into Radius without requiring a new Radius release. |
+| **Recipe** | A set of instructions that provisions a Radius resource to a Radius environment. Recipes are implemented in Bicep or Terraform. |
+| **Resource Provider (RP)** | A component responsible for handling create, read, update, delete, and list (CRUDL) operations for a specific resource type. |
+
+### Goals
+
+* Provide platform engineers with the ability to deploy to specific platforms other than Radius, like ACI.
```
ytimocin, line 38: I think this is a very good idea.

architecture/2025-04-compute-extensibility.md
```diff
@@ -0,0 +1,474 @@
...
+* Provide a recipe-based platform engineering experience that is consistent for user-defined types and core types.
```
ytimocin, line 39: We need a step to validate the recipe that is being registered to not have runtime errors.

architecture/2025-04-compute-extensibility.md
```diff
@@ -0,0 +1,474 @@
...
+|-----------|-------------|
+| **Independently upgradeable in a runtime environment** | Recipes are independently upgradeable. |
+| **Isolated and over a network protocol** | Recipes are currently implemented through Bicep and Terraform. Both run locally and communicate to the target platform over secure network protocols. |
+| **Strongly typed and support versioning** | Recipes support versioning through the use of OCI-compliant registries. They are not strongly typed in the sense of having compile-time validation, but they can be validated upon registration in an environment as having the correct input parameters and output properties. |
+
+### User scenarios
+
+#### Configure a non-Kubernetes Radius environment
+
+As a platform engineer I can initialize a new Radius environment that is connected to a non-Kubernetes compute platform so that Radius can authenticate and deploy to non-Kubernetes platforms.
```
ytimocin, line 61: We should think about how existing users can migrate to this new version of Radius. We may need to introduce a `rad migrate` command.

architecture/2025-04-compute-extensibility.md
```diff
@@ -0,0 +1,474 @@
...
+}
+```
+
+#### Initialize a workspace
+
+As a platform engineer I can use `rad init` (or the equivalent rad `workspace`, `group` and `environment` commands), plus `rad recipe register`, to set up a non-Kubernetes compute platform.
+
+`rad init` would be unchanged (in terms of the user experience). Platform engineers would have to add recipes for each core type and UDT they plan to use. By default, the command will register resource types and recipes for Kubernetes provisioning.
+
+We could consider adding a flag to `rad init` that would identify a specific set of recipes. For example, `rad init aci` would install a set of default recipes for the ACI platform. However, this work is not in scope for this design.
```
ytimocin, line 137: If we are going to have `rad init aci` then the current `rad init` should be `rad init kubernetes`. Or, `rad init` can ask for a platform.

architecture/2025-04-compute-extensibility.md
```diff
@@ -0,0 +1,474 @@
...
+
+#### Extend Radius to a new platform by creating new recipes
+
+The `rad recipe register` CLI command would be unchanged, as it already has the ability to associate a recipe to an environment and a resource type, and it supports setting default values for specified recipe parameters for that environment.
```
ytimocin, line 141: We may need to have to validate the recipes registered not to have errors during runtime.

architecture/2025-04-compute-extensibility.md
```diff
@@ -0,0 +1,474 @@
...
+* **Earlier delivery**: ACI recipe provisioning can be added and released before Kubernetes recipe is released.
+
+### Disadvantages vs Recommended Plan
+
+* **Less architectural flexibility**: Core types remain hard-coded in Radius, limiting some extensibility options, e.g., the ability of a customer to copy and modify a core type.
+* **Inconsistent resource type model**: Core types and UDTs would have different implementation approaches. However, customers can choose to ignore the core types and implement their own.
+
+## Open Questions
+
+* What is the impact on the Radius graph, and how would we continue to support it? We will could implement the graph as the relationships defined in Bicep, or implement connections in a Bicep/Terraform reusable module.
```
ytimocin, line 452: Radius graph can be something that could be drawn from the index in the data store.

architecture/2025-04-compute-extensibility.md
```diff
@@ -0,0 +1,474 @@
...
+* Is the Radius group concept affected by this design?
+* Can everything currently deployed by Go code be deployed using recipes? We think so, but need to prove it through prototyping.
```
ytimocin, line 454: How will this (moving the logic from Go code into Bicep/Terraform) affect the performance? Should we think about this too?

architecture/2025-04-compute-extensibility.md
```diff
@@ -0,0 +1,474 @@
...
+* **Stable application model**: Maintains the familiar Radius application model that customers are already using.
+* **Incremental change**: Adding recipe support to core types is an additive change, and adding ACI provisioning with recipes is not disruptive to the existing Kubernetes provisioning code.
+* **Earlier delivery**: ACI recipe provisioning can be added and released before Kubernetes recipe is released.
+
+### Disadvantages vs Recommended Plan
+
+* **Less architectural flexibility**: Core types remain hard-coded in Radius, limiting some extensibility options, e.g., the ability of a customer to copy and modify a core type.
+* **Inconsistent resource type model**: Core types and UDTs would have different implementation approaches. However, customers can choose to ignore the core types and implement their own.
+
+## Open Questions
```
ytimocin, line 450: Can UDT system do everything currently done in Go?

ytimocin, line 450: Can recipes depend on each other? Can we find ourselves in an import cycle?

ytimocin, line 450: How about metrics in different platforms?

architecture/2025-04-compute-extensibility.md
kachawla, line 61: > a new Radius environment that is connected to a non-Kubernetes compute platform

creating a Radius environment doesn't spin up the compute that the applications will be deployed to, so I think "connected to" could be misleading

architecture/2025-04-compute-extensibility.md
```diff
@@ -0,0 +1,474 @@
...
+-        }
+-      }
+-    ]
+  }
+}
+```
+
+> NOTE: The resource types shown here are new versions, e.g., `environments@2025-05-01-preview`, not changes to the existing versions of the resource types. This allows us to avoid breaking existing deployments.
+
+The `compute` platform is removed so that we do not have hard-coded support for specific platforms. The functionality enabled by the `compute.identity` property would be implemented via recipes so that we do not need hard-coded support for specific platforms. We could consider adding a recipe to the `environment` type if there are features enabled by `compute` that we could not achieve using recipes on the other core types.
```
kachawla, line 104: What functionality is enabled by `compute.identity`? Does it exist today?

architecture/2025-04-compute-extensibility.md
brooke-hamilton: I will fix and I think we need more detail on how an environment is or is not affected by converting provisioning to recipes.

architecture/2025-04-compute-extensibility.md
brooke-hamilton: I will add more detail on what the `compute` section does.

brooke-hamilton: I will add more detail on what the `compute` section does.

architecture/2025-04-compute-extensibility.md
```diff
@@ -0,0 +1,474 @@
...
+
+## Principles
+
+Radius extensibility should enable creating extensions that are:
+
+| Principle | Description |
+|-----------|-------------|
+| **Independently upgradeable in a runtime environment** | Recipes are independently upgradeable. |
+| **Isolated and over a network protocol** | Recipes are currently implemented through Bicep and Terraform. Both run locally and communicate to the target platform over secure network protocols. |
+| **Strongly typed and support versioning** | Recipes support versioning through the use of OCI-compliant registries. They are not strongly typed in the sense of having compile-time validation, but they can be validated upon registration in an environment as having the correct input parameters and output properties. |
```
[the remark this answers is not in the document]
brooke-hamilton: This will be in the detailed design.

architecture/2025-04-compute-extensibility.md
brooke-hamilton: Metrics, like the time it takes to perform a deployment?

brooke-hamilton: I think we would have UDTs that depend on each other, but let's get more detail on whether recipes should depend on each other.

brooke-hamilton: For provisioning, I think the answer is yes. However, we still have to support connections and the graph, and I think that is the next thing to get detail on.

architecture/2025-04-compute-extensibility.md
brooke-hamilton: I don't think there will be a large effect because most of the time spent during provisioning is waiting for resources to be created on the target platform. But we should measure it.

architecture/2025-04-compute-extensibility.md
brooke-hamilton: 😎 idea. I think we may need to get it from the UDTs themselves - before the recipe runs.

architecture/2025-04-compute-extensibility.md
brooke-hamilton: If we put expectations on the recipes, like that they accept certain parameters (in addition to what is required for UDTs) we will have to validate.

architecture/2025-04-compute-extensibility.md
brooke-hamilton: OK but can we still have `rad init` default to Kubernetes? 😺

architecture/2025-04-compute-extensibility.md
brooke-hamilton: Good point. Hopefully migrating is simple because the UDT versions of the core resource types would be similar to the existing core types.

architecture/2025-04-compute-extensibility.md
brooke-hamilton: We are tracking recipe validation in the development plan.

architecture/2025-04-compute-extensibility.md
brooke-hamilton: A separate feature spec and detailed design is planned that will cover the user experience for extensibility.

sylvainsf (approved):
Looks great! :shipit:

PR #93, opened 2025-05-20
UDT connections support

nithyatsu:
This document captures the requirements for a connection involving a UDT resource

### DIFF
--- a/architecture/2025-05-user-defined-types-connections.md
+++ b/architecture/2025-05-user-defined-types-connections.md
@@ -1,0 +1,185 @@
+# User-Defined Types
+
+* **Author**: `Nithya Subramanian` (`@nithyatsu`)
+
+## Overview
+
+A Radius application typically defines resources of different types and connects them using [connections](https://docs.radapp.io/guides/author-apps/containers/howto-connect-dependencies/). These connections mainly serve two purposes:
+
+1.	Build application graph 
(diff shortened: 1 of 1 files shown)


architecture/2025-05-user-defined-types-connections.md
```diff
@@ -0,0 +1,185 @@
...
+}
+```
+
+## Design
+
+### Applications.Core/containers to UDT
+
+When there is a `connections` construct in a `Applications.Core/containers` resource definition with source set to a UDT resource's id, 
+1. The UDT should be visible as a dependency in the application graph.
+2. All of the properties of the UDT that are of numeric or string type should be set as environment variable in the container. The environment variable key would be `CONNECTION_CONN_NAME_VARIABLE_NAME` and value would be the value of varoable.
```
lakshmimsft, line 142: question for @kachawla , @zachcasper 
(correct me if I'm mistaken) Currently for existing connections in Containers, we create environment variables for OutputVariables, ComputedValues when their types are string/float/int.
We could continue with the same approach for Container->UDT connections. 
(I'm bringing this up since I saw an earlier version of Connections PR follow the same approach as existing and maybe there will need to be more work done to get ALL properties in schema to have env variables). So, I'm checking if that is a requirement.

nithyatsu: based on offine discussion and the use case for external service, we are deciding to propogate all properties than just recipe output.

PR #96, opened 2025-06-04
Feature spec: compute extensibility user experience

willtsai:
Add feature spec for compute extensibility to define requirements and user experience


architecture/2025-06-compute-extensibility-feature-spec.md
```diff
@@ -0,0 +1,469 @@
...
+
+## Key dependencies and risks
+<!-- What dependencies must we take in order to enable this scenario? -->
+<!-- What other risks are you aware of that need to be mitigated. If you have a mitigation in mind, summarize here. -->
+<!-- **Dependency Name** – summary of dependency.  Issues/concerns/risks with this dependency -->
+<!-- **Risk Name** – summary of risk.  Mitigation plan if known. If it is not yet known, no problem. -->
+- **Bicep/Terraform Capabilities and Limitations**: The feasibility of implementing all necessary provisioning logic (currently in Go for Kubernetes) within Bicep or Terraform recipes.
+    - Risk: Some complex logic might be difficult or impossible to replicate.
+    - Mitigation: Early Proof of Concept (POC) for Kubernetes provisioning in Bicep/Terraform. Implement ACI provisioning first as a less complex target to identify limitations. Consider Terraform if Bicep has significant gaps for certain platforms.
+- **Radius Application Graph Integrity**: Ensuring that recipes can correctly define and maintain relationships and connections between resources (e.g., `containers.connections`) as the current system does.
```
sylvainsf, line 78: We've talked about this a lot and I cover this in my design, if we don't move forward with my design we can still move forward with the nested types concept and using connections which would expressly create those relationships for the app graph on behalf of the recipe authors.

architecture/2025-06-compute-extensibility-feature-spec.md
```diff
@@ -0,0 +1,469 @@
...
+- Provide platform engineers with the ability to deploy Radius applications to specific platforms other than the default Kubernetes, such as Azure Container Instances (ACI), and other compute, secret store, and gateway platforms.
+- Provide abstraction punch-through mechanisms to allow platform engineers and developers to leverage platform-specific features and configurations not directly supported by Radius core (e.g. confidential containers).
+- Provide a recipe-based platform engineering experience that is consistent for user-defined types and core types.
+- Expand the ability to provision a single application definition to multiple clouds by adding the capability to provision to multiple compute platforms, secret stores, and gateway types.
+- Architecturally separate Radius core logic from platform-specific provisioning code.
+- Enable community-provided extensions to support new platforms without Radius code changes.
+
+### Non-goals (out of scope)
+<!-- What are we explicitly not trying to accomplish? -->
+- Running the Radius control plane on a non-Kubernetes platform.
```
ytimocin, line 30: Does this mean that with extensibility and core types as UDTs we shouldn't expect to be able to host UCP in a different platform?

architecture/2025-06-compute-extensibility-feature-spec.md
```diff
@@ -0,0 +1,469 @@
...
+| **Recipe** | A set of instructions that provisions a Radius resource to a Radius environment. Recipes are implemented in Bicep or Terraform. |
+| **Resource Provider (RP)** | A component responsible for handling create, read, update, delete, and list (CRUDL) operations for a specific resource type. |
+
+## Topic Summary
+<!-- A paragraph or two to summarize the topic area. Just define it in summary form so we all know what it is. -->
+Radius will be enhanced to support multiple compute platforms, secret stores, and gateway resources through a recipe-based extensibility model. This approach decouples Radius's core logic from platform-specific provisioning code. Core resource types (`containers`, `gateways`, and `secretStores`) will be implemented as User-Defined Types (UDTs) and will allow platform engineers to register Bicep or Terraform recipes for them. Radius will provide default recipes for Kubernetes and Azure Container Instances (ACI), but platform engineers can use, modify, or replace these to customize how Radius provisions resources to different environments, or to add support for entirely new platforms without requiring changes to Radius core.
```
zachcasper, line 17: Shouldn't core resource types include volumes?

architecture/2025-06-compute-extensibility-feature-spec.md
```diff
@@ -0,0 +1,469 @@
...
+
+### Scenario 2: Customize platform deployment behavior for core resources
+A platform engineer customizes the default Kubernetes recipes for `Applications.Core/containers@2025-05-01-preview` (and similarly for `gateways` and `secretStores`) to align with company-specific infrastructure policies, such as adding mandatory security sidecars, custom labels for cost tracking, or integrating with internal monitoring systems, by modifying the recipe and re-registering it.
+
+### Scenario 3: Extend Radius to a new, unsupported platform
+A platform engineer creates new Bicep or Terraform recipes to enable Radius to deploy `Applications.Core/containers@2025-05-01-preview` to AWS Fargate, `Applications.Core/gateways@2025-05-01-preview` to an AWS Application Load Balancer, and `Applications.Core/secretStores@2025-05-01-preview` to AWS Secrets Manager. They then register these recipes in their Radius environment.
```
zachcasper, line 68:
```suggestion
A platform engineer creates new Bicep or Terraform recipes to enable Radius to deploy `Applications.Core/containers@2025-05-01-preview` to AWS Elastic Container Service (ECS), `Applications.Core/gateways@2025-05-01-preview` to an AWS Application Load Balancer, and `Applications.Core/secretStores@2025-05-01-preview` to AWS Secrets Manager. They then register these recipes in their Radius environment.
```

architecture/2025-06-compute-extensibility-feature-spec.md
```diff
@@ -0,0 +1,469 @@
...
+<!-- What big questions must we answer in order to clarify our plan for this scenario.  When and how do you plan to answer those questions (prototype feature x, user research, competitive research, etc.) -->
+- **Assumption**: All essential provisioning logic currently in Radius's Go code for Kubernetes (for containers, gateways, secret stores) can be effectively replicated using Bicep or Terraform recipes.
+    - Validation: POC implementation of Kubernetes recipes for core functionalities.
+- **Assumption**: Recipes can reliably output the necessary information for Radius to construct and maintain the application graph, including resource connections.
+    - Validation: Test recipe deployments and inspect the resulting Radius graph.
+- **Question**: What is the most effective way for recipes to declare their input parameters and output properties to ensure consistency and enable validation?
+    - Exploration: Define a clear contract for recipes, potentially using a schema or metadata definition that can be validated.
+- **Question**: How can platform engineers easily test and debug their custom recipes before registering them in a live environment?
```
sylvainsf, line 100: I will address this in the type/recipe testing framework design

architecture/2025-06-compute-extensibility-feature-spec.md
```diff
@@ -0,0 +1,469 @@
...
+
+## Desired user experience outcome
+<!-- <Write this as an “I statement” that expresses the new capability from user perspective … i.e. After this scenario is implemented “I can do, x, y, z, steps in cloud native app developer and seamlessly etc... As a result <summarize positive impact on my work / business>  -->
```
zachcasper, line 112: > Azure Container Instances, AWS Fargate, or a future platform

Should be serverless container platforms including Azure Container Instances, Azure Container Apps, AWS ECS, Google CloudRun, and functions platforms including AWS Lambda, Azure Functions, and Google Cloud Functions.

architecture/2025-06-compute-extensibility-feature-spec.md
zachcasper: Fargate is just a compute option for ECS. Not the platform.

architecture/2025-06-compute-extensibility-feature-spec.md
```diff
@@ -0,0 +1,469 @@
...
+                api: {
+                  containerPort: 5000
+                }
+              }
+              env: {
+                REDIS_CONNECTION: cache.properties.connectionStrings.default
+              }
+            }
+            // This container requests confidential compute
+            extensions: [
```
sylvainsf, line 264: Do we support extensions on UDTs? Wouldn't these be additional optional properties consumed as recipe inputs instead?

architecture/2025-06-compute-extensibility-feature-spec.md
```diff
@@ -0,0 +1,469 @@
...
+                    }
+                    'Applications.Datastores/redisCaches': {
+                        default: {
+                            templateKind: 'bicep'
+                            plainHttp: true
+                            templatePath: 'ghcr.io/radius-project/recipes/azure/rediscaches:latest'
+                        }
+                    }
+                }
+        -       providers: {
```
zachcasper, line 177: The provider stores the AWS and Azure details that get past to the recipe context right? https://docs.radapp.io/reference/context-schema/#environment So we'll need the AWS account and region, and Azure subscription and resource group in the environment somewhere.

architecture/2025-06-compute-extensibility-feature-spec.md
```diff
@@ -0,0 +1,469 @@
...
+
+### Feature 3: Default Recipes for Kubernetes and ACI
+<!-- One or two sentence summary -->
+Develop and provide default, production-quality Bicep/Terraform recipes for the new UDT versions of `containers`, `gateways`, and `secretStores` that target Kubernetes (replicating current functionality) and Azure Container Instances (ACI). These will serve as out-of-the-box solutions and examples.
+
+### Feature 4: Recipe Pack Creation and Management
+<!-- One or two sentence summary -->
+Implement a system for packaging and versioning sets of related recipes (Recipe Packs) to simplify distribution and management. This includes defining a clear structure for Recipe Packs and providing CLI commands for creating, updating, and applying them. Default Recipes for Kubernetes and ACI will be provided by Radius as part of the initial Recipe Packs.
+
+### Feature 5: CLI Enhancements for Recipe Management (as necessary)
```
sylvainsf, line 459: I'd like to at least see a CLI change that takes the recipe pack manifest as input and will register all of them in one command. It would be great to see this for types as well.

architecture/2025-06-compute-extensibility-feature-spec.md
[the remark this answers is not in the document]
sylvainsf: For now, it's do-able.

architecture/2025-06-compute-extensibility-feature-spec.md
```diff
@@ -0,0 +1,469 @@
...
+#### Deploying an application with mixed standard and confidential containers
+
+This scenario demonstrates how a single application definition, containing both standard and confidential compute requirements, can be deployed to different environments, with Radius leveraging environment-specific recipes to fulfill those requirements.
+
+1.  **Platform Engineer: Configure Environments with Appropriate Recipes**
+    *   **Standard Environment (`std-env`):**
+        *   The platform engineer configures `std-env` with recipes for `Applications.Core/containers@2025-05-01-preview` that deploy to standard compute (e.g., regular ACI or Kubernetes pods). These recipes might ignore or log a warning for confidential container requests if they don't support them.
+        *   Example recipe registration (conceptual):
+            ```bash
+            rad recipe register std-container-recipe --environment std-env \
```
zachcasper, line 199: I strongly believe we should not have any recipe names. This leaks the implementation abstraction into the developers' world. If the developer needs a different recipe, that should be a different resource type.

architecture/2025-06-compute-extensibility-feature-spec.md
```diff
@@ -0,0 +1,469 @@
...
+                    'Applications.Datastores/redisCaches': {
+                        default: {
+                            templateKind: 'bicep'
+                            plainHttp: true
+                            templatePath: 'ghcr.io/radius-project/recipes/azure/rediscaches:latest'
+                        }
+                    }
+                }
+        -       providers: {
+        -           // provider configurations no longer hard-coded here
```
nithyatsu, line 178: Are we saying we will add info such as subscription and respurce grpup to the bicep recipe itself?

architecture/2025-06-compute-extensibility-feature-spec.md
```diff
@@ -0,0 +1,469 @@
...
+- **Extensibility Barriers:** Extending Radius to support new, unsupported platforms requires modifying Radius core code, which is a significant barrier for most platform teams and leads to maintenance overhead.
+- **Inconsistent Tooling:** Managing applications across different platform types often requires different toolsets and deployment pipelines, increasing complexity.
+
+### Positive user outcome
+<!-- What is the positive outcome for the user if we deliver this, i.e. what is the value proposition for the user? Remember, this is user-centric. -->
+As a platform engineer, I can confidently adopt Radius across my organization, knowing I can extend its capabilities to support any compute platform, secret store, or gateway my teams require, without waiting for built-in support or modifying Radius core. I can register, customize, and share recipes that define how Radius provisions these resources, ensuring consistency with our application definitions while maintaining flexibility in our infrastructure choices and adhering to organizational standards. This empowers my development teams to leverage Radius benefits regardless of the underlying infrastructure.
```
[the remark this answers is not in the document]
sylvainsf: Good point, we should include volumes, Will T said extenders are out of scope which I agree as I believe they'd just be replaced by new types/recipes.

architecture/2025-06-compute-extensibility-feature-spec.md
```diff
@@ -0,0 +1,469 @@
...
+                }
+              }
+              env: {
+                REDIS_CONNECTION: cache.properties.connectionStrings.default
+              }
+            }
+            // This container requests confidential compute
+            extensions: [
+              {
+                kind: 'confidentialCompute' // Example extension kind
```
zachcasper: If there are additional properties needed to be set by the developer, that should be a different resource type. I do not think extension is the right thing here, nor do I understand how one would create an extension.

architecture/2025-06-compute-extensibility-feature-spec.md
```diff
@@ -0,0 +1,469 @@
...
+
+### Detailed user experience
+ <!-- <List of steps the user goes through from the start to the end of the scenario to provide more detailed view of exactly what the user is able to do given the new capabilities>  -->
+<!-- Step 1
+Step 2
+… -->
+#### Setting up and deploying an application to a Radius environment using existing Recipes for core types:
+
+1.  **Discover existing Recipes**:
```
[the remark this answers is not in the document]
Reshrahim: We can leverage something like an Artifact Hub -https://artifacthub.io/ to discover Recipes

architecture/2025-06-compute-extensibility-feature-spec.md
```diff
@@ -0,0 +1,469 @@
...
+    *   Or, if packaged as an OCI artifact: `rad recipe-pack register --environment my-aci-env --oci-ref oci://ghcr.io/my-org/recipe-packs/aci-production-pack:1.0.0`
+    *   This command would iterate through the recipes defined in the manifest and register each one to the specified environment, similar to individual `rad recipe register` calls.
+1.  **Environment Utilizes Recipes from the Pack**:
+    *   Once the Recipe Pack is registered, the environment is configured with all the specified recipes for the core UDTs.
+    *   When applications are deployed to this environment, Radius automatically uses the corresponding recipes from the pack to provision `Applications.Core/containers@2025-05-01-preview`, `Applications.Core/gateways@2025-05-01-preview`, and `Applications.Core/secretStores@2025-05-01-preview` resources.
+1.  **Manage and Update Recipe Packs**:
+    *   Platform engineers can update the Recipe Pack manifest (e.g., point to new recipe versions, change default parameters) and re-register it. The CLI could offer options to overwrite existing registrations or manage versions of the pack within the environment.
```
lakshmimsft, line 378: We need to design/review/implement recipe versioning. Will that be a prerequisite before we work on recipe packs?

architecture/2025-06-compute-extensibility-feature-spec.md
```diff
@@ -0,0 +1,488 @@
...
+                // This should go into the runtimes.kubernetes property
+-               {
+-                   kind: 'kubernetesMetadata'
+-                   labels: {
+-                   'team.contact.name': 'frontend'
+-                   }
+-               }
+            ]
+            // This container requests confidential compute
++           runtimes: {
```
sk593, line 288: Should the confidential compute property be part of the ACI-specific recipe? Or is this an interim case while we look at Radius supporting conf containers?

architecture/2025-06-compute-extensibility-feature-spec.md
```diff
@@ -0,0 +1,469 @@
...
+            application: application
+            // Redis specific properties
+          }
+        }
+        ```
+
+3.  **Developer/CI/CD: Deploy the Same Application Definition to Both Environments**
+    *   **Deploy to Standard Environment:**
+        *   `rad deploy ./app.bicep --environment std-env`
+        *   Radius uses the `std-container-recipe` registered in `std-env`.
```
zachcasper, line 309: Radius will use the `default` recipe unless named by the developer. But specifying the recipe name is only supported by portable resource types today.

architecture/2025-06-compute-extensibility-feature-spec.md
```diff
@@ -0,0 +1,469 @@
...
+When I, as a platform engineer, try to use Radius to deploy and manage applications across my organization, I face significant hurdles if our infrastructure strategy involves more than just Kubernetes. For example, if a development team wants to deploy a service to AWS Elastic Container Service (ECS) for cost or simplicity, or if we need to integrate with a managed cloud gateway service instead of the default Radius gateway on Kubernetes, I can't easily make Radius do this. I'm forced to tell my teams they can only use Radius for Kubernetes, or I have to build and maintain complex workarounds outside of Radius. This means either Radius doesn't fit our diverse needs, or I'm stuck maintaining a custom version of Radius or else needing to contribute core code changes upstream in to Radius, which is a huge operational burden and makes upgrades very risky. These limitations prevent us from fully leveraging Radius as a unified application platform across our varied infrastructure.
```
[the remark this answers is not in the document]
sylvainsf: This is a great idea, because then recipe authors don't have to merge upstream changes though they'd still have to test for breaking changes.

architecture/2025-06-compute-extensibility-feature-spec.md
sylvainsf: We already have a requirement from some users that we will need to move those to the environment

architecture/2025-06-compute-extensibility-feature-spec.md
```diff
@@ -0,0 +1,469 @@
...
+    *   Platform engineers can update recipe definitions (e.g., to a new version from OCI or a modified local file) and re-register them using `rad recipe register` (which would effectively update the association).
+    *   They can list registered recipes and their associations.
+1.  **Configure Environment-Specific Recipe Parameters (Optional)**:
+    *   If recipes have parameters that need to be set globally for an environment, configure them using `rad recipe update` or during registration.
+
+#### Packaging and Registering a "Recipe Pack" for an Environment:
+
+1.  **Define a Recipe Pack**:
+    *   A platform engineer creates a manifest file (e.g., `recipe-pack.json` or `recipe-pack.yaml`) that defines a collection of recipes. This manifest would list each core resource type (e.g., `Applications.Core/containers@2025-05-01-preview`, `Applications.Core/gateways@2025-05-01-preview`, `Applications.Core/secretStores@2025-05-01-preview`) and associate it with a specific recipe (OCI URI or local path) and its default parameters.
+    *   Example `recipe-pack.yaml`:
```
zachcasper, line 339: Is there anything here that isn't just an Environment? Sort of seems redundant.

architecture/2025-06-compute-extensibility-feature-spec.md
```diff
@@ -0,0 +1,469 @@
...
+    *   Repeat for `gateways` and `secretStores` UDTs.
+    *   Example: `rad recipe register aci-container-recipe --environment my-aci-env --resource-type Applications.Core/containers@2025-05-01-preview --template-path oci://ghcr.io/radius-project/recipes/core/aci-container:1.0.0 --parameters defaultCpu=1 defaultMemoryInGB=2`
+1.  **Manage and Update Recipes**:
+    *   Platform engineers can update recipe definitions (e.g., to a new version from OCI or a modified local file) and re-register them using `rad recipe register` (which would effectively update the association).
+    *   They can list registered recipes and their associations.
+1.  **Configure Environment-Specific Recipe Parameters (Optional)**:
+    *   If recipes have parameters that need to be set globally for an environment, configure them using `rad recipe update` or during registration.
+
+#### Packaging and Registering a "Recipe Pack" for an Environment:
```
zachcasper, line 335: A "resource pack" would be more useful. It would be a manifest of resource types and associated recipes. I would specify the resource pack during `rad init`.

architecture/2025-06-compute-extensibility-feature-spec.md
```diff
@@ -0,0 +1,488 @@
...
+    *   **Deploy to Confidential Environment:**
+        *   `rad deploy ./app.bicep --environment confi-env`
+        *   Radius uses the `confi-container-recipe` registered in `confi-env`.
+        *   The `frontend` container is deployed as a standard container (as its definition doesn't request confidential compute).
+        *   The `backend` container's `extensions.confidentialCompute` property is interpreted by the `confi-container-recipe`, and it is deployed as a confidential container (e.g., an ACI confidential container).
+        *   The `cache` is deployed, and the connection between `backend` (now confidential) and `cache` is established.
+
+This scenario highlights that the application definition remains consistent. The underlying infrastructure and specific compute capabilities (standard vs. confidential) are determined by the recipes configured in the target Radius environment, allowing for flexible deployment to diverse compute platforms without altering the core application logic or Bicep code.
+
+#### Creating and registering custom Recipes for core types:
```
nicolejms, line 322: does this mean that we would allow end users to use their own core types (containers, gateways, secretstores) and disable ours?

architecture/2025-06-compute-extensibility-feature-spec.md
```diff
@@ -0,0 +1,469 @@
...
+            - resourceType: "Applications.Core/secretStores@2025-05-01-preview"
+            name: "aci-prod-keyvault"
+            templateKind: "bicep"
+            templatePath: "oci://ghcr.io/my-org/recipes/azure/keyvault-secretstore:1.0.0"
+            parameters:
+                skuName: "premium"
+        ```
+1.  **Package the Recipe Pack (Optional but Recommended)**:
+    *   The manifest file and any local recipe files (if not using OCI URIs exclusively) could be bundled into an OCI artifact or a simple archive (e.g., .zip, .tar.gz) for easier distribution and versioning.
+1.  **Register the Recipe Pack to an Environment**:
```
zachcasper, line 369: Register isn't the right verb here. We're configuring the environment. `rad environment create --recipe-pack <location>` seems more accurate.

architecture/2025-06-compute-extensibility-feature-spec.md
```diff
@@ -0,0 +1,152 @@
...
+<!-- Step 1
+Step 2
+… -->
+1.  **Initialize Workspace & Environment (as Platform Engineer)**:
+    *   Use `rad init` or equivalent `rad workspace`, `group`, and `environment` commands. The new environment version (e.g., `Applications.Core/environments@2025-05-01-preview`) will not have a hard-coded `compute` kind.
+    *   By default, `rad init` might still register UDTs and recipes for Kubernetes provisioning for core types like `Applications.Core/containers@2025-05-01-preview`.
+2.  **Discover/Create Recipes**:
+    *   Find community-provided recipes for desired platforms (e.g., ACI, AWS Fargate) from an OCI registry or Radius documentation.
+    *   Or, create custom Bicep/Terraform recipes for `Applications.Core/containers@2025-05-01-preview`, `Applications.Core/gateways@2025-05-01-preview`, and `Applications.Core/secretStores@2025-05-01-preview` to target a specific platform or customize existing behavior.
+3.  **Register Recipes for Core Types**:
```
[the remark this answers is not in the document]
sylvainsf: Not currently, we discussed this as well as the ability to only allow certain apps to use certain options/properties in a recipe.

architecture/2025-06-compute-extensibility-feature-spec.md
sylvainsf: I wouldn't re-use the container resource for functions, we should consider publishing a new core type of function.

architecture/2025-06-compute-extensibility-feature-spec.md
```diff
@@ -0,0 +1,488 @@
...
+              env: {
+                REDIS_CONNECTION: cache.properties.connectionStrings.default
+              }
+            }
+            connections: {
+              cache: {
+                source: cache.id
+              }
+            }
+            extensions: [
```
brooke-hamilton, line 268: I think the need for extensions, aka "punch through", goes away with RRTs and recipes. If a platform engineering team needs platform-specific configuration they would implement it in the recipe and maybe provide the developers with configuration options in the RRT.

brooke-hamilton (changes requested):
Great document! 🤖 🚀

architecture/2025-06-compute-extensibility-feature-spec.md
```diff
@@ -0,0 +1,488 @@
...
+                        default: {
+                            templateKind: 'bicep'
+                            plainHttp: true
+                            templatePath: 'ghcr.io/radius-project/recipes/azure/aci-gateway:latest'
+                            parameters: {
+                                defaultCpu: 1
+                                defaultMemoryInGB: 2
+                            }
+                        }
+                    }
```
brooke-hamilton, line 158: This example nicely demonstrates how recipes can be packaged with an environment. Do we need recipe packs if we have this capability? For example, if I want to deploy to XYZ platform, I can define an environment that contains all the recipes needed to deploy to that platform.

architecture/2025-06-compute-extensibility-feature-spec.md
willtsai: This question has come up a few times. I've added it to the "Questions to answer" section of this spec.

architecture/2025-06-compute-extensibility-feature-spec.md
willtsai: Good point, I've changed it to leverage the `rad environment update my-env --recipe-pack ...` command instead.

architecture/2025-06-compute-extensibility-feature-spec.md
willtsai: I think Recipes versioning will be a separate feature that we don't need to implement as a part of Recipe Packs. However, we should include versioning in scope for the implementation of Recipe Packs from the get go, even if the underlying Recipes packaged within the Pack don't support versioning yet.

Clarified this in the goals/non-goals

architecture/2025-06-compute-extensibility-feature-spec.md
```diff
@@ -0,0 +1,488 @@
...
+    *   They adjust any application properties to align with the schema of the new UDTs and the parameters expected by the registered recipes.
+4.  **Test Migrated Applications**:
+    *   Deploy the updated application Bicep files to the upgraded Radius environment.
+    *   Thoroughly test the application's functionality, connectivity, and configuration to ensure it behaves as expected under the new recipe-based provisioning.
+5.  **Iterate and Adjust**:
+    *   If issues arise, consult migration logs, recipe documentation, and UDT schemas to make necessary adjustments to application definitions or recipe parameters in the environment.
+6.  **Update CI/CD Pipelines**:
+    *   Modify CI/CD pipelines to use the updated application Bicep files and any new CLI commands or Radius versions required for the UDT model.
+
+## Key investments
```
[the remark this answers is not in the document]
willtsai: Yes, I've added error scenario details to the user scenarios section.

architecture/2025-06-compute-extensibility-feature-spec.md
willtsai: Agreed, a `rad environment update my-env --recipe-pack ./recipe-pack.yaml` command is proposed in the detailed user scenarios section.

architecture/2025-06-compute-extensibility-feature-spec.md
```diff
@@ -0,0 +1,615 @@
...
+- Exposing the entire set of container configurations in order to allow for platform specific capabilities may be too broad of an abstraction punch-through, resulting in container infrastructure configurations being split or duplicated across two places (i.e. in the container resource definition itself and in the Recipe).
+- May lead to a proliferation of properties in the standard container resource type by developers, making it harder to manage and control for the platform engineering team.
+
+#### Option 2: Not built into the standard container properties
+This option is where the advanced capabilities are not exposed in the standard container properties provided by Radius out of the box, but rather require platform engineers to modify the container resource definitions to enable these capabilities. This could be done by creating a custom Radius Resource Type (RRT) that extends the standard container RRT to include platform-specific properties or configurations.
+
+**Pros:**
+- Keeps the standard container resource type clean and focused on core properties, avoiding clutter from platform-specific configurations or misuse of advanced container configurations by developers.
```
sylvainsf, line 602: How is this cleaner than a runtimes property not used by developers in 99% of cases?

architecture/2025-06-compute-extensibility-feature-spec.md
```diff
@@ -0,0 +1,615 @@
...
+- Allows platform engineers to define and enforce specific configurations for advanced capabilities, ensuring consistency across applications deployed on that platform.
+- The Radius maintainers retain the option to add advanced capabilities to the standard container resource type in the future, if needed, without breaking existing applications. For example, if confidential containers become an standard offering across all platforms, the Radius maintainers can add a `confidentialCompute` property to the standard container resource type and provide default Recipes that implement it.
+
+**Cons:**
+- Platform engineers must modify the container resource definitions to enable advanced capabilities, which may require additional effort and coordination.
+- Limits the ability of developers to leverage advanced container features without involving platform engineers, potentially slowing down the development process.
+- Default Recipes may not be registered for the custom container RRT that is created to leverage platform-specific capabilities, requiring platform engineers to create and register custom Recipes for each platform they want to deploy to.
+
```
willdavsmith, line 610: could we publish both to our official definitions?

`Radius.Resources/containers`
`Radius.Resources/extensibleContainers`

architecture/2025-06-compute-extensibility-feature-spec.md
```diff
@@ -0,0 +1,469 @@
...
+    - Risk: Poor user experience due to difficult-to-debug deployment failures.
+    - Mitigation: Implement robust validation checks upon recipe registration (schema, parameters, outputs). Enhance CLI tooling for local recipe validation and testing.
+
+## Key assumptions to test and questions to answer
+<!-- If you are making assumptions that, if incorrect, would cause us to significantly alter our approach to this scenario, make them explicit here.  Also call out how / when you plan to validate key assumptions. -->
+<!-- What big questions must we answer in order to clarify our plan for this scenario.  When and how do you plan to answer those questions (prototype feature x, user research, competitive research, etc.) -->
+- **Assumption**: All essential provisioning logic currently in Radius's Go code for Kubernetes (for containers, gateways, secret stores) can be effectively replicated using Bicep or Terraform recipes.
+    - Validation: POC implementation of Kubernetes recipes for core functionalities.
+- **Assumption**: Recipes can reliably output the necessary information for Radius to construct and maintain the application graph, including resource connections.
```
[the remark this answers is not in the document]
sylvainsf: Just want to get this corrected because others will potentially read this as it's a very important design. Proposed change: **Assumption**: AppCore can reliably construct the application graph and provide recipes with all information needed to execute via their connections to other resources.

architecture/2025-06-compute-extensibility-feature-spec.md
```diff
@@ -0,0 +1,615 @@
...
+
+**Pros:**
+- Platform specific capabilities can be leveraged using built-in properties of the standard container resource type, without requiring platform engineers to modify the container resource.
+- Provides an avenue for developers to "punch-through" the Radius abstraction, which may be especially necessary for onboarding existing (brownfield) applications on to Radius.
+- Default Recipes can be leveraged for the standard container resource type for platforms that don't require platform-specific configurations. For example, the default Kubernetes Recipe can be used for the standard container resource type without any modifications, while the ACI Recipe can be modified to leverage the platform-specific capabilities.
+
+**Cons:**
+- Exposing the entire set of container configurations in order to allow for platform specific capabilities may be too broad of an abstraction punch-through, resulting in container infrastructure configurations being split or duplicated across two places (i.e. in the container resource definition itself and in the Recipe).
```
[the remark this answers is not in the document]
sylvainsf: We discussed this week about making this more generic like "optional_properties" or something similar. I bring it up because this is not just a design for containers, and there will be many places where "runtimes" does not make sense.

architecture/2025-06-compute-extensibility-feature-spec.md
```diff
@@ -0,0 +1,839 @@
...
+This option is a hybrid approach where the advanced capabilities are exposed in the standard container properties, but their enablement is controlled by a configuration in the Radius Environment definition. Platform engineers can enable or disable the punch-through for developers on a per-Environment basis. In this implementation, the configuration would be a boolean property (e.g. `allowAdvancedContainerCapabilities`) in the Environment definition that gets passed to the Recipe when it is registered. The Recipe can then use this configuration to determine whether to apply the advanced capabilities or not. The default Environments provided by Radius would have this configuration set to `true`, allowing developers to leverage advanced capabilities out of the box. The rationale is that default Radius Environments will primarily be used for testing and experimenting with Radius features, and thus enabling the punch-through by default is preferable. We expect platform engineers to create custom Environments for production or even development use cases, where they may choose to disable the punch-through for developers.
```
lakshmimsft, line 790: nit: suggestion `enableContainerOverrides` for shorter name

architecture/2025-06-compute-extensibility-feature-spec.md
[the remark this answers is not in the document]
willtsai: added your suggestion, thanks @sylvainsf

architecture/2025-06-compute-extensibility-feature-spec.md
```diff
@@ -0,0 +1,839 @@
...
+    application: app.id
+-   // These properties are removed from the volume resource
+-   //  and are instead defined in the recipe
+-   kind: 'azure.com.keyvault'
+-   resource: keyvault.id
+-   secrets: {
+-     mysecret: {
+-       name: 'mysecret'
+-     }
+-   }
```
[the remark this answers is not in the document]
lakshmimsft: is there an implied restriction for an environment to use one type of volume since a default recipe will be associated with  `Applications.Core/volumes@2025-05-01-preview`

zachcasper (commented):
High level thoughts:

* What are the proposed schema changes for the core types? I would expect the new schema to be defined in this doc.
* I'm very confused and concerned about `allowAdvancedContainerCapabilities` but there are not a lot of details here. I'm not supportive of an unstructured map. We should find the platform-specific properties that we want to expose for ACI and Kubernetes and put them in the extensions block.
* What does a vanilla environment look like? Does it have all of the recipes enumerated?
* Where are the recipes stored? How does that work in an air-gapped environment.
* We need a user story for modifying the schema for a core type. I can use rad resource-type create but I will need the original resource definition. Today there is no way to get the current definition.

architecture/2025-06-compute-extensibility-feature-spec.md
```diff
@@ -0,0 +1,708 @@
...
+
+### Detailed user experience
+ <!-- <List of steps the user goes through from the start to the end of the scenario to provide more detailed view of exactly what the user is able to do given the new capabilities>  -->
+<!-- Step 1
+Step 2
+… -->
+#### Setting up and deploying an application to a Radius environment using existing Recipes for core types:
```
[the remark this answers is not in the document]
willtsai: yes, that is right, i'll edit to make this more clear

architecture/2025-06-compute-extensibility-feature-spec.md
```diff
@@ -0,0 +1,708 @@
...
+
+1.  **Discover existing Recipes**:
+    *   Find community-provided recipes for desired platforms (e.g., ACI, AWS Fargate) from an OCI registry or Terraform module repo, or something like [Artifact Hub](https://artifacthub.io/).
```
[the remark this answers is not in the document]
willtsai: I think there's two paths:
1. Platform engineer uses the "stock" maintained and tested recipe packs that come with Radius for various container platforms.
2. Platform engineer wants to mix-n-match the "stock" recipes with some others, perhaps community-contributed ones they find or (more likely) bring their own custom recipes. They can either build their own recipe packs using this mixed bag of recipes, or they can register each individual recipe into their environment.

I'll make this distinction more clear with illustrated examples of both.

architecture/2025-06-compute-extensibility-feature-spec.md
```diff
@@ -0,0 +1,708 @@
...
+1.  **Initialize Workspace & Environment (as Platform Engineer)**:
+    *   Use `rad workspace`, `group`, and `environment` commands to create a Radius Environment and/or Workspace. 
+    *   Define an environment in a Bicep file and then deploy it using `rad deploy env.bicep`. The new environment version (e.g., `Applications.Core/environments@2025-05-01-preview`) will not have a hard-coded `compute` kind. For example, an ACI environment might look like:
+        ```diff
+        extension radius
+
+        resource env 'Applications.Core/environments@2025-05-01-preview' = {
+            name: 'my-aci-env'
+            properties: {
+        -       compute: {
```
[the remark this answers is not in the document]
willtsai: good catch, the k8s namespace will move to the `providers` property in the Environment resource

architecture/2025-06-compute-extensibility-feature-spec.md
```diff
@@ -0,0 +1,708 @@
...
+        -           // compute kind is no longer hard-coded here
+        -       }
+                recipes: {
+                    'Applications.Core/containers': {
+                        default: {
+                        templateKind: 'bicep'
+                        plainHttp: true
```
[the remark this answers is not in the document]
willtsai: this is a typo, i'll remove it

architecture/2025-06-compute-extensibility-feature-spec.md
```diff
@@ -0,0 +1,708 @@
...
+                        templatePath: 'ghcr.io/radius-project/recipes/azure/aci-container:latest'
```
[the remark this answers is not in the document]
willtsai: good suggestion, i'll update them

architecture/2025-06-compute-extensibility-feature-spec.md
```diff
@@ -0,0 +1,708 @@
...
+                        }
+                        }
+                    }
+                    'Applications.Core/gateways': {
+                        default: {
+                            templateKind: 'bicep'
+                            plainHttp: true
+                            templatePath: 'ghcr.io/radius-project/recipes/azure/aci-gateway:latest'
+                            parameters: {
+                                defaultCpu: 1
```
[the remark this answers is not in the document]
willtsai: yes, copy-paste mistake :)

architecture/2025-06-compute-extensibility-feature-spec.md
```diff
@@ -0,0 +1,708 @@
...
+                            }
+                        }
+                    }
+                    'Applications.Core/secretStores': {
+                        default: {
+                            templateKind: 'bicep'
+                            plainHttp: true
+                            templatePath: 'ghcr.io/radius-project/recipes/azure/aci-keyvault:latest'
+                            parameters: {
+                                defaultSku: 'standard'
```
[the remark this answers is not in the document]
willtsai: it should just be `sku`, i'll correct it

architecture/2025-06-compute-extensibility-feature-spec.md
```diff
@@ -0,0 +1,839 @@
...
+               // This should go into the runtimes.kubernetes property
+        -       {
+        -           kind: 'kubernetesMetadata'
+        -           labels: {
+        -           'team.contact.name': 'frontend'
+        -           }
+        -       }
+            ]
+            // This container requests confidential compute
+        +   runtimes: { // Consider renaming this to `options` or `platform`, etc. for clarity
```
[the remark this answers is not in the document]
willtsai: I don't like `extensions` because there is already an `extensions` property that will hold the Dapr configs. I propose we go with `platformOptions`

architecture/2025-06-compute-extensibility-feature-spec.md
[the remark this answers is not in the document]
willtsai: Yes, the implementation of `volumes` will be done on the Recipe going forward

architecture/2025-06-compute-extensibility-feature-spec.md
```diff
@@ -0,0 +1,708 @@
...
+        *   `rad deploy ./app.bicep --environment confi-env`
+        *   Radius uses the `confi-container-recipe` registered in `confi-env`.
+        *   The `frontend` container is deployed as a standard container (as its definition doesn't request confidential compute).
+        *   The `backend` container's `extensions.confidentialCompute` property is interpreted by the `confi-container-recipe`, and it is deployed as a confidential container (e.g., an ACI confidential container).
+        *   The `cache` is deployed, and the connection between `backend` (now confidential) and `cache` is established.
+
+This scenario highlights that the application definition remains consistent. The underlying infrastructure and specific compute capabilities (standard vs. confidential) are determined by the recipes configured in the target Radius environment, allowing for flexible deployment to diverse compute platforms without altering the core application logic or Bicep code.
+
+#### Creating and registering custom Recipes for core types:
+1.  **Create and Register Recipes for Core Types**:
```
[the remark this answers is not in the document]
willtsai: you're right, I will update this to show registration of default recipe packs.

architecture/2025-06-compute-extensibility-feature-spec.md
```diff
@@ -0,0 +1,839 @@
...
+    *   Thoroughly test the application's functionality, connectivity, and configuration to ensure it behaves as expected under the new recipe-based provisioning.
+5.  **Iterate and Adjust**:
+    *   If issues arise, consult migration logs, recipe documentation, and RRT schemas to make necessary adjustments to application definitions or recipe parameters in the environment.
+6.  **Update CI/CD Pipelines**:
+    *   Modify CI/CD pipelines to use the updated application Bicep files and any new CLI commands or Radius versions required for the RRT model.
+
+#### Initialize Radius for specific platforms with default recipes:
+
+1.  **Kubernetes**:
+* `rad init kubernetes` initializes a Radius environment for Kubernetes with default recipes for core types.
```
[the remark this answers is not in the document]
willtsai: on second thought, I agree with you. we'll keep existing `rad init` as is, just initializing a local kubernetes set up using the k8s recipe pack. I'll remove this user story.

architecture/2025-06-compute-extensibility-feature-spec.md
```diff
@@ -0,0 +1,1044 @@
...
+        +           kubernetes: {
+        +               namespace: 'default'
+        +           }
+               }
+            }
+        }
+        ```
+        > Note: the `compute` property is removed in favor of Recipe configurations, with the Kubernetes namespace moved to the providers section.
+
+    * The default Recipe packs will have the `allowPlatformOptions` parameter set to `true` for core types like `Applications.Core/containers@2025-05-01-preview`, which allows platform engineers to punch through the Radius abstraction and use platform-specific options (e.g., `containerGroupProfile` for ACI) in their application definitions.
```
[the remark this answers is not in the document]
willtsai: The idea is to minimize the out-of-box friction for users and leave it up to platform engineers to lock things down if they desire.

architecture/2025-06-compute-extensibility-feature-spec.md
```diff
@@ -0,0 +1,1044 @@
...
+
+    * By default, `rad init` will register recipes for Kubernetes provisioning for core types like `Applications.Core/containers@2025-05-01-preview` so that Radius may continue providing a local kubernetes experience out of the box.
```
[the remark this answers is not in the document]
willtsai: potentially introduce ACI, ECS, etc. as a part of the `rad init --full` experience

architecture/2025-06-compute-extensibility-feature-spec.md
```diff
@@ -0,0 +1,1044 @@
...
+
+> The proposal is add core RRTs and Recipes to a new repo called `radius-project/resource-types-contrib` where both core (built-in) and extended (optional) types and Recipes will be stored in a structure like: <img src="2025-06-compute-extensibility-feature-spec/types-contrib-repo-structure.png" alt="diagram showing types-contrib repo structure that includes root directories for core and extended types" width="700"/>
+
+#### User Story 7: As a platform engineer or app developer, I want to migrate existing Radius applications to use the new RRT-based core types and recipes, so that I can take advantage of the new extensibility features without breaking existing applications:
+
+1.  **Understand the New Model**:
+    *   Platform engineers and developers review documentation on the RRT-based core types (`Applications.Core/containers@2025-05-01-preview`, `Applications.Core/gateways@2025-05-01-preview`, `Applications.Core/secretStores@2025-05-01-preview`, `Applications.Core/volumes@2025-05-01-preview`) and the recipe-driven approach.
```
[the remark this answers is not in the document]
willtsai: Since this is the first iteration of the compute extensibility model, I think we should keep the `preview` until we've gotten adopters using it in production to validate whether it will change before we decide removing it.

architecture/2025-06-compute-extensibility-feature-spec.md
```diff
@@ -0,0 +1,1044 @@
...
+* Radius uses the registered recipe for `Applications.Core/secretStores@2025-05-01-preview` to provision the secret store (e.g., Azure Key Vault) and the recipe for `Applications.Core/containers@2025-05-01-preview` to deploy the container with the secret store mounted as a volume in the container.
+
+> Note: this replaces the current implementation that requires users to associate a KeyVault to a volume resource before mounting the volume to the container, i.e. https://docs.radapp.io/reference/resource-schema/core-schema/volumes/azure-keyvault/
+
+> This approach will require deprecation of the current [Azure KeyVault Radius Volumes resource type](https://docs.radapp.io/reference/resource-schema/core-schema/volumes/azure-keyvault/) where the Azure KeyVault will just be a Secret Store resource type provisioned using a recipe for Azure KeyVault that can be mounted to a Radius Container resource as a volume.
+
+#### User Story 9: As an application developer, I want to create custom gateway resources and configure them to route traffic to my application containers, so that I can manage ingress traffic using my organization's preferred gateway solution:
```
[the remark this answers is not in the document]
willtsai: I mean a gateway resource backed by an ingress controller of their choice (this is the "custom" part), e.g. a gateway resource backed by nginx ingress controller. I will clarify this.

architecture/2025-06-compute-extensibility-feature-spec.md
```diff
@@ -0,0 +1,1044 @@
...
+
+#### User Story 10: As an application developer, I want to create custom volume resources and mount them to my containers, so that I can use custom storage solutions or configurations:
+
+Given: my platform engineer has set up a Radius environment with recipes registered for `Applications.Core/Containers@2025-05-01-preview` and `Applications.Core/volumes@2025-05-01-preview` resources. The volume recipe is configured to provision custom volumes based on the specifications determined by my organization.
+
+1. **Define a Custom Volume Resource**:
+    * The developer defines a custom volume resource in their application Bicep file that will be provisioned by the default recipe.
```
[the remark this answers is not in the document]
willtsai: Currently, there is no way to customize the type of storage resources that can be provisioned to be mounted as storage volumes on a container, e.g. mounting Azure blob storage as a volume on a container

architecture/2025-06-compute-extensibility-feature-spec.md
```diff
@@ -0,0 +1,1156 @@
...
+        *   The `backend` container is deployed as a Kubernetes Pod with the specified tolerations and labels because the `allowPlatformOptions` parameter was set to `true` in the Kubernetes recipe pack by the platform engineer. ACI-specific properties are ignored in this case, as they are not applicable to Kubernetes.
+        *   The `cache` is deployed, and the connection between `backend` and `cache` is established.
+
+> This scenario highlights that the application definition remains consistent. The underlying infrastructure and specific compute capabilities (ACI standard vs. ACI confidential vs. Kubernetes) are determined by the recipes configured in the target Radius environment, allowing for flexible deployment to diverse compute platforms without altering the core application logic or Bicep code.
+
+> Note that Radius will have to merge the configurations from the environment and container definitions with the configurations provided within the `platformOptions` property. For example, if the `platformOptions.kubernetes.metadata.labels` is specified in the container definition, it will override the labels defined at the environment level.
```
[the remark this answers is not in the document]
willtsai: You are correct - there needs to be a technical design for this that will determine all these implementation details.

architecture/2025-06-compute-extensibility-feature-spec.md
```diff
@@ -0,0 +1,1156 @@
...
+                recipePacks: [
+                    {
+                        name: 'azure-aci-pack'
+                        uri: 'git::https://github.com/project-radius/resource-types-contrib.git//recipe-packs/azure-aci-pack?ref=1.0.0'
+                    }
+                ]
+                // providers property remains scoped to the environment
+                providers: {
+                    azure: {
+                        scope: '/subscriptions/<SUBSCRIPTION_ID>/resourceGroups/<RESOURCE_GROUP_NAME>'
```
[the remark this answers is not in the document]
willtsai: During a design discussion, we proposed a slight tweak to your proposal: subscription ID (required) can be overridden by the Recipe if it provides a `subscriptionId`.

If the neither the Environment nor Recipe specifies a `resourceGroup` and the resource to be deployed needs to be scoped to a resource group, then the deployment would fail.

architecture/2025-06-compute-extensibility-feature-spec.md
```diff
@@ -0,0 +1,1131 @@
...
+    - Exploration: Define a clear contract for recipes, potentially using a schema or metadata definition that can be validated.
+- **Question**: How can platform engineers easily test and debug their custom recipes before registering them in a live environment?
+    - Exploration: Investigate CLI tooling or local sandbox environments for recipe testing.
+- **Question**: Recipe Packs seems redundant as sets of recipes can already be registered to an Environment, which effectively acts as a Recipe Pack. Is there a need for a separate concept of Recipe Packs, or can we simply guide users to bundle their Recipes into Environment definitions?
+    - Answer: The convenience of Recipe packs will be needed so that platform engineers don't have to piece together individual Recipes for each environment from scratch as a single Recipe Pack may be re-used for many Environments. Putting everything together manually each time also opens up the door for error (e.g. forgetting to include gateway Recipes). The Environment definition is a good place to group *collections of Recipe packs* - e.g. bundle ACI and OpenAI Recipe packs together in an Environment definition.
+
```
nithyatsu, line 107: Would we provide ability to register multiple recipe packs to an environment?

architecture/2025-06-compute-extensibility-feature-spec.md
```diff
@@ -0,0 +1,1131 @@
...
+                    }
+                ]
+            }
+        }
+        ```
+   * Create (deploy) these Recipe Packs:
+        ```bash
+        rad deploy computeRecipePack.bicep
+        rad deploy dataRecipePack.bicep
+        ```
```
nithyatsu, line 175: Should we use rad register recipe-pack computeRecipePack.bicep ? These are still pointers and are not deploying a resource itself correct?

architecture/2025-06-compute-extensibility-feature-spec.md
```diff
@@ -0,0 +1,1131 @@
...
+        -       recipes: {}
+            }
+        }
+        ```
+    
+    > The `environments.properties.recipes` is removed in favor of `environments.properties.recipePacks`, which means users will no longer be allowed to add individual recipes directly to the environment. Recipes must now be packed into Recipe Packs resources before they may be added to an environment, and a Recipe Pack can contain one or more recipes. This change simplifies environment management and promotes reuse of common configurations.
+
+    > If recipes are duplicated across the recipe packs, Radius will overwrite the previous recipe with the latest one applied. This means that the `recipePacks` array property on the environment is effectively a prioritized ordered list, with the last recipe pack in the list having the highest priority in case of any recipe conflicts. In the example above, if both computeRecipePack and dataRecipePack contain a recipe for the same resource type, the one from dataRecipePack will take precedence. This behavior is intentional to ensure predictability and will be clearly documented for the user.
+
```
nithyatsu, line 197: if both compute and data have differnt recipe links for cool-namespace/cool-resourcetype ,otherwise users would be left with env having multiple recipes for a single type, but only (last) one takes effect. At first glace, someone might update the recipe pointed by first pack and wonder why it is not getting reflected and then look into subsequent recipe packs to find latest.  would it not be better to point out the conflict and avoid it?

architecture/2025-06-compute-extensibility-feature-spec.md
```diff
@@ -0,0 +1,1131 @@
...
+- `Radius.Security/secrets`: these default Recipes provision the most common secret store for the underlying platform (e.g. Kubernetes Secrets, Azure Key Vault, AWS Secrets Manager).
+- `Radius.Storage/volumes`: these default Recipes provision the most common storage solutions for the underlying platform (e.g. Kubernetes Persistent Volumes, Azure Files, Amazon Elastic File System).
+
+> `Radius.Compute/gateways` for L7 ingress will not be included in the default Recipe Packs since it doesn't make sense to force a default given the wide range of preferences for ingress controllers (e.g. Contour, NGINX, Cilium, etc.) by users. Instead, L7 ingress will be implemented separately in the gateways Recipes, and these will not be included as a part of default Recipe packs for each compute platform. If a user wants L7 ingress, they will need to separately bring in their own Recipes for gateways that will implement an ingress controller of their choice (Contour, NGINX, Cilium, Azure Application Gateway, AWS Application Load Balancer, etc.).
+
```
nithyatsu, line 245: Would this not cause a few limitations for applications using Radius such as
- no TLS termination
- cannot route to a specific container within a pod incase of multiple container on a pod ( there might be logging/ security containers/ etc in the same pod)
- cannot make url based routing which might be useful in scenarios such as for detecting auth urls and redirecting to a different service

It feels more useful to provide a sample L7 gateway and  if thats not a user favorite, they can always extend and bring in their own recipe. L4 gateway feels bit too limiting.

architecture/2025-06-compute-extensibility-feature-spec.md
```diff
@@ -0,0 +1,1131 @@
...
+        > The `scope` property is now an object with subscriptionId and resourceGroupName as separate properties for Azure environments. `subscriptionId` (required) can be overridden by the Recipe if it also sets a `subscriptionId`. `resourceGroup` (optional) can be set in the Environment or by the Recipe (value set in the Recipe overrides the value set in the Environment). If the neither the Environment nor Recipe specifies a `resourceGroup` (optional) and the resource to be deployed needs to be scoped to a resource group, then the deployment would fail. This change allows for users to have Recipes set the `resourceGroup` dynamically upon resource creation.
+
+        > If no Recipe Packs are provided, then the environment is created with the Recipe Pack which uses Bicep to deploy to Kubernetes. This Recipe Pack will include at least containers, gateways, and secrets. This preserves the existing behavior where an Environment without recipes registered can still deploy to Kubernetes given the implementation of containers, etc. in the imperative Go code.
+
```
nithyatsu, line 307: would we not want to add the recipes we have today (portable resources) to default pack?

architecture/2025-06-compute-extensibility-feature-spec.md
```diff
@@ -0,0 +1,1135 @@
...
+- **Inconsistent Tooling:** Managing applications across different platform types often requires different toolsets and deployment pipelines, increasing complexity.
+
+### Positive user outcome
+<!-- What is the positive outcome for the user if we deliver this, i.e. what is the value proposition for the user? Remember, this is user-centric. -->
+As a platform engineer, I can confidently adopt Radius across my organization, knowing I can extend its capabilities to support any compute platform, secret store, gateway, or volume my teams require, without waiting for built-in support or modifying Radius core. I can register, customize, and share recipes that define how Radius provisions these resources, ensuring consistency with our application definitions while maintaining flexibility in our infrastructure choices and adhering to organizational standards. This empowers my development teams to leverage Radius benefits regardless of the underlying infrastructure.
+
+> Note that compute extensibility will allow for one compute platform configured per Radius Environment
+
```
nithyatsu, line 60: 👍

architecture/2025-06-compute-extensibility-feature-spec.md
nithyatsu: I see we are modelling recipe-packs as radius resources. this makes sense.

architecture/2025-06-compute-extensibility-feature-spec.md
```diff
@@ -0,0 +1,1135 @@
...
+> Note: if there are existing recipes registered for the same resource type in the environment, this command will overwrite them with the recipes defined in the pack.
+
+**Update the Kubernetes namespace in the Environment**:
+* Use `rad environment update` to change the Kubernetes namespace for an existing environment.
+* Example command to update the Kubernetes namespace:
+```bash
+rad environment update my-k8s-env --kubernetes-namespace new-namespace
+```
+* This command will update the `providers.kubernetes.namespace` property in the environment definition, which will be used by Radius when deploying applications to this environment.
+* The `--kubernetes-namespace` flag is also available in the `rad environment create` command to set the initial namespace when creating a new environment.
```
nithyatsu, line 599: would it not be a good practice to deploy each application + its resources in a unique namespace? Currently, Applications RP deploys each Application and its resources in its own namespace so that administrators of the Kubernetes cluster to correctly configure RBAC and limit the scope of permissions.

architecture/2025-06-compute-extensibility-feature-spec.md
willtsai: We bottomed out on this in a discussion, so I've added the following notes to the spec to clarify the behavior:

> Only one Recipe per Resource Type is allowed within a single Recipe Pack. If multiple Recipes for the same Resource Type are included, the creation process should fail with an appropriate error message.

> If Recipes are duplicated across the Recipe Packs added to an Environment, Radius will fail the environment deployment with a clear error message indicating the conflicting Recipes. This ensures that recipe conflicts are detected early and prevents unpredictable behavior. The duplication is detected at the Resource Type level, e.g. if both computeRecipePack and dataRecipePack contain a Recipe for the same Resource Type, the deployment will fail.

architecture/2025-06-compute-extensibility-feature-spec.md
willtsai: yes, we would need to do that. I will update this note to reflect the detail you mentioned:

> If no Recipe Packs are provided, then the environment is created with the Recipe Packs which use Bicep to deploy to Kubernetes. The Recipe Packs will include Recipes for at least containers, gateways, secrets, as well as sqlDatabases, redisCaches, mongoDatabases, rabbitMQQueues, Dapr/pubSubBrokers, Dapr/secretStores, and Dapr/ to maintain parity with the "portable" resource experience today. This preserves the existing behavior where an Environment without recipes registered can still deploy to Kubernetes given the implementation of containers, etc. in the imperative Go code.

architecture/2025-06-compute-extensibility-feature-spec.md
```diff
@@ -0,0 +1,1135 @@
...
+            properties: {
+                recipePacks: [local-k8s-pack.id]
+                recipeParameters: {
+                    Radius.Compute/containers@2025-05-01-preview: {
+                        allowPlatformOptions: true
+        +               // The Kubernetes metadata.labels at the environment level can be specified here,
+        +               //   this will be passed to the Recipes which will apply them to all containers deployed to this environment.
+        +               metadata.labels: "{ 'team.contact.name': 'frontend' }"
+                    }
+                        }
```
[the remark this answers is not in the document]
nithyatsu: I just followed up the discussion on discord and understand allowPlatformOptions would be the ops knob, to enable developers to supply the punch through options.  If this sounds good, we can resolve this comment.

architecture/2025-06-compute-extensibility-feature-spec.md
```diff
@@ -0,0 +1,1135 @@
...
+
+        RECIPE PACK         RESOURCE TYPE                    RECIPE KIND     RECIPE VERSION      RECIPE LOCATION
+        computeRecipePack   Radius.Compute/containers        terraform                           https://github.com/project-radius/resource-types-contrib.git//recipes/compute/containers/kubernetes?ref=v0.48
+        computeRecipePack   Radius.Security/secrets          terraform                           https://github.com/project-radius/resource-types-contrib.git//recipes/security/secrets?ref=v0.48
+        computeRecipePack   Radius.Storage/volumes           terraform                           https://github.com/project-radius/resource-types-contrib.git//recipes/storage/volumes?ref=v0.48
+        dataRecipePack      Radius.Data/redisCaches          terraform                           https://github.com/project-radius/resource-types-contrib.git//recipes/data/redisCaches?ref=v0.48
+        ```
+
+        > Note: the `RECIPE VERSION` is equivalent to the `TEMPLATE VERSION` column that exists in the Radius CLI output today but remains blank until Recipes versioning is implemented. We're just preserving the existing output, but renaming it to `RECIPE VERSION`.
```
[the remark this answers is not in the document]
willtsai: We would think about Recipe Pack versioning as a separate effort, specifically as an effort to bring versioning for Resource Types in general (which doesn't exist today and is a greater effort)

architecture/2025-06-compute-extensibility-feature-spec.md
willtsai: Open question: need to determine if we will allow named Recipes in the Recipe Packs. That will impact the dupe detection logic.

If we decide to not allow named Recipes (i.e. only one Recipe allowed per Resource Type in an Environment), then we will need to change all the legacy Resource Types to not allow named Recipes.

architecture/2025-06-compute-extensibility-feature-spec.md
```diff
@@ -0,0 +1,1135 @@
...
+            name: 'my-k8s-env'
+            properties: {
+        -       compute: {
+        -           // compute kind is no longer hard-coded here
+        -       }
+        +       recipePacks: [local-k8s-pack.id]
+                // providers property remains scoped to the environment
+                providers: {
+        +           // The Kubernetes namespace is moved here from properties.compute
+        +           kubernetes: {
```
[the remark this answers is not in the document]
willtsai: Makes sense to not have to continue the concept of a `scope` object if we're no longer concatenating the two properties. Updated to:

> The `scope` property is deprecated in favor of splitting it out into its `subscriptionId` and `resourceGroupName` constituent parts as separate properties for Azure environments. `subscriptionId` (required) must be set in the Environment, while `resourceGroup` (optional) can be set in the Environment or by the Recipe (value set in the Recipe overrides the value set in the Environment). If the neither the Environment nor Recipe specifies a `resourceGroup` (optional) and the resource to be deployed needs to be scoped to a resource group, then the deployment would fail. This change allows for users to have Recipes set the `resourceGroup` dynamically upon resource creation.

architecture/2025-06-compute-extensibility-feature-spec.md
```diff
@@ -0,0 +1,1135 @@
...
+* The `properties.providers.azure.scope` property in the `environments` resource type is changed to an object with the subscriptionId and resourceGroupName properties separated. `subscriptionId` (required) can be overridden by the Recipe if it provides a `subscriptionId`. `resourceGroup` (optional) can be set in the Environment or by the Recipe (value set in the Recipe overrides the value set in the Environment). If the neither the Environment nor Recipe specifies a `resourceGroup` (optional) and the resource to be deployed needs to be scoped to a resource group, then the deployment would fail.
+* The following [Recipe properties](https://docs.radapp.io/reference/resource-schema/core-schema/environment-schema/#recipe-properties) are renamed for better clarity going forward:
+    * `templateKind` -> `recipeKind`
+    * `templatePath` -> `recipeLocation`
+* Recipes accept a new `allowPlatformOptions` parameter that determines whether platform-specific configurations via their own schemas (e.g., ACI containerGroupProfile, Kubernetes podSpec) are allowed in the resource type definition.
```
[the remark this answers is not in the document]
willtsai: TBD - I went with `allow` to depict a stronger sense of permission gating by the ops team rather than a capabilities gating. But open to discussing this in the implementation phase.

willtsai:
Should `Radius.Core/environments` and `Radius.Core/applications` actually be `Radius.Config/environments` and `Radius.Config/applications` to align with `Radius.Config/recipePacks`? And then have a requirement that everything in the `Radius.Config/*` namespace are not resource types that can be provisioned by Recipes, etc.

architecture/2025-06-compute-extensibility-feature-spec.md
zachcasper: When we reviewed the resource type designs with our users, we got clear feedback that developers should not be able to select a recipe. If they needed alternative deployment behavior, the platform engineer should expose a new option via a property, or create a new resource type. This is why resource types do not have a recipe property (only the legacy portable resource types do because platform engineers could not modify the schema).

My position is that there are no named recipes in resource types, therefore should not be in recipe packs. When Environments are deployed, Radius validates that the resource types in the recipe pack set are unique, else the Environment deployment fails.

zachcasper:
> Should `Radius.Core/environments` and `Radius.Core/applications` actually be `Radius.Config/environments` and `Radius.Config/applications` to align with `Radius.Config/recipePacks`? And then have a requirement that everything in the `Radius.Config/*` namespace are not resource types that can be provisioned by Recipes, etc.

Interesting thought. I don't have strong opinions here. I lean towards keeping the set of namespaces as simple as possible—I see few benefits for breaking things down into separate namespaces. Thoughts on how to decide:

* Start with the number of namespaces at an absolute minimum --> Say let's start with one
* We should have a separate namespace for application resources (containers, secrets, volumes, databases, etc.) and for system resources (environments, recipe packs, terraform settings, etc.) --> So we now have two (Radius.Config and Radius.Compute)
* We should consider how RBAC would work. Platform engineers would setup role definitions with actions (CRUDL) and scope (resource type namespaces) --> RBAC scope could be a namespace or a resource type, so this has no impact
* Add any practicalities to make the system easier to understand --> We already agreed on Radius.Data and Radius.Security

So, I would say we would have:

* `Radius.Config/environments`
* `Radius.Config/recipePacks`
* `Radius.Config/terraformSettings` (coming soon)
* `Radius.Compute/applications` (for simplicity)
* `Radius.Compute/containers`
* ...

architecture/2025-06-compute-extensibility-feature-spec.md
```diff
@@ -0,0 +1,1135 @@
...
+                providers: {
+        +           // The Kubernetes namespace is moved here from properties.compute
+        +           kubernetes: {
+        +               namespace: 'default'
+        +           }
+               }
+            }
+        }
+        ```
+        > The `compute` property is removed in favor of Recipe configurations, with the Kubernetes namespace moved to the providers section. If a Kubernetes provider is specified but no namespace is provided, then the namespace defaults to 'default' to align with Kubernetes conventions. Note that this is changed from how it's [implemented today](https://github.com/radius-project/radius/blob/main/pkg/cli/cmd/env/create/create.go#L119-L121). If the providers property is not specified, default to providers.kubernetes.namespace='default' and show a warning to the user: "Since no provider is specified for the environment, the default provider is set to 'kubernetes' with namespace 'default'.
```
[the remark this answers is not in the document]
willtsai: While this feedback makes sense, we need to investigate the impact of removing the `application-` prefix.

willtsai:
> > Should `Radius.Core/environments` and `Radius.Core/applications` actually be `Radius.Config/environments` and `Radius.Config/applications` to align with `Radius.Config/recipePacks`? And then have a requirement that everything in the `Radius.Config/*` namespace are not resource types that can be provisioned by Recipes, etc.
> 
> Interesting thought. I don't have strong opinions here. I lean towards keeping the set of namespaces as simple as possible—I see few benefits for breaking things down into separate namespaces. Thoughts on how to decide:
> 
> * Start with the number of namespaces at an absolute minimum --> Say let's start with one
> * We should have a separate namespace for application resources (containers, secrets, volumes, databases, etc.) and for system resources (environments, recipe packs, terraform settings, etc.) --> So we now have two (Radius.Config and Radius.Compute)
> * We should consider how RBAC would work. Platform engineers would setup role definitions with actions (CRUDL) and scope (resource type namespaces) --> RBAC scope could be a namespace or a resource type, so this has no impact
> * Add any practicalities to make the system easier to understand --> We already agreed on Radius.Data and Radius.Security
> 
> So, I would say we would have:
> 
> * `Radius.Config/environments`
> * `Radius.Config/recipePacks`
> * `Radius.Config/terraformSettings` (coming soon)
> * `Radius.Compute/applications` (for simplicity)
> * `Radius.Compute/containers`
> * ...

After discussion, we ended up with this. Everything that is a Radius construct gets named into `Radius.Core/*`
* `Radius.Core/environments` (Radius construct)
* `Radius.Core/recipePacks` (Radius construct)
* `Radius.Core/applications` (Radius construct)
* `Radius.Config/terraformSettings` (coming soon)
* `Radius.Compute/containers`
* ...

architecture/2025-06-compute-extensibility-feature-spec.md
willtsai: During discussion, we agreed on not allowing named Recipes. Spec updated to reflect this.

zachcasper (approved):

architecture/2025-06-compute-extensibility-feature-spec.md
```diff
@@ -0,0 +1,1136 @@
...
+
+There were several questions about whether Environments, to which platform engineers may register multiple Recipes, essentially serve the same purpose and thus negates the need for Recipe Packs. While this is a valid point, it lacks the flexibility that a dedicated Recipe Pack feature provides where a Recipe Pack can have a lifecycle independent of the Environment and thus may be re-used for many Environments. The Environment definition remains a good place to group collections of Recipe packs - e.g. bundle ACI and OpenAI Recipe packs together in an Environment definition.
+
+### Recipe Packs: Data Model vs. Tooling Approach
+
+#### Data model approach (PREFERRED)
+Recipe packs are modeled as Radius resources, which means they are objects that get saved into the Radius datastore and thus would show up in the app graph data, etc. This approach establishes a new pattern that treats Recipe packs as core Radius resource objects while the Recipes encapsulated in the packs themselves are not. This also means that Recipe packs would be deployed at the Radius resource group level and referenced in each environment within the resource group vs. being deployed within each environment.
+
```
nithyatsu, line 1076: It also helps reduce bloat on the environment resource. We have many other properties that environment supports as well, and it is a good idea to refactor some of the information out so that we dont eventually have to create large environment objects/ serialization limits.

architecture/2025-06-compute-extensibility-feature-spec.md
```diff
@@ -0,0 +1,1136 @@
...
+    `Radius.Config/*` --> `Radius.Config/*` (new namespace to support future configurations like `terraformSettings`)
+    `Radius.Resources/*` --> `Radius.Resources/*` (user defined custom resource types)
+    ```
+
+## Key investments
+<!-- List the features required to enable this scenario(s). -->
+
+### Feature 1: RRT Implementation of Core Application Model Types
+<!-- One or two sentence summary -->
+Re-implement existing core types (`Applications.Core/environments`, `Applications.Core/applications`, `Applications.Core/containers`, `Applications.Core/gateways`, `Applications.Core/secrets`, `Applications.Core/volumes`) as Radius Resource Types (RRT) with new, versioned resource type names (i.e. `Radius.Core/environments`, `Radius.Core/applications`, `Radius.Compute/containers`, `Radius.Compute/gateways`, `Radius.Security/secrets`, `Radius.Storage/volumes`). These RRTs will form the basis of the extensible application model. These new core types must support Connections and may be modified by platform engineers to allow for the configuration of platform-specific capabilities (e.g. confidential containers) in the resource types themselves.
```
brooke-hamilton, line 909: If we implement `Applications.Core/environments` and `Applications.Core/applications` as RRTs, would customers be able to modify them and create recipes for them? How would we handle this in the CLI and API? Environments and Applications are primary concepts in the CLI, graph, and dashboard, and have expectations about behavior and validation. What would happen if someone registered a new version of those types?

I think there could be value in allowing customers to add on or extend the Environment and Application to track data that is important to them, but that seems out of scope for this feature spec. What do you think?

lakshmimsft (approved):

sk593 (approved):

architecture/2025-06-compute-extensibility-feature-spec.md
willtsai: updated, please take another look @brooke-hamilton

willtsai:
> Can you move this feature spec into the features directory?

done.

PR #98, opened 2025-06-11
Design doc for Radius upgrades + Flux

willdavsmith:


tools/2025-06-radius-upgrades-flux.md
```diff
@@ -0,0 +1,436 @@
...
+#### Scenario 1: Successful Automated Upgrade
+```yaml
+# Platform engineer updates Radius version in Git
+apiVersion: helm.toolkit.fluxcd.io/v2
+kind: HelmRelease
+metadata:
+  name: radius
+  namespace: flux-system
+  annotations:
+    radapp.io/upgrade-enabled: "true"  # Opt-in to safety checks
```
ytimocin, line 58: Is this the only safety mechanism that would block the reconciler from running the preflight checks and the upgrade?

If so, this should be disabled by defautl and then users should enable it if they want.

tools/2025-06-radius-upgrades-flux.md
```diff
@@ -0,0 +1,436 @@
...
+5. Flux deploys the new version
+
+#### Scenario 3: Troubleshooting Failed Checks
+```bash
+# Platform engineer investigates blocked upgrade
+kubectl describe helmrelease radius -n flux-system
+
+# Annotations show:
+# radapp.io/upgrade-hold: "Cannot upgrade from 0.46.0 to 0.48.0: version skipping not supported"
+# radapp.io/upgrade-checked-version: "0.46.0"
```
ytimocin, line 110: Is this the only way the users can see the errors in preflight checks?

tools/2025-06-radius-upgrades-flux.md
```diff
@@ -0,0 +1,436 @@
...
+metadata:
+  annotations:
+    # User-controlled: Opt-in to upgrade safety checks
+    radapp.io/upgrade-enabled: "true"
+    
+    # Controller-managed: Last checked version (avoids redundant checks)
+    radapp.io/upgrade-checked-version: "0.46.0"
+    
+    # Controller-managed: Hold reason when upgrades are blocked
+    radapp.io/upgrade-hold: "Cannot upgrade from 0.46.0 to 0.48.0: version skipping not supported"
```
ytimocin, line 218: What is the rollback scenario here in this case? Will they run helm rollback manually?

tools/2025-06-radius-upgrades-flux.md
ytimocin: I guess this is false by default.

tools/2025-06-radius-upgrades-flux.md
```diff
@@ -0,0 +1,436 @@
...
+                       │ FluxHelmRelease │
+                       │ Reconciler      │
+                       │                 │
+                       │ • Version Check │
+                       │ • Preflight Run │
+                       │ • Suspend/Resume│
+                       └─────────────────┘
+```
+
+### Component Design
```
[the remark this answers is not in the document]
ytimocin: For the upgrade scenario, Radius controller will run the preflight checks to see if an upgrade is going to be successful.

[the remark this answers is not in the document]
ytimocin: Installing Radius via Helm will not be able to bring all the moving parts together, as far as I know. Because `rad install kubernetes` and `rad init` will do some other operations other than just installing the Radius Helm release.

tools/2025-06-radius-upgrades-flux.md
```diff
@@ -0,0 +1,436 @@
...
+# Events show:
+# Warning  PreflightFailed  Preflight checks failed: Cannot upgrade from 0.46.0 to 0.48.0
+```
+
+## Design
+
+### Architecture Overview
+
+```
+┌─────────────────┐    ┌───────────────────┐
```
lakshmimsft, line 121: will Flux be a dependency for radius?

tools/2025-06-radius-upgrades-flux.md
```diff
@@ -0,0 +1,436 @@
...
+- [ ] Controller service registration
+- [ ] RBAC configuration
+- [ ] Functional test implementation
+- [ ] End-to-end validation
+
+### Phase 3: Documentation & Release
+- [ ] User documentation
+- [ ] Example configurations
+
+## Open Questions
```
lakshmimsft, line 406: do we need more information on error handing/cleanup when each of these steps occur

PR #99, opened 2025-06-18
Radius Upgrades with GitOps

willdavsmith:


architecture/2025-06-upgrades-gitops.md
```diff
@@ -0,0 +1,127 @@
...
+  annotations:
+    "helm.sh/hook": pre-install,pre-upgrade
+    "helm.sh/hook-weight": "-5"
+    "helm.sh/hook-delete-policy": before-hook-creation,hook-succeeded
+spec:
+  template:
+    spec:
+      containers:
+      - name: preflight
+        image: ghcr.io/radius-project/preflight:{{ .Chart.AppVersion }}
```
ytimocin, line 40: I feel like another image will bring other maintenance issues with it. Is this going to be installed with each `rad init` and/or `rad install`?

architecture/2025-06-upgrades-gitops.md
```diff
@@ -0,0 +1,127 @@
...
+
+**How**: Helm hook job runs preflight checks before upgrade starts.
+
+**Context**: [Helm hooks](https://helm.sh/docs/topics/charts_hooks/) run at specific lifecycle points. Pre-upgrade hooks execute before chart resources are updated, and job failure blocks the entire upgrade.
+
+ **Real-world example**: [Kong's Helm chart](https://github.com/Kong/charts/blob/main/charts/kong/templates/migrations-pre-upgrade.yaml) uses this pattern for database migrations before upgrades.
+
+**Implementation**: 
+- Create Job template with `helm.sh/hook: pre-install,pre-upgrade`
+- Package existing `pkg/upgrade/preflight` code from `pkg/upgrade/preflight/registry.go`
```
ytimocin, line 23: This will need to have a main.go and Dockerfile. I feel like it may be over-engineering at this point of `rad upgrade kubernetes` and `rad rollback kubernetes`.

architecture/2025-06-upgrades-gitops.md
ytimocin: One more image for air-gapped environments.

ytimocin: There will have to be CLI changes to be able to set this image and tag.

architecture/2025-06-upgrades-gitops.md
```diff
@@ -0,0 +1,127 @@
...
+ctrl.NewControllerManagedBy(mgr).
+    For(&argov1alpha1.Application{}). // ArgoCD  
+    Complete(&ArgoUpgradeReconciler{})
+```
+
+**Pros**:
+- Simplest to implement and maintain, taps into Flux and ArgoCD resources and doesn't require large Radius Helm chart changes
+
+**Cons**:
+- Requires separate reconcilers for Flux vs ArgoCD
```
ytimocin, line 122: Can we at least have a Base reconciler for different GitOps implementations? Like `base.go` and then `flux.go` and `argocd.go` may have different functions based on the need.

architecture/2025-06-upgrades-gitops.md
```diff
@@ -0,0 +1,127 @@
...
+- Possible race conditions - Flux HelmRelease controller and ArgoCD Application controller might start the upgrade before the Radius upgrade controller can stop it
+
+## Open Questions
+
+Q: Should we consider using the `rad` CLI directly in the `preflight` container?
```
willtsai, line 127: Would this help with eliminating the duplicated code problem in option 1?

brooke-hamilton (changes requested):
Describe why the proposed option is the right option. Also, the design template has a section for "alternatives considered". I think the two options you do not recommend should be there with an explanation of why they are not recommended.

architecture/2025-06-upgrades-gitops.md
willdavsmith: There needs to be a flag on rad install kubernetes to not install the preflight container

PR #123, opened 2025-11-11
Add on-call team as codeowner for Dependabot updates

kachawla:
Add radius-project/on-call as additional codeowner for .github/workflows/ to allow dependency updates without blocking on maintainers/approvers.


DariuszPorowski (changes requested):
File contains errors, this part is missing:

image

kachawla:
> File contains errors, this part is missing:
> 
> image

Fixed.

